Issue #3, opened 2019-06-26
Transition from Door Closed Start to Movement is too permissive


trisapeace:
In `sessionState.js`, the code for the transition from Door Closed Start to Movement is 
```
                case STATE.DOOR_CLOSED_START:
                    {
                        if (xethru.state != XETHRU_STATE.NO_MOVEMENT || motion.signal == "active") {
                            state = STATE.MOVEMENT;
                        }
                        break;
                    }
```

But that means that the transition happens when the XeThru has states such as `UNKNOWN`, `INITIALIZING`, and `ERROR` and also when the XeThru has no state at all. (Will the XeThru ever not have a state in the DB?). I feel like this is too permissive and that XeThru states that would cause this transition to happen would be only `MOVEMENT`, `BREATHING`, and maybe `MOVEMENT_TRACKING`.

mariocimet:
Yay! @trisapeace puts this one down with #14

PR #14, opened 2019-07-03
Restrict transition from DOOR_CLOSED_START to MOVEMENT...

trisapeace:
... to specific XeThru states that indicate movement (Breathing, Movement, and Movement Tracking) rather than leaving it open to
all XeThru states that were not NO_MOVEMENT

Updated tests accordingly.

Fixes Issue #3


trisapeace:
@mariocimet feel free to merge this when you've had a look-see. Thanks!

Issue #1, opened 2019-06-11
"Door Open, Start Session" is a standby state where the system is blind


sampathsatti:
"Door Open, Start Session" is a session state that is triggered by the door opening when the system is in the "no presence, no session state".

1. The nomenclature is a slight misnomer, as the session doesn't actually start.

2. Rather the system is an standby state, where it waits for the door to close. 

3. Additionally, the system is blind to any behavior in the bathroom

sampathsatti:
"Door Open Start Sessions" must start a session when motion is active and move into the motion state

mariocimet:
assigning @trisapeace to make a failing test for this

trisapeace:
@sampathsatti I started a branch called "fix_door_open_blindness" where I added test that should cover the movement transition between DOOR_OPEN_START and MOVEMENT_DETECTED. 

It works under the assumption that the criteria for movement is the same as the transition from NO_PRESENCE_NO_SESSION to MOVEMENT_DETECTED.

Please use this branch for further changes for this Issue. Then you'll be able to run `npm test` to see the tests pass :-) Thanks!

mariocimet:
As mentioned in the pull request, there's some tests passing for this, but I wonder if we've specified the problem correctly. I think that what would happen now is that the system would be *less blind* but still possibly blind if movement_fast is under the threshold (even if the SmartThings Motion Sensor is activated).

mariocimet:
This needs a wholesale bottom-up system change: we're planning on redoing the state machine anyways and this would be an outcome of that.

PR #17, opened 2019-07-07
Fix open door blindness

mariocimet:
Addressing #1 . Tests are passing - but I wonder whether this is the implementation that we want. This is a more stringent transition than some of the other ones, particularly this one:

`                case STATE.DOOR_CLOSED_START:
                    {
                        if (xethru.state != XETHRU_STATE.NO_MOVEMENT || motion.signal == MOTION_STATE.MOVEMENT) {
                            state = STATE.MOVEMENT;
                        }
                        break;
                    }`

### DIFF
--- a/SessionState.js
+++ b/SessionState.js
@@ -57,6 +57,9 @@ class SessionState {
                         if (door.signal == DOOR_STATE.CLOSED) {
                             state = STATE.DOOR_CLOSED_START;
                         }
+                        else if (xethru.mov_f > residual_mov_f && xethru.state != XETHRU_STATE.NO_MOVEMENT && motion.signal == MOTION_STATE.MOVEMENT) {
+                            state = STATE.MOTION_DETECTED;
+                        }
                         break;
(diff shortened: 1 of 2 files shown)


trisapeace:
I'm not sure. 

I guess it depends on whether we trust door opens in the same way as we trust door open-then-close or the same way as when nothing happens with the door. I don't have a good feeling for how hard it is to pass the movement threshold, so it's probably up to you two to make the call. It would be no problem to change the tests to the more-forgiving case.

mariocimet:
Yeah.... I don't really have a good feeling for how hard it is either. I think the way it is right now is fine, and the bigger question is how we establish thresholds and other parameters in the code. One way would be to collect a bunch of data and establish the cutoff at some low percentile of the movement data points (i.e. the bottom 1 percentile cutoff).

mariocimet:
We don't actually want this right now, addressing this issue will likely have to wait until Milestone 2

PR #45, opened 2020-08-17
Adding fallback phone functionality

sampathsatti:


index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
...
+  }
+  //else do nothing
+}
+
+async function fallbackMessage(sessionid) {
+console.log("Fallback message being sent");
+let session = await db.getSessionWithSessionId(sessionid); // Gets the updated state for the chatbot
+var location = session.locationid;
+locationData = await db.getLocationData(location)
+if(session.chatbot_state == 'Waiting for Response') {
```
trisapeace, line 518: I suggest using the `STATE.WAITING_FOR_RESPONSE` constant instead of a string. This will be more robust if we decide to change the strings in the future.

index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
...
+      await db.saveChatbotSession(session);
+  }
+  //else do nothing
+}
+
+async function fallbackMessage(sessionid) {
+console.log("Fallback message being sent");
+let session = await db.getSessionWithSessionId(sessionid); // Gets the updated state for the chatbot
+var location = session.locationid;
+locationData = await db.getLocationData(location)
```
trisapeace, line 517: Like in `reminderMessage`, I would move this DB call into your the body of your `if` statement.

index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
...
-    let session = await db.getMostRecentSession(location); // Gets the updated state for the chatbot
-    locationData = await db.getLocationData(location)
-    if(session.chatbot_state == 'Started') {
-        //send the message
-        await sendTwilioMessage(locationData.twilio_number, session.phonenumber, `This is a reminder to check on the bathroom`)
-        session.chatbot_state = 'Waiting for Response';
-        await db.saveChatbotSession(session);
-    }
-    //else do nothing
+async function reminderMessage(sessionid) {
```
trisapeace, line 499: At some point we really need to start writing tests for the non-state-machine parts of this codebase.

index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
     locationData = await db.getLocationData(location)
```
trisapeace, line 492: It's not part of this pull request, but since you're in the area, I notice that there are some indentation issues in the `sendTwilioMessage` function. Now might be a good time to clean that up.

index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
...
-        await sendTwilioMessage(locationData.twilio_number, session.phonenumber, `This is a reminder to check on the bathroom`)
-        session.chatbot_state = 'Waiting for Response';
-        await db.saveChatbotSession(session);
-    }
-    //else do nothing
+async function reminderMessage(sessionid) {
+  console.log("Reminder message being sent");
+  let session = await db.getSessionWithSessionId(sessionid); // Gets the updated state for the chatbot
+  var location = session.locationid;
+  locationData = await db.getLocationData(location)
```
trisapeace, line 503: I would move this DB call into your the body of your if statement. Database calls are expensive in terms of time. We want to avoid making more calls than necessary. In this case, we don't need to know the locationData unless the chatbot_state indicates that we need to send the message, right?

index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
...
+  if(session.chatbot_state == 'Started') {
```
trisapeace, line 504: I suggest using the `STATE.STARTED` constant instead of a string. This will be more robust if we decide to change the strings in the future.

index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
...
+      //send the message
+      await sendTwilioMessage(locationData.twilio_number, session.phonenumber, `This is a reminder to check on the bathroom`)
+      session.chatbot_state = 'Waiting for Response';
```
trisapeace, line 507: I suggest using the `STATE.WAITING_FOR_RESPONSE` constant instead of a string. This will be more robust if we decide to change the strings in the future.

index.js
```diff
@@ -492,21 +492,42 @@ async function sendInitialChatbotMessage(session) {
...
+var location = session.locationid;
+locationData = await db.getLocationData(location)
+if(session.chatbot_state == 'Waiting for Response') {
+    //send the message
+    twilioClient.messages.create({
+      body: `An alert to check on the washroom at ${locationData.location_human} was not responded to. Please check on it`,
+      from: locationData.twilio_number,
+      to: locationData.fallback_number
+  })
+  .then(message => console.log(message.sid))
```
trisapeace, line 525: What is `message.sid`?

db/004-fallbackphone.sql
```diff
@@ -0,0 +1,22 @@
...
+    migrationId := 4;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        -- ADD SCRIPT HERE
+        ALTER TABLE locations ADD COLUMN fallback_number text default '+17786810411';
```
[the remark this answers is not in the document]
sampathsatti: Great point, and something I didn't really think about. In this case, I'm fine w/ the risk since I already get a spam call or two a day, and my phone can be the default fallback until we change it...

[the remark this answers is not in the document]
sampathsatti: Thanks, I will make the change.

README.md
```diff
@@ -18,3 +18,8 @@ is created
 ### To run the tests locally
 
 1. Run `npm test`
+
+#### Goal
```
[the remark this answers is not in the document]
sampathsatti: This commit was a result of me testing out the devenv branch as a candidate for going through the docker, k8s workflow and this seems to be a artifact of that. I'll revert any changes to readme.

db/db.js
```diff
@@ -178,6 +178,24 @@ async function getMostRecentSession(locationid) {
   }
 }
 
+// Gets session with a specific SessionID
+
+async function getSessionWithSessionId(sessionid) {
+  try{
```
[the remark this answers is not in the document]
sampathsatti: I'm not resolving this right now, maybe looping in @schwarrrtz since he implement a linter recently on the buttons codebase?

index.js
sampathsatti: Great point, Should definitely use the enum file that exists for this explicit reason :)

index.js
sampathsatti: Done, and I noticed that we don't also have uniformity on using semi-colons or not at all - I'll look into linter options...

index.js
sampathsatti: Cleaned.

index.js
sampathsatti: Done for both functions... And I'll keep the cost of function calls in mind going forward :)

index.js
sampathsatti: https://support.twilio.com/hc/en-us/articles/223134387-What-is-a-Message-SID- Looks like this but I don't know why we're printing that to the logs. @mariocimet might know more than me?

db/004-fallbackphone.sql
```diff
@@ -0,0 +1,22 @@
+DO $migration$
```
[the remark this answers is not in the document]
sampathsatti: Wondering if having a managed postgres database will imply that the implementation for this will have to be different?

db/db.js
```diff
@@ -383,6 +383,15 @@ async function isOverdoseSuspected(xethru, session, location) {
     let condition2 = 1 * (session.still_counter > still_threshold); //seconds
     let condition3 = 1 * (sessionDuration > sessionDuration_threshold);
     let condition4 = 1 * (0);
+
+    if(condition2 == 1){
+      let alertReason = 'Duration';
+    }
+
+    if( (condition1==1) || (condition3==1) ){
```
mariocimet, line 391: when you use 'let' inside these blocks, you're only declaring a local variable - so unless the code that's using them is inside what gets executed by the if statement, alertReason will be undefined, and the update od_flag query will fail.

sampathsatti: Ahh that makes sense. Thanks!

db/db.js
```diff
@@ -365,6 +383,17 @@ async function isOverdoseSuspected(xethru, session, location) {
     let condition2 = 1 * (session.still_counter > still_threshold); //seconds
     let condition3 = 1 * (sessionDuration > sessionDuration_threshold);
     let condition4 = 1 * (0);
+
+    let alertReason='NotSet'
```
trisapeace, line 387: Is it possible for both `condition2` and either `condition1` or `condition3` to be `1` at the same time?

If so, this logic looks like `alertReason` will be set to `Stillness` when maybe you'd expect it to be `Duration and Stillness`. Are you concerned about this level of granularity?

Also, I suggest that you make these strings into constants.

db/redis.js
```diff
@@ -1,9 +1,10 @@
+require('dotenv').config();
 const Redis = require("ioredis");
 const radarData = require('./radarData.js');
 const doorData = require('./doorData.js');
 const stateData = require('./stateData.js');
-redisConnectionString = process.env.REDIS_CONNECTION_STRING
-const client = new Redis(6379,"10.245.30.53"); // uses defaults unless given configuration object
+require('dotenv').config();
```
trisapeace, line 6: This is a repeat of Line 1 of this file.

index.js
```diff
@@ -476,39 +472,88 @@ io.on('connection', (socket) => {
...
+  }
+//else do nothing
+} 
+
+
 //Heartbeat Helper Functions
+
+async function checkHeartbeat() {
+
+ for(let i = 0; i < locations.length; i++){
```
trisapeace, line 534: I'm finding the single-space indentation very difficult to read. I don't think we do this anywhere else.

index.js
```diff
@@ -476,39 +472,88 @@ io.on('connection', (socket) => {
...
+  let currentTime = moment();
+  let latestXethru = moment(XeThruData.timestamp, "x");
+  let XeThruDelayMillis = currentTime.diff(latestXethru);
+
+  if(XeThruDelayMillis > XETHRU_THRESHOLD_MILLIS && !location.xethru_sent_alerts) {
+    console.log(`XeThru Heartbeat threshold exceeded; sending alerts for ${location.locationid}`)
+    await db.updateSentAlerts(location.locationid, true)
+    sendAlerts(location.locationid)
+  }
+  else if((XeThruDelayMillis < XETHRU_THRESHOLD_MILLIS) && location.xethru_sent_alerts) {
```
trisapeace, line 549: This looks like it's a copy-paste of `index.js` lines 339-342. If it needs to be in both places, could it be put into a function that is called by both places? That will help if something needs to change in the logic then it will only need to be changed in one place.

db/db.js
mariocimet: My guess would be that we'd want to be more conservative and default to the more severe one, which is stillness?

@sampathsatti what's your take on this?

SessionState.js
```diff
@@ -3,7 +3,7 @@ const XETHRU_STATE = require('./SessionStateXethruEnum.js');
 const OD_FLAG_STATE = require('./SessionStateODFlagEnum');
 const DOOR_STATE = require('./SessionStateDoorEnum.js');
 const Sentry = require('@sentry/node');
-Sentry.init({ dsn: 'https://a95c3fc9c5fd49b29dbe5672228184a0@sentry.io/2556408' });
+// Sentry.init({ dsn: 'https://a95c3fc9c5fd49b29dbe5672228184a0@sentry.io/2556408' });
```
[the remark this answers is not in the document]
mariocimet: we don't!

@sampathsatti can you confirm whether this is the right sentry URL to use?

PR #48, opened 2020-09-14
Mc heartbeat

mariocimet:


index.js
```diff
@@ -19,7 +19,7 @@ const app = express();
 const http = require('http').Server(app);
 const io = require('socket.io')(http);
 
-const XETHRU_THRESHOLD_MILLIS = 60*10*1000;
+const XETHRU_THRESHOLD_MILLIS = 60*1000;
```
trisapeace, line 22: We were off by a factor of 10?

PR #52, opened 2020-10-01
Minimal changes to distinguish between multiple bathrooms at a location (CU-cumnc0)

trisapeace:
- The /sms endpoint is hit when the Responder phone replies to an alert.
  Using the `req.body.From` as the lookup phone number means that we're
  using the Responder phone number. This was fine when there was only
  a single bathroom per location because the Responder phone would line
  up with exactly one bathroom. However, now that we have some locations
  with multiple bathrooms, we need to the use the `req.body.To` number
  to disambiguate.
- Investigating this bug revealed a lot of test data in the `sessions`
  table that I didn't want to change. So instead of updating the
  `phonenumber` column in `sessions` (which may also be confusing since
  the term `phonenumer` in the `locations` table specifically refers to
  the Responder phone number), I instead just changed the query to join
  with `locations` and get the `twilio_number` for the given location
- This may make our performance worse, but since the `locations` table
  doesn't have many entries, I'd be surprised if it's a big difference
- It also means that we aren't really using the `phonenumber` column in the
`sessions` table anymore. But I don't want to remove it until we figure out 
whether we need that data for anything.

### DIFF
--- a/db/db.js
+++ b/db/db.js
@@ -198,9 +198,9 @@ async function getSessionWithSessionId(sessionid){
 
 
 // Gets the last session data in the table for a specified phone number
-async function getMostRecentSessionPhone(phone){
+async function getMostRecentSessionPhone(twilioPhone){
   try {
-      const results = await pool.query("SELECT * FROM sessions WHERE phonenumber = $1  AND start_time > (CURRENT_TIMESTAMP - interval '7 days') ORDER BY start_time DESC LIMIT 1", [phone]);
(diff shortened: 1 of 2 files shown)


trisapeace:
Task linked: [CU-cumnc0 Fix session updating for locations with multiple bathroom](https://app.clickup.com/t/cumnc0)

mariocimet (approved):

mariocimet:
Looks great! good to merge for me. @sampathsatti any thoughts

sampathsatti:
Looks good to me too!

sampathsatti (approved):

trisapeace:
Oh! Looks like this got merged into `master`. Is this OK @sampathsatti ? Or should it be reverted and merged into `devenv`?

PR #54, opened 2020-10-09
Add ESLint (CU-fuu0rw)

trisapeace:
Hi, I was doing some ESLint stuff for Buttons and figured I could whip something up here too if you're interested.

- Used the same .eslintrc.json as Buttons and the brave-alert-lib because we want to keep our formatting consistent across all our JS projects
- Most of the lint errors were indentation, unused variables, and undeclared variables, which means this PR has a lot of changes that *shouldn't* affect functionality.

There are two changes that **_I am concerned_** about though, so I split them out into two different commits:
* Added an index into the `locations` array in a for-loop where it was using an undeclared and unassigned variable `location` instead. **_I'm pretty sure this is a bug that is really in the code_**. Is it a big deal? Does it need to be fixed more urgently than this pull request? (https://github.com/bravetechnologycoop/ODetect-Backend-Local/commit/bdfe84967a451da5647eebe2aea6c508446fabf4). This also helps prove the benefit of using a linter :smile: 
* Removed `resetSession` in `index.js` because it wasn't being used anywhere. I don't think this should affect anything, but wanted to check whether it was *supposed* to be used somewhere and isn't (https://github.com/bravetechnologycoop/ODetect-Backend-Local/commit/1b69e38c34b780136fdab4fbd61d01d433b34612)

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,52 +33,52 @@ var start_times = {};
 
 // Update the list of locations every minute
 setInterval(async function (){
-  let locationTable = await db.getLocations()
-  let locationsArray = []
-  for(let i = 0; i < locationTable.length; i++){
-    locationsArray.push(locationTable[i].locationid)
-  }
-  locationsArray;
-  io.sockets.emit('getLocations', {data: locationsArray});
-  for(let i = 0; i < locationsArray.length; i++){
-    await checkHeartbeat(locationsArray[i])
-  }
+    let locationTable = await db.getLocations()
+    let locationsArray = []
+    for(let i = 0; i < locationTable.length; i++){
+        locationsArray.push(locationTable[i].locationid)
+    }
+    locationsArray;
+    io.sockets.emit('getLocations', {data: locationsArray});
+    for(let i = 0; i < locationsArray.length; i++){
+        await checkHeartbeat(locationsArray[i])
+    }
 }, LOCATION_UPDATE_FREQUENCY)
 
 
 // These states do not start nor close a session
 let VOIDSTATES = [
-  STATE.RESET,
-  STATE.NO_PRESENCE_NO_SESSION,
-  STATE.DOOR_OPENED_START,
-  STATE.MOVEMENT,
-  STATE.STILL,
-  STATE.BREATH_TRACKING,
-  STATE.STARTED,
-  STATE.WAITING_FOR_RESPONSE,
-  STATE.WAITING_FOR_CATEGORY,
-  STATE.WAITING_FOR_DETAILS
+    STATE.RESET,
+    STATE.NO_PRESENCE_NO_SESSION,
+    STATE.DOOR_OPENED_START,
+    STATE.MOVEMENT,
+    STATE.STILL,
+    STATE.BREATH_TRACKING,
+    STATE.STARTED,
+    STATE.WAITING_FOR_RESPONSE,
+    STATE.WAITING_FOR_CATEGORY,
+    STATE.WAITING_FOR_DETAILS
 ];
 
 // These states will start a new session for a certain location
 let TRIGGERSTATES = [
-  STATE.DOOR_CLOSED_START,
-  STATE.MOTION_DETECTED
+    STATE.DOOR_CLOSED_START,
+    STATE.MOTION_DETECTED
 ];
 
 // These states will close an ongoing session for a certain location
 let CLOSINGSTATES = [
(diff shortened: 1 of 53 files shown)


trisapeace:
What could make this easier to review is if you ignore the purely whitespace differences in GitHub like this:
image

index.js
```diff
@@ -33,52 +33,52 @@ var start_times = {};
...
-  }
+    let locationTable = await db.getLocations()
+    let locationsArray = []
+    for(let i = 0; i < locationTable.length; i++){
+        locationsArray.push(locationTable[i].locationid)
+    }
+    locations = locationsArray;
+    io.sockets.emit('getLocations', {data: locationsArray});
+    for(let i = 0; i < locationsArray.length; i++){
+        await checkHeartbeat(locations[i])
```
trisapeace, line 44: Here is the change that I think was actually a bug.

trisapeace: Looks like @mariocimet already fixed this in `master`.

integrationTesting/alertListener.js
```diff
@@ -12,24 +12,23 @@ app.use(bodyParser.json());
...
+    })
 });
 
 app.get('/*', async function (req, res) {
     res.send('You reached the ODetect test alert listener');
 });
 
 
-server = app.listen(8081)
+app.listen(8081)
```
sampathsatti, line 33: So cool that the linter can pick up things like - "Hmm server isn't used later, let's not declare it at all"

index.js
```diff
@@ -231,217 +231,197 @@ smartapp
...
+                autoResetSession(currentLocationId);
+            }
+        }
     }
-  }
 }, WATCHDOG_TIMER_FREQUENCY)
 
 // Handler for income SmartThings POST requests
-app.post('/api/st', function(req, res, next) {
+app.post('/api/st', function(req, res) {
```
sampathsatti, line 281: Confirming that our middleware doesn't use succession of any kind?

sampathsatti: @mariocimet

index.js
```diff
@@ -231,217 +231,197 @@ smartapp
...
-    }
-    else{
-      console.log("There is no open session to reset!")
-    }
-  }
-  catch {
-    console.log("Could not reset open session");
-  }
-}
-
```
sampathsatti, line 251: I think we used reset session when a button on the front end was clicked. I wonder if the link is broken :?

sampathsatti: Actually, we removed this functionality from the front end interface. So removing this function makes sense. When we do session resets through code, autoResetSession is used instead. The naming of that function doesn't make sense since though it is called automatically by the code, functionality is still session reset (We did have separate functions to have different chatbot/session notes)

sampathsatti (approved):
Looks great to me, thanks for implementing the linter Theresa! 

We should definitely test on dev before deploying to prod, in which case, if we do confirm that locations/locations is a bug, we should carve that commit out and deploy.

integrationTesting/alertListener.js
trisapeace: I know! It's going to be so useful. And once it's in place, then we can start adding stricter rules. I think that this will be really important as we get more people working in this code base. Consistency is so important for readability and code quality.

index.js
trisapeace: Ooo, actually there is a neat way around this. We can add an `eslint-disable-next-line` comment right above it which will make it not complain about the unused `next` variable. We used this in Buttons for ones like this where it could useful in the future to remember that those parameters exist even if we don't use them now. I'll update this where I can.

index.js
trisapeace: Thanks for explaining! This must have been why I had the vague feeling that reset wasn't being used. Makes sense.

index.js
trisapeace: Fixed by adding `next` parameter back in and then telling `eslint` to not checkout for `no-unused-vars` on this line.

trisapeace:
@sampathsatti 

Mario's already fixed the `location` bug in master (https://github.com/bravetechnologycoop/ODetect-Backend-Local/blob/devenv/index.js#L44). So no need to split that one out separately.

If I deploy this to `odetect-dev` can you please help me with testing it? I haven't figured out a good way to do so.

trisapeace:
Task linked: [CU-fuu0rw Automate linting](https://app.clickup.com/t/fuu0rw)

mariocimet (approved):

PR #55, opened 2020-10-23
Fix fallback message sending

trisapeace:
* Typo that means that no fallback messages are ever sent (because their toPhoneNumber value is always empty)
* Typo that means that the log message is incorrect


sampathsatti (approved):
Looks good to be merged!

mariocimet (approved):

PR #58, opened 2020-11-20
Fix README instructions (front-end and redis IP)

trisapeace:
- Missed a step in the front-end instructions. Need to make sure that it's directing to the correct server
- Had the wrong instructions on where to find the value for the REDIS_CLUSTER_IP value


sampathsatti (approved):
Looks good, thanks for making these edits Theresa!

mariocimet (approved):

PR #61, opened 2020-12-22
CU-h2y2nt im21 endpoint

sampathsatti:
This PR implements functionality to handle and store door sensor data from the IM21 sensors and the Particle Device vitals on the Redis Database.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,15 +10,19 @@ the code was deployed.
 
 ## [Unreleased]
 
+### Added
+- Device health endpoint (CU-h2y2nt)
+- IM21 door sensor endpoint (CU-h2y2nt)
+
 ## [1.2.0] - 2020-12-10
 ### Added
 - New mustache frontend (CU-dwr2xb)
-- Script which runs all database migrations 
+- Script which runs all database migrations
 - Integration tests (CU-gcuxtz)
 - Utility functions for test setup and teardown (CU-gcuxtz)
 - Helpers for test environment variables (CU-gcuxtz)
-- Changelog 
-- Build scripts for tagging and pushing containers 
+- Changelog
+- Build scripts for tagging and pushing containers
 - Linting
 
 ### Fixed
--- a/IM21DoorStatusEnum.js
+++ b/IM21DoorStatusEnum.js
@@ -1,0 +1,11 @@
+'use strict'
+
+const IM21_DOOR_STATE = {
+    OPEN: '02',
+    CLOSED: '00',
+    HEARTBEAT_OPEN: '0A',
+    HEARTBEAT_CLOSED: '08',
+    LOW_BATT: '04'
+}
+
+module.exports = Object.freeze(IM21_DOOR_STATE)
--- a/db/db.js
+++ b/db/db.js
@@ -30,7 +30,7 @@ function createSessionFromRow(r) {
 }
 
 function createLocationFromRow(r) {
-    return new Location(r.locationid, r.display_name, r.deviceid, r.phonenumber, r.detectionzone_min, r.detectionzone_max, r.sensitivity, r.led, r.noisemap, r.mov_threshold, r.duration_threshold, r.still_threshold, r.rpm_threshold, r.xethru_sent_alerts, r.xethru_heartbeat_number)
+    return new Location(r.locationid, r.display_name, r.deviceid, r.phonenumber, r.detectionzone_min, r.detectionzone_max, r.sensitivity, r.led, r.noisemap, r.mov_threshold, r.duration_threshold, r.still_threshold, r.rpm_threshold, r.xethru_sent_alerts, r.xethru_heartbeat_number, r.door_particlecoreid, r.radar_particlecoreid)
 }
 
 // The following functions will route HTTP requests into database queries
@@ -379,6 +379,23 @@ async function getLocationData(locationid) {
 
 }
 
+// Retrieves the locationid corresponding to a particle device coreID
+async function getLocationIDFromParticleCoreID(coreID){
+    try{
+        const results = await pool.query("SELECT (locationid) FROM locations WHERE door_particlecoreid = $1 OR radar_particlecoreid = $1", [coreID])
+        if(results == undefined){
+            console.log('Error: No location with associated coreID exists')
+            return null
+        }
+        else{
+            return results.rows[0].locationid
+        }
+    }
+    catch(e){
+        console.log(`Error running the getLocationIDFromParticleCoreID query: ${e}`)
+    }
+}
+
 // Retrieves the locations table
 async function getLocations() {
     try {
@@ -397,10 +414,10 @@ async function getLocations() {
 }
 
 // Updates the locations table entry for a specific location with the new data
-async function updateLocationData(deviceid, phonenumber, detection_min, detection_max, sensitivity, noisemap, led,  rpm_threshold, still_threshold, duration_threshold, mov_threshold, location) {
+async function updateLocationData(deviceid, phonenumber, detection_min, detection_max, sensitivity, noisemap, led,  rpm_threshold, still_threshold, duration_threshold, mov_threshold, door_particlecoreid, radar_particlecoreid, location) {
     try {
-        let results = await pool.query("UPDATE locations SET deviceid = $1, phonenumber = $2, detectionzone_min = $3, detectionzone_max = $4, sensitivity = $5, noisemap = $6, led = $7, rpm_threshold = $8, still_threshold = $9, duration_threshold = $10, mov_threshold = $11 WHERE locationid = $12 returning *", 
-            [deviceid, phonenumber, detection_min, detection_max, sensitivity, noisemap, led, rpm_threshold, still_threshold, duration_threshold, mov_threshold, location]);
+        let results = await pool.query("UPDATE locations SET deviceid = $1, phonenumber = $2, detectionzone_min = $3, detectionzone_max = $4, sensitivity = $5, noisemap = $6, led = $7, rpm_threshold = $8, still_threshold = $9, duration_threshold = $10, mov_threshold = $11, door_particlecoreid = $12, radar_particlecoreid = $13 WHERE locationid = $14 returning *", 
+            [deviceid, phonenumber, detection_min, detection_max, sensitivity, noisemap, led, rpm_threshold, still_threshold, duration_threshold, mov_threshold, door_particlecoreid, radar_particlecoreid, location]);
         return results.rows[0]; 
     }
     catch(e) {
@@ -409,10 +426,10 @@ async function updateLocationData(deviceid, phonenumber, detection_min, detectio
 }
 
 // Adds a location table entry
-async function createLocation(locationid, deviceid, phonenumber, mov_threshold, still_threshold, duration_threshold, unresponded_timer, auto_reset_threshold, door_stickiness_delay,xethru_heartbeat_number,twilio_number,fallback_phonenumber, unresponded_session_timer) {
+async function createLocation(locationid, deviceid, phonenumber, mov_threshold, still_threshold, duration_threshold, unresponded_timer, auto_reset_threshold, door_stickiness_delay,xethru_heartbeat_number,twilio_number,fallback_phonenumber, unresponded_session_timer, door_particlecoreid, radar_particlecoreid) {
     try {
-        await pool.query("INSERT INTO locations(locationid, deviceid, phonenumber, mov_threshold, still_threshold, duration_threshold, unresponded_timer, auto_reset_threshold, door_stickiness_delay,xethru_heartbeat_number,twilio_number,fallback_phonenumber, unresponded_session_timer) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13)", 
-            [locationid, deviceid, phonenumber, mov_threshold, still_threshold, duration_threshold, unresponded_timer, auto_reset_threshold, door_stickiness_delay,xethru_heartbeat_number,twilio_number,fallback_phonenumber, unresponded_session_timer]);
+        await pool.query("INSERT INTO locations(locationid, deviceid, phonenumber, mov_threshold, still_threshold, duration_threshold, unresponded_timer, auto_reset_threshold, door_stickiness_delay,xethru_heartbeat_number,twilio_number,fallback_phonenumber, unresponded_session_timer, door_particlecoreid, radar_particlecoreid) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15)", 
+            [locationid, deviceid, phonenumber, mov_threshold, still_threshold, duration_threshold, unresponded_timer, auto_reset_threshold, door_stickiness_delay,xethru_heartbeat_number,twilio_number,fallback_phonenumber, unresponded_session_timer, door_particlecoreid, radar_particlecoreid]);
         console.log("New location inserted to Database");
     }
     catch(e) {
@@ -459,6 +476,7 @@ module.exports = {
     saveChatbotSession,
     startChatbotSessionState,
     getMostRecentSessionPhone,
+    getLocationIDFromParticleCoreID,
     getLocationData,
     getLocations,
     updateLocationData,
--- a/db/db.js
+++ b/db/db.js
@@ -68,7 +68,7 @@ async function getMostRecentSession(locationid, client) {
         }
     }
     catch(e){
-        console.log(`Error running the getMostRecentSession query: ${e}`);
+        helpers.log(`Error running the getMostRecentSession query: ${e}`);
     }
 }
 
@@ -86,7 +86,7 @@ async function getSessionWithSessionId(sessionid){
         }
     }
     catch(e){
-        console.log(`Error running the getSessionWithSessionId query: ${e}`);
+        helpers.log(`Error running the getSessionWithSessionId query: ${e}`);
     }
 }
 
@@ -110,7 +110,7 @@ async function getMostRecentSessionPhone(twilioNumber, client){
         }
     }
     catch(e){
-        console.log(`Error running the getMostRecentSessionPhone query: ${e}`);
+        helpers.log(`Error running the getMostRecentSessionPhone query: ${e}`);
     }
 }
 
@@ -126,7 +126,7 @@ async function getHistoryOfSessions(locationid) {
         }
     }
     catch(e) {
-        console.log(`Error running the getHistoryOfSessions query: ${e}`)
+        helpers.log(`Error running the getHistoryOfSessions query: ${e}`)
     }
 }
 
@@ -142,7 +142,7 @@ async function getAllSessionsFromLocation(location) {
         }
     }
     catch(e) {
-        console.log(`Error running the getAllSessionsFromLocation query: ${e}`);
+        helpers.log(`Error running the getAllSessionsFromLocation query: ${e}`);
     }
 }
 
@@ -158,7 +158,7 @@ async function getLastUnclosedSession(locationid) {
         }
     }
     catch(e){
-        console.log(`Error running the getLastUnclosedSession query: ${e}`);
+        helpers.log(`Error running the getLastUnclosedSession query: ${e}`);
     }
 }
 
@@ -176,7 +176,7 @@ async function createSession(phone, locationid, state, client) {
         return results.rows[0];    
     }
     catch(e){
-        console.log(`Error running the createSession query: ${e}`)
+        helpers.log(`Error running the createSession query: ${e}`)
     }
 }
 
@@ -201,7 +201,7 @@ async function updateSessionEndTime(sessionid, client) {
 
     }
     catch(e){
-        console.log(`Error running the updateSessionEndTime query: ${e}`);
+        helpers.log(`Error running the updateSessionEndTime query: ${e}`);
     }
 }
 
@@ -224,7 +224,7 @@ async function updateSessionState(sessionid, state, client) {
         }
     }
     catch(e){
-        console.log(`Error running the updateSessionState query: ${e}`);
+        helpers.log(`Error running the updateSessionState query: ${e}`);
     }
 }
 
@@ -240,7 +240,7 @@ async function updateSentAlerts(location, sentalerts) {
         }
     }
     catch(e){
-        console.log(`Error running the updateSentAlerts query ${e}`);
+        helpers.log(`Error running the updateSentAlerts query ${e}`);
     }
 }
 
@@ -264,7 +264,7 @@ async function updateSessionResetDetails(sessionid, notes, state, client) {
         }
     }
     catch(e){
-        console.log(`Error running the updateSessionResetDetails query: ${e}`);
+        helpers.log(`Error running the updateSessionResetDetails query: ${e}`);
     }
 }
 
@@ -281,7 +281,7 @@ async function updateSessionStillCounter(stillcounter, sessionid,locationid) {
         }
     }
     catch(e){
-        console.log(`Error running the updateSessionStillCounter query: ${e}`);
+        helpers.log(`Error running the updateSessionStillCounter query: ${e}`);
     }
 }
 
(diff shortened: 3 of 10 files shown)


mariocimet:
looking pretty good - could you please add tests for this? One approach would be to add another 'describe' block which replicates the existing tests using the new door sensor endpoint (for which you could define another helper function like im21door() )

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ the code was deployed.
 
 ## [Unreleased]
 
+## [1.3.0] - 2020-12-22
+### Added
+- Endpoints
```
mariocimet, line 15: would wanna be a bit more specific here, like
- Device health endpoint
- IM21 door sensor endpoint

index.js
```diff
@@ -344,6 +345,45 @@ app.post('/api/doorTest', async(req, res) => {
     await handleSensorRequest(locationid)
 });
 
+app.post('/api/door', async(request, response) => {
+    const coreId = request.body.coreid
+    const locationid = db.getLocationFromCoreId(coreId)
```
mariocimet, line 350: probably want an 'await' here

index.js
```diff
@@ -344,6 +345,45 @@ app.post('/api/doorTest', async(req, res) => {
...
+  
+    redis.addIM21DoorSensorData(locationid, doorSignal);
+    handleSensorRequest(locationid)
+    response.status(200).json("OK")
+});
+
+// Handler for device vitals such as wifi strength
+app.post('api/devicevitals', async(request, response) => {
+    const coreId = request.body.coreid
+    const locationid = db.getLocationFromCoreId(coreId)
```
mariocimet, line 377: same - as above, an await could be a good idea

db/db.js
```diff
@@ -379,6 +379,24 @@ async function getLocationData(locationid) {
 
 }
 
+// Retrieves the locationid corresponding to a particle device coreID
+
+async function getLocationFromCoreID(coreID){
+    try{
+        const results = await pool.query("SELECT (locationid) FROM locations WHERE door_coreID = $1 OR radar_coreID = $1", [coreID]);
```
mariocimet, line 386: could you explain this to me a little bit? why this ambiguity between the two types of ID?

trisapeace:
Task linked: [CU-h2y2nt Create new door sensor endpoint](https://app.clickup.com/t/h2y2nt)

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ the code was deployed.
 
 ## [Unreleased]
 
+## [1.3.0] - 2020-12-22
```
trisapeace, line 13: I'd also suggest just leaving this changelog message under the `[Unreleased]` section. The exact version number can be figured out when we do the prod deployment.

IM21DoorStatusEnum.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const IM21_DOOR_STATE = {
+    OPEN: '02',
+    CLOSED: '00',
+    HEARTBEAT_OPEN: '0A',
+    HEARTBEAT_CLOSED: '08',
+    LOW_BATT: '04'
+};
```
trisapeace, line 9: We are moving away from using semi-colons. Could you please take them out of any new lines of code that you write? That will make it easier when we make the linter more strict (which may be coming soon).

IM21DoorStatusEnum.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const IM21_DOOR_STATE = {
```
trisapeace, line 3: Thanks for making this a constant! :smile:  This is makes everything else much more readable.

db/009-add-coreID-to-locations.sql
```diff
@@ -0,0 +1,22 @@
...
+    migrationId := 9;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        -- ADD SCRIPT HERE
+        ALTER TABLE locations ADD COLUMN door_coreID text;
```
trisapeace, line 15: We are pretty inconsistent already with our column naming scheme. However, I don't think we use capital letters anywhere right now. So may I suggest `door_coreid` and `radar_coreid`?

Could you please explain a little more about where the `core` part comes from? Is it the name or are you using it as description?

db/db.js
```diff
@@ -379,6 +379,24 @@ async function getLocationData(locationid) {
 
 }
 
+// Retrieves the locationid corresponding to a particle device coreID
+
+async function getLocationFromCoreID(coreID){
```
trisapeace, line 384: If you're only getting the locationId, then I suggest that your change the name of the function to make that clear `getLocationIdFromCoreID` or something like that.

db/db.js
```diff
@@ -379,6 +379,24 @@ async function getLocationData(locationid) {
 
 }
 
+// Retrieves the locationid corresponding to a particle device coreID
+
```
trisapeace, line 383: Can you please remove this extra line?

db/db.js
trisapeace, line 386: I don't think you need the brackets around `locationid`.

db/db.js
```diff
@@ -379,6 +379,24 @@ async function getLocationData(locationid) {
...
+async function getLocationFromCoreID(coreID){
+    try{
+        const results = await pool.query("SELECT (locationid) FROM locations WHERE door_coreID = $1 OR radar_coreID = $1", [coreID]);
+        if(results == undefined){
+            console.log('Error: No location with associated coreID exists')
+            return null;
+        }
+        else{
+            return results.rows[0];
```
trisapeace, line 392: This sends you back an object with just the `locationid` key, right? It would probably be nicer for your calling function if you just return the value of `results.rows[0].locationid`.

db/db.js
```diff
@@ -379,6 +379,24 @@ async function getLocationData(locationid) {
...
+        if(results == undefined){
+            console.log('Error: No location with associated coreID exists')
+            return null;
+        }
+        else{
+            return results.rows[0]; 
+        }
+    }
+    catch(e){
+        console.log(`Error running the getLocationFromCoreID query: ${e}`);
```
trisapeace, line 396: If you change the function name, please change it here too :smile:

db/redis.js
```diff
@@ -53,6 +53,16 @@ async function addDoorTestSensorData(request, response){
     response.status(200).json("OK")
 }
 
+const addIM21DoorSensorData = (locationid, doorSignal) => {
```
trisapeace, line 56: It looks like the majority of the functions in this file are using the style:
```
function functionName(params) {
...
}
```

instead of 
```
const functionName = (params) => {
...
}
```

Could you please use that instead for this function and `addVitals`?

db/redis.js
```diff
@@ -92,6 +102,8 @@ async function getLatestLocationStatesData(locationid){
 module.exports = {
     addDoorSensorData,
     addDoorTestSensorData,
+    addIM21DoorSensorData,
```
trisapeace, line 105: Thanks for putting this is alphabetical order :smile:

db/redis.js
```diff
@@ -53,6 +53,16 @@ async function addDoorTestSensorData(request, response){
     response.status(200).json("OK")
 }
 
+const addIM21DoorSensorData = (locationid, doorSignal) => {
+    if(doorSignal=="closed"||doorSignal=="open"){
```
trisapeace, line 57: You probably want to use `DOOR_STATE` from `SessionStateDoorEnum.js` here instead of `open` and `closed`.

db/redis.js
```diff
@@ -53,6 +53,16 @@ async function addDoorTestSensorData(request, response){
...
+        client.xadd("door:"+locationid,  "*","signal", doorSignal);
```
trisapeace, line 58: Looks like you have an extra space in the middle of this line.

index.js
trisapeace, line 350: Hmmm, was this smart enough to give you just the `locationid` back from the `rows[0]` in `db.getLocationFromCoreId`? I'm surprised.

index.js
```diff
@@ -344,6 +345,45 @@ app.post('/api/doorTest', async(req, res) => {
...
+        doorSignal = "HeartBeat"
+    }
+  
+    redis.addIM21DoorSensorData(locationid, doorSignal);
+    handleSensorRequest(locationid)
+    response.status(200).json("OK")
+});
+
+// Handler for device vitals such as wifi strength
+app.post('api/devicevitals', async(request, response) => {
```
trisapeace, line 375: The other endpoint definitions all start with a forward slash:
`app.post('/api/devicevitals', ...` 
would be more consistent.

index.js
```diff
@@ -344,6 +345,45 @@ app.post('/api/doorTest', async(req, res) => {
     await handleSensorRequest(locationid)
 });
 
+app.post('/api/door', async(request, response) => {
```
trisapeace, line 348: Checkout this Express Validatior that @mariocimet started using in Buttons: https://github.com/bravetechnologycoop/BraveButtons/blob/master/server/server.js#L235

It allows for good error handling if the parameters that you expect, like `body.coreid`, are actually in the `request` object or not. I think this is something that we should start using for all our endpoints.

CHANGELOG.md
trisapeace: Please also include the ClickUp number.

index.js
```diff
@@ -344,6 +345,45 @@ app.post('/api/doorTest', async(req, res) => {
...
+    const coreId = request.body.coreid
+    const locationid = db.getLocationFromCoreId(coreId)
+  
+    const message = JSON.parse(request.body.data)
+    const signal = message.data
+    var doorSignal
+  
+    if (signal==IM21_DOOR_STATUS.OPEN){
```
trisapeace, line 356: For `open` and `closed`, you should probably use `DOOR_STATE` from `SessionStateDoorEnum.js`.

For `LowBatt` and `HeartBeat`, it doesn't look like these are being used for anything. You're just sending `doorSignal` to `addIM21DoorSensorData` which only cares if it's `open` or `closed`. I suggest that you just remove lines 362-367. In these cases, it will just send `undefined` as `doorSignal` to `addIM21DoorSensorData` which will then do nothing. Would that do what you want here?

trisapeace (commented):
Good start!

I agree with Mario, some tests for the changes in `index.js` would be much appreciated.

db/db.js
sampathsatti: each locationID uses two argon devices, one for the radar and the other for the door. Each has its own unique coreID

db/009-add-coreID-to-locations.sql
sampathsatti: `coreid` is a term used by particle for the UUID of their hardware devices. The radar, and door sensor Argons each have their unique coreid, hence the column names.

I will make it lowercase.

index.js
sampathsatti: So my rationale for doing it this way was thus - 

- The door:locationID key on redis is the logical place to add data from the sensor such as low battery/heartbeat - which are definitely data points that we want to log. 

- However, the current statemachine would need to be modified before we can have entries other than `open` or `closed` in this stream. 

- as I was implementing this, I had to choose between ignoring any signal data other than `00` or `02` **or** parse it and then choose to not add it to the stream. Given the above two points, I chose the latter as it signalled an intention to support using low battery/heartbeat data when we overhaul the state machine. 

I do see your point about it being a stub that leads to nowhere, and I suspect that is frowned upon in code? If so, I'll delete these lines...

index.js
sampathsatti: This is great! Thanks for sharing Theresa.

I think I'll start doing this in another pull request, since it might need a few other helper functions on the db (and possibly more columns)?

Additionally, the xethru endpoint will also need similar validation and we might as  well tackle all the endpoints at once.

I've created a clickup task

https://app.clickup.com/t/hx0jjc

db/009-add-coreID-to-locations.sql
trisapeace: Would it make sense then to include `argon` in the name too? Like `door_argoncoreid` and `radar_argoncoreid`. That might also stop the confusion with the `getLocationFromCoreID` (renamed to `getLocationFromArgonCoreID`) function.

index.js
trisapeace: No, I think that you have a good point. If it'll help us remember to do it later and it's accurate, then it seems OK to leave it there. Thanks for explaining.

index.js
sampathsatti: Didn't test this, you have a good point. I will modify the function to return the `location `row, and then select the `locationid `column using `location.locationid`

db/009-add-coreID-to-locations.sql
sampathsatti: I'm not sure if it helps to have `argon `in the name, since the only `coreID` that we use pertains to Particle devices (and it encompasses all devices in their ecosystem, including photon and boron devices which we might be using in the future)

CHANGELOG.md
```diff
@@ -10,19 +10,19 @@ the code was deployed.
 
 ## [Unreleased]
 
-## [1.3.0] - 2020-12-22
-### Added
-- Endpoints
+### Added 2020-12-22
```
trisapeace, line 13: Thanks for changing this. But the date should now be removed too. The dates are only for releases.

index.js
```diff
@@ -343,21 +344,22 @@ app.post('/api/doorTest', async(req, res) => {
...
     if (signal==IM21_DOOR_STATUS.OPEN){
-        doorSignal = "open"
+        doorSignal = SESSIONSTATE_DOOR.OPEN
     }
     else if (signal==IM21_DOOR_STATUS.CLOSED){
-        doorSignal = "closed"
+        doorSignal = SESSIONSTATE_DOOR.CLOSED
     }
     else if (signal==IM21_DOOR_STATUS.LOW_BATT){
         doorSignal = "LowBatt"
```
trisapeace, line 365: Should `"LowBatt"` and `"HeartBeat`" be added to `SESSIONSTATE_DOOR` enum since you plan to leave them in?

db/009-add-coreID-to-locations.sql
```diff
@@ -0,0 +1,22 @@
...
+    migrationId := 9;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        -- ADD SCRIPT HERE
+        ALTER TABLE locations ADD COLUMN door_coreid text;
```
trisapeace, line 15: Then how about `door_particlecoreid` and `sensor_particlecoreid`?

I find that there are already a lot of columns in the code base which take a lot of effort to understand. For example, I need to basically write myself out a diagram of all the phone numbers each time to figure out which one is doing what. And I'd like to avoid that in the future, especially for things related to a specific hardware, it makes sense to me to be specific about it. Would we use this same column if we swap out to a different door sensor? I'm thinking probably not. However, if we would, then naming it something more generic like just `door_id` would be more flexible.

db/db.js
```diff
@@ -379,6 +379,24 @@ async function getLocationData(locationid) {
 
 }
 
+// Retrieves the locationid corresponding to a particle device coreID
+
+async function getLocationFromCoreID(coreID){
+    try{
+        const results = await pool.query("SELECT (locationid) FROM locations WHERE door_coreid = $1 OR radar_coreid = $1", [coreID])
```
trisapeace, line 386: OK, if you'd like to leave this one as getting the `Location`, then it would be better if this were a `SELECT *` instead of just a `SELECT locationid` and then map the resulting rows in `Location` objects from `Location.js`.

In fact, this reminds me that `Location.js` should be updated to include the two new columns and then you should check on all the DB functions that use it to make sure that they all take those new columns into consideration.

db/db.js
```diff
@@ -379,6 +379,44 @@ async function getLocationData(locationid) {
 
 }
 
+
+//Retrieves the partcle door coreID corresponding to a locationid
+//This is required for ServerTest.js
+
+async function getDoorParticleCoreIDFromLocationID(locationid){
+    try{
+        const results = await pool.query("SELECT (door_particlecoreid) FROM locations WHERE locationid = $1", [locationid])
```
trisapeace, line 388: Do you need these brackets around `(door_particlecoreid)`?

db/db.js
```diff
@@ -379,6 +379,44 @@ async function getLocationData(locationid) {
 
 }
 
+
+//Retrieves the partcle door coreID corresponding to a locationid
+//This is required for ServerTest.js
+
+async function getDoorParticleCoreIDFromLocationID(locationid){
```
trisapeace, line 386: Where is this used?

db/db.js
```diff
@@ -379,6 +379,44 @@ async function getLocationData(locationid) {
...
+    try{
+        const results = await pool.query("SELECT (door_particlecoreid) FROM locations WHERE locationid = $1", [locationid])
+        if(results == undefined){
+            console.log('Error: No row with the locationid exists')
+            return null
```
trisapeace, line 391: The reason that I was wondering where it was used is because I was planning to check whether the caller was reasonably handling a `null` return value. So if you end up adding it to be used somewhere, please think about this too ;-)

db/db.js
```diff
@@ -379,6 +379,44 @@ async function getLocationData(locationid) {
...
+    catch(e){
+        console.log(`Error running the getDoorParticleCoreIDFromLocationID query: ${e}`)
+    }
+}
+
+// Retrieves the locationid corresponding to a particle device coreID
+
+async function getLocationIDFromParticleCoreID(coreID){
+    try{
+        const results = await pool.query("SELECT (locationid) FROM locations WHERE door_particlecoreid = $1 OR radar_particlecoreid = $1", [coreID])
```
trisapeace, line 406: Do you need these brackets around `(locationid)`?

index.js
```diff
@@ -342,7 +344,48 @@ app.post('/api/doorTest', async(req, res) => {
     const {locationid} = req.body;
     await redis.addDoorTestSensorData(req, res);
     await handleSensorRequest(locationid)
-});
+})
+
+app.post('/api/door', async(request, response) => {
+    const coreId = request.body.coreid
+    const location = await db.getLocationIDFromParticleCoreID(coreId)
+    const locationid = location.locationid
```
trisapeace, line 352: This will cause an error if `locationid` id `null`. We probably just want to log the null and return a decent error code in this case.

db/db.js
```diff
@@ -379,6 +379,44 @@ async function getLocationData(locationid) {
...
+        if(results == undefined){
+            console.log('Error: No location with associated coreID exists')
+            return null
+        }
+        else{
+            return results.rows[0]
```
trisapeace, line 412: I still find this return value inconsistent with the name of the function. I would expect either:

A) The function is called `getLocationFromParticleCoreID`, the query is `SELECT * FROM locations...`, and the return value is the whole first row `results.rows[0]`.

OR

B) The function is called `getLocationidFromParticleCoreID`, the query is `SELECT locationid FROM locations...`, and the return value is just the valuve `results.rows[0].locationid`

db/db.js
```diff
@@ -379,6 +379,44 @@ async function getLocationData(locationid) {
...
+        }
+    }
+    catch(e){
+        console.log(`Error running the getLocationFromCoreID query: ${e}`)
```
trisapeace, line 416: The function name in this log message should match the function's name.

db/db.js
```diff
@@ -397,10 +435,10 @@ async function getLocations() {
...
+async function updateLocationData(deviceid, phonenumber, detection_min, detection_max, sensitivity, noisemap, led,  rpm_threshold, still_threshold, duration_threshold, mov_threshold, door_particlecoreid, radar_particlecoreid, location) {
     try {
-        let results = await pool.query("UPDATE locations SET deviceid = $1, phonenumber = $2, detectionzone_min = $3, detectionzone_max = $4, sensitivity = $5, noisemap = $6, led = $7, rpm_threshold = $8, still_threshold = $9, duration_threshold = $10, mov_threshold = $11 WHERE locationid = $12 returning *", 
-            [deviceid, phonenumber, detection_min, detection_max, sensitivity, noisemap, led, rpm_threshold, still_threshold, duration_threshold, mov_threshold, location]);
+        let results = await pool.query("UPDATE locations SET deviceid = $1, phonenumber = $2, detectionzone_min = $3, detectionzone_max = $4, sensitivity = $5, noisemap = $6, led = $7, rpm_threshold = $8, still_threshold = $9, duration_threshold = $10, mov_threshold = $11, door_particlecoreid = $12, radar_particlecoreid = $13 WHERE locationid = $14 returning *",
```
trisapeace, line 440: That's really cool. I didn't know about that `returning *` syntax at the end.

index.js
```diff
@@ -342,7 +344,48 @@ app.post('/api/doorTest', async(req, res) => {
...
+    redis.addIM21DoorSensorData(locationid, doorSignal)
+    handleSensorRequest(locationid)
+    response.status(200).json("OK")
+})
+
+// Handler for device vitals such as wifi strength
+app.post('/api/devicevitals', async(request, response) => {
+    const coreId = request.body.coreid
+    const location = await db.getLocationIDFromParticleCoreID(coreId)
+    const locationid = location.locationid
```
trisapeace, line 380: Also need some error handling here for when `location` is `null`.

test/serverTest.js
```diff
@@ -9,11 +9,14 @@ const redis = imports.redis
...
 const XETHRU_STATE = require('../SessionStateXethruEnum.js')
 const MOV_THRESHOLD = 17
+const IM21_DOOR_STATUS = require('../IM21DoorStatusEnum')
+const ST_DOOR_STATUS = require('../SessionStateDoorEnum')
 
 chai.use(chaiHttp)
 
 const testLocation1Id = 'TestLocation1'
 const testLocation1PhoneNumber = '+15005550006'
+const door_coreID = 'door_particlecoreid'
```
trisapeace, line 19: What do you need this for?

test/serverTest.js
```diff
@@ -159,4 +176,92 @@ describe('ODetect server', () => {
             expect(sessions[0].od_flag).to.equal(1)
         })
     })
+
+    describe('POST request radar and door events with mock im21 door sensor', () => {
```
trisapeace, line 180: These look so nice and clear. Good job!

trisapeace (commented):
Lookin' good. I added some more comments though :wink: Nothing major.

db/db.js
mariocimet: I think this is no longer used - it used to be part of the mock im21Door function - it would take a locationid as a parameter and then use that to find a doorCoreID to include in the post request, but I took that out in the last changes.

test/serverTest.js
mariocimet: this is a parameter in the im21Door mock helper

index.js
mariocimet: good call

db/redis.js
```diff
@@ -53,6 +54,16 @@ async function addDoorTestSensorData(request, response){
     response.status(200).json("OK")
 }
 
+async function addIM21DoorSensorData(locationid, doorSignal){
+    if(doorSignal==SESSIONSTATE_DOOR.CLOSED||doorSignal==SESSIONSTATE_DOOR.OPEN){
+        await client.xadd("door:"+locationid,  "*","signal", doorSignal)
+    }
+}
+
+async function addVitals(locationid, signalStrength, cloudDisconnects){
```
trisapeace, line 63: Does this need to be an `async` function if there aren't any `await`s inside?

db/redis.js
```diff
@@ -53,6 +54,16 @@ async function addDoorTestSensorData(request, response){
     response.status(200).json("OK")
 }
 
+async function addIM21DoorSensorData(locationid, doorSignal){
+    if(doorSignal==SESSIONSTATE_DOOR.CLOSED||doorSignal==SESSIONSTATE_DOOR.OPEN){
```
trisapeace, line 58: I would be eternally grateful for spaces around the operators `==` and `||` in this line :pray:

package-lock.json
```diff
@@ -96,96 +96,11 @@
         }
       }
     },
-    "@sentry/apm": {
```
trisapeace, line 99: Why did `package-lock.json` change but `package.json` didn't? Had this just fallen behind?

trisapeace (approved):
Looks good! Glad to hear that the testing went well. Was it useful for testing to use `helpers.log` instead of `console.log`? The reason I ask is because it's going to make my PR where I make the same change all over the place now quite messy. Not the end of the world though.

PR #67, opened 2021-01-13
Add express validator for use on post requests (CU-hx0jjc)

risatronic:
- Add express validator to use on post requests
- Updated `db.js` to use PG_DATABASE for env variable as my database name and username are different

Did not add validator to Twilio endpoint as that will be using `brave-alert-lib`, but can add or remove it from any routes as people see fit. I also wasn't sure how much I should be checking in there because only Twilio seems to be using more than just one thing pulled from the body aside from smart things but that validation happens within its own function & as mentioned Twilio validation is covered in the `brave-alert-lib` already but am very open to feedback :)


trisapeace:
Task linked: [CU-hx0jjc Add Express validations to endpoints](https://app.clickup.com/t/hx0jjc)

risatronic:
I also just noticed that there are more post routes in a different PR so maybe I should wait until that is merged and then revisit this 🤔

db/db.js
```diff
@@ -10,7 +10,7 @@ const pool = new pg.Pool({
     host: helpers.getEnvVar('PG_HOST'),
     port: helpers.getEnvVar('PG_PORT'),
     user: helpers.getEnvVar('PG_USER'),
-    database: helpers.getEnvVar('PG_USER'),
+    database: helpers.getEnvVar('PG_DATABASE'),
```
trisapeace, line 13: Can you take a look around to see where else `PG_USER` is being used for both? There is probably stuff in the README and maybe in `setup_postgres.sh` that need to be changed to do this too.

index.js
```diff
@@ -325,24 +326,63 @@ if(!helpers.isTestEnvironment()){
...
-    smartapp.handleHttpCallback(req, res);
-});
+app.post('/api/st', Validator.body(['lifecycle']).exists(), function(req, res, next) {    // eslint-disable-line no-unused-vars -- next might be used in the future
+    try {
+        const validationErrors = Validator.validationResult(req)
+  
+        if(validationErrors.isEmpty()){
+            smartapp.handleHttpCallback(req, res)
+        } else {
+            helpers.log(`Bad request, parameters missing ${JSON.stringify(validationErrors)}`)
```
trisapeace, line 336: Does the validationErrors object already have a nice string to display? I wonder if we even need to put the `Bad request, parameters missing` part around it.

index.js
```diff
@@ -325,24 +326,63 @@ if(!helpers.isTestEnvironment()){
...
 
 // Handler for income XeThru POST requests
-app.post('/api/xethru', async (req, res) => {
-    // eslint-disable-next-line no-unused-vars -- might be useful in the future to know what all we have access to in the body
-    const {deviceid, locationid, devicetype, state, rpm, distance, mov_f, mov_s} = req.body;
-
-    redis.addXeThruSensorData(req, res);
-    handleSensorRequest(locationid);
-});
+app.post('/api/xethru', Validator.body(['locationid']).exists(), async (req, res) => {
```
trisapeace, line 347: This also needs `state`, `rpm`, `distance`, `mov_f`, and `mov_s` in the `body` as used in `redis.addXeThruSensorData`.

trisapeace: Actually, now that I'm looking at this... if you're up for even more refactoring, it would be really nice if these route functions were the only ones to know about the request and response objects. To do this, functions that are called from them, like `redis.addXeThruSensorData(req, res)` would need to change to take in `locationid, state, rpm, distance, mov_f, mov_s` as separate parameters and not send back the `200` response.

This would be good for a bunch of reasons...
1. It would make it easier to see exactly what is being used and why it's being validated here.
2. It would make the called functions easier to test because you wouldn't need to match the structure of the request object or mock a response object
3. Would be generally good separation of concerns (i.e. why should the Redis module need to care that the data it was getting came from an API endpoint)

However, this change isn't strictly related to your Task. So it's up to you. It would be great if you could do the easy ones, but if it starts being tricky (like for how `smartapp.handleHttpCallback` sends the request off to `signature.signatureIsVerified` for even more processing), then don't bother with those ones.

index.js
```diff
@@ -325,24 +326,63 @@ if(!helpers.isTestEnvironment()){
...
-    await redis.addDoorTestSensorData(req, res);
-    await handleSensorRequest(locationid)
-});
+app.post('/api/doorTest', Validator.body(['locationid']).exists(), async(req, res) => {
+    try {
+        const validationErrors = Validator.validationResult(req)
+  
+        if(validationErrors.isEmpty()){
+            const {locationid} = req.body
+            await redis.addDoorTestSensorData(req, res)
```
trisapeace, line 374: This function also uses `req.body.signal`

index.js
```diff
@@ -633,9 +671,6 @@ let server;
 server = app.listen(8080);
 console.log('brave server listening on port 8080')
 
-
```
trisapeace, line 636: Thank you for removing unnecessary newlines!

index.js
```diff
@@ -325,24 +326,63 @@ if(!helpers.isTestEnvironment()){
 }
 
 // Handler for income SmartThings POST requests
-app.post('/api/st', function(req, res, next) {    // eslint-disable-line no-unused-vars -- next might be used in the future
-    smartapp.handleHttpCallback(req, res);
-});
+app.post('/api/st', Validator.body(['lifecycle']).exists(), function(req, res, next) {    // eslint-disable-line no-unused-vars -- next might be used in the future
```
trisapeace, line 329: Could you please add some tests to show that these Validators work as expected?

index.js
```diff
@@ -325,24 +326,63 @@ if(!helpers.isTestEnvironment()){
 }
```
trisapeace, line 326: The current `if`-statement in `app.get('/logout'` can probably be replaced by a Validator check.

The current `if`-statment in `app.get('/dashboard'` can probably be replaced by a Validator check.

`app.get('/dashboard/:locationId'` could use the Validator to check for `req.session.user` and `req.cookies.user_sid`. No need for it to validate `locationId` though since we wouldn't get into this route without one.

trisapeace (commented):
On the right track! Thanks for doing this :smile:

index.js
risatronic: Definitely something I can look into, but yeah as you point out I'm not sure I want to mess around with functions from other libraries where stuff is obfuscated like in `handleHttpCallback`

index.js
risatronic: I have a question about this in the dev channel because I am not sure what the header authorisation token for smartapp is or how to find it, it gets to that function though meaning it passes the validator which is also in that message so I'm not sure if that would be testing enough to check?

https://bravely.slack.com/archives/CE32NB85N/p1610578347334700

index.js
risatronic: I only applied it to post routes as I wasn't sure - if I do that does `sessionChecker` on line 89 still need to exist or could I also just use that on `app.get('/'`?

db/db.js
risatronic: Sure can! :) I also just noticed that the env file @mariocimet shared with me doesn't actually include `$PG_CONNECTION_STRING` but it's mentioned in the readme, and I don't know that it's actually being used anywhere

risatronic: It is used after `-d` in `setup_postgresql_local.sh` though and the current env info/template is in 1password so I would also need to update that to be clear we'd need a `$PG_DATABASE` environment variable 🤔

risatronic: there is no `.env.example` file as far as I know is why

index.js
risatronic: I'm not sure, I will set it off in postman and see 😅 for this one the error that comes back is re smartapp looking for an authorisation header that I don't have so I didn't notice it as it doesn't make it to the `else`, but this came from where `Validator` is used in Buttons so I think it might be needed 🤔 I'll check though

risatronic: ```2021-01-15T00:14:07.178Z - Bad request, parameters missing {"errors":[{"msg":"Invalid value","param":"lifecycle","location":"body"}]}```

risatronic: I think missing is more descriptive than invalid value personally, especially because lifecycle wants to be `===` to something and it being wrong is different than not existing at all, but am happy to do whatever!

risatronic: (that's what `.exists()` comes back with)

index.js
risatronic: line *94 I was looking at a different branch locally

index.js
risatronic: Also this lines up with the linting because it doesn't like that random object that is just `= req.body` and then most of what's inside it goes unused

db/db.js
trisapeace: `$PG_CONNECTION_STRING` was removed by @mariocimet  when he added in `PG_HOST`, `PG_USER`, etc. I guess he forgot to update the README.

Yup, you can update the `.env` in 1Password to include it. It can have extra things that aren't used yet, so you can add it there now so that you don't forget later if you want.

index.js
trisapeace: Yup, looks like they purposely are not super explicit in their messages:
https://express-validator.github.io/docs/custom-error-messages.html

index.js
trisapeace: Can you stub out the smartapp function that looks for the token?

index.js
trisapeace: Interesting. Good point. I hadn't looked at `sessionChecker` too closely. It actually seems to me like it's doing something strange. I would have expected it to redirect to `/login` if the session was missing and otherwise to go on to `next()`. If I'm understanding this correctly. If it did that, then it could also be used in `/dashboard` and `/dashboard/:locationId`... I think.

trisapeace:
Also, yes, it's probably worth waiting to also add this to Sampath and Mario's PR once it's merged. I think they are close.

index.js
risatronic: @trisapeace can you please confirm that what we'd be wanting in this case would be something like

```
if(validationErrors.isEmpty()){
    const { locationid, state, rpm, distance, mov_f, mov_s } = req.body

    redis.addXeThruSensorData(locationid, state, rpm, distance, mov_f, mov_s)
    handleSensorRequest(locationid)

    res.status(200).send()
} 
```

db/db.js
risatronic: My current plan for this is to be converting it back to `PG_USER` before merging because I'm not sure when the best time to update that across things will be, if anyone thinks it should actually be now though I can try to figure that out. I kept forgetting the other day when pushing linting things and it was breaking Travis 😩

risatronic: nvm, it has now been added to the `.travis.yml` file and is in the `.env` on 1password per discussion earlier today :) and after running the build only 4 times it decided to pass with that change 🤨 Better than an automatic break though I guess haha

risatronic: Will update other instances as needed 👍

index.js
risatronic: ^I'm just looking at this again now, did that seem to be a better idea than validator for those get requests? If so I can shuffle it around, I think it's kinda left over from when `/dashboard` didn't exist, it says middleware by it though so maybe it can just go away 🤔

risatronic: @trisapeace @mariocimet  I finally have the rest of things working so was wondering about this?

index.js
risatronic: Is it just this function you were wanting added tests for? If the express thing doesn't work the existing tests fail is how it's otherwise working

risatronic:
this is driving me literally insane because I can't tell if tests are janky because of what I did or...? They all pass locally so I can't replicate it on my machine, or at least how to do that isn't immediately apparent. The most times of running them it's taken before to get a pass is 4, I've done it like 10+ times now and always some combo of the last few is failing so if anyone has any insight I'd greatly appreciate it 😭

risatronic:
@mariocimet

@trisapeace took a look at failing Travis builds with me and suggested adding awaits into the api/xethru endpoint (more info here: https://app.clickup.com/t/hx0jjc?comment=168274643) which worked to make it happy but we weren't sure if that would be ok for real life situations, ie. whether we should leave this as updated or amend the tests

I seem to remember a conversation around experimenting with sleeps for the tests also but couldn't find a task for it so I'm not sure one was made, if it was then this could possibly resolve that 🤔

index.js
risatronic: I have added tests for this but Samsung itself throws an error on top of returning a 401 (which the test does look for and gets), is that a problem? I'm not sure how to get around it

index.js
```diff
@@ -94,20 +95,20 @@ app.use((req, res, next) => {
...
 // middleware function to check for logged-in users
 var sessionChecker = (req, res, next) => {
-    if (req.session.user && req.cookies.user_sid) {
-        res.redirect('/dashboard');
+    if (!req.session.user && !req.cookies.user_sid) {
+        res.redirect('/login')
     }
     else {
-        next();
+        next()
```
risatronic, line 102: having this the other way around said too many redirects when I tested it, so did a negative check instead

index.js
```diff
@@ -94,20 +95,20 @@ app.use((req, res, next) => {
...
     }
 };
 
 app.get('/', sessionChecker, (req, res) => {
-    res.redirect('/login');
+    res.redirect('/dashboard');
```
risatronic, line 107: updated this because with sessionChecker in place, if someone is logged in and tries to go to /, they should be redirected to the landing page vs login page

index.js
```diff
@@ -94,20 +95,20 @@ app.use((req, res, next) => {
...
 });
 
 app.route('/login')
-    .get(sessionChecker, (req, res) => {
+    .get((req, res) => {
```
risatronic, line 111: removed sessionChecker from this because it makes no sense with new logic

index.js
```diff
@@ -327,71 +318,131 @@ if(!helpers.isTestEnvironment()){
...
+            helpers.log(`Bad request, parameters missing ${JSON.stringify(validationErrors)}`)
+            res.status(400).send()
         }
+    } catch (err) {
+        helpers.log(err)
+        res.status(500).send()
+    }
+})
+
+app.post('/api/door', Validator.body(['coreid']).exists(), async(request, response) => {
```
mariocimet, line 378: as noted in our call could be a good idea to check for the fields we use that it gets from request.body.data

risatronic: Updated :)

mariocimet: looks good! this is good to go then

mariocimet: I think I'd prefer a squash commit with a nice message? @trisapeace you agree?

mariocimet (approved):

index.js
risatronic: Haha yeah that was definitely my plan cause it's a bit of a mess on that front

trisapeace:
@risatronic Could you please resolve the conflicts with the base branch so that I can see exactly what will end up being merged? Thanks!

risatronic:
> @risatronic Could you please resolve the conflicts with the base branch so that I can see exactly what will end up being merged? Thanks!

can do, I didn't realise there were conflicts but I guess that makes sense with the deployment & rebase haha

risatronic:
@trisapeace done :)

.travis.yml
```diff
@@ -38,6 +38,8 @@ matrix:
         - PG_PORT=5433
         - PG_HOST=localhost
         - REDIS_CLUSTER_IP_TEST=localhost
+        - PG_DATABASE=brave
```
trisapeace, line 41: After your merge, the `PG_DATABASE` value is in here twice.

index.js
```diff
@@ -98,20 +99,20 @@ app.use((req, res, next) => {
 
 // middleware function to check for logged-in users
 var sessionChecker = (req, res, next) => {
-    if (req.session.user && req.cookies.user_sid) {
-        res.redirect('/dashboard');
+    if (!req.session.user && !req.cookies.user_sid) {
```
trisapeace, line 102: I think that this should be an OR `||` instead of an AND `&&`. Since if they are missing either of these things, we want them to sign in again.

In general, the opposite of `A && B` is `!A || !B`.

index.js
trisapeace: Thanks for these explanatory notes :smile:

index.js
```diff
@@ -327,71 +318,131 @@ if(!helpers.isTestEnvironment()){
...
+    try {
+        const validationErrors = Validator.validationResult(req)
+
+        if(validationErrors.isEmpty()){
+            // eslint-disable-next-line no-unused-vars -- might be useful in the future to know what all we have access to in the body
+            const { deviceid, locationid, devicetype, state, rpm, distance, mov_f, mov_s } = req.body;
+
+            await redis.addXeThruSensorData(locationid, state, rpm, distance, mov_f, mov_s);
+            await handleSensorRequest(locationid);
+            res.status(200).send()
```
trisapeace, line 347: Looks like there used to be a `.json("OK")` instead of `send()` for this response before. It probably doesn't matter, but to be on the safe side, you could add it back in.

index.js
```diff
@@ -327,71 +318,131 @@ if(!helpers.isTestEnvironment()){
...
-            doorSignal = "LowBatt"
-        }
-        else if (signal==IM21_DOOR_STATUS.HEARTBEAT_OPEN || signal==IM21_DOOR_STATUS.HEARTBEAT_CLOSED){
-            doorSignal = "HeartBeat"
+        if(validationErrors.isEmpty()){
+            const { locationid, signal } = req.body
+
+            await redis.addDoorTestSensorData(locationid, signal)
+            await handleSensorRequest(locationid)
+            res.status(200).send()
```
trisapeace, line 367: This also used to have a `.json("OK")` instead of `send()`

index.js
```diff
@@ -327,71 +318,131 @@ if(!helpers.isTestEnvironment()){
...
+            helpers.log(`Bad request, parameters missing ${JSON.stringify(validationErrors)}`)
+            res.status(400).send()
         }
+    } catch (err) {
+        helpers.log(err)
+        res.status(500).send()
+    }
+})
+
+app.post('/api/door', Validator.body(['coreid', 'data.*.deviceid', 'data.*.data', 'data.*.control']).exists(), async(request, response) => {
```
trisapeace, line 378: Where are `data.*.deviceid', 'data.*.data', 'data.*.control'` being used?

I see `request.data.data` on line 391, but don't see the others or the reason for the stars. Can you please help me understand?

index.js
```diff
@@ -327,71 +318,131 @@ if(!helpers.isTestEnvironment()){
...
-        response.status(400).json('No location for CoreID')
-    }else {
-        const data = JSON.parse(request.body.data)
-        const signalStrength = data.device.network.signal.strength
-        const cloudDisconnects = data.device.cloud.disconnects
-
-        redis.addVitals(locationid, signalStrength, cloudDisconnects)
-        response.status(200).json("OK")
-
+app.post('/api/devicevitals', Validator.body(['coreid']).exists(), async(request, response) => {
```
trisapeace, line 421: Looks like the Validator could also check for `data`, `data.device.network.signal.strength` and `data.device.cloud.disconnects`. Or was this left out on purpose?

index.js
trisapeace: Looks like you figured it out. Thanks for adding the tests. They look good.

index.js
risatronic: I had actually made a note to ask about that and then forgot 😅

index.js
risatronic: This came from a discussion with Mario about checking the content of data which is an object, this is the syntax for checking object keys within body for the express validator :)

index.js
risatronic: Not purposefully left out that I know of, @mariocimet only mentioned checking inside data for the purpose of the one route where I have `data.*.thing` in it but that's easy to add!

index.js
risatronic: In addition to send or as a replacement? Can do either way :)

.travis.yml
risatronic: Will fix that, thanks!

index.js
trisapeace: It looks like our other functions are using just `json("OK")`. I guess it doesn't need a `send()` if you do that :shrug:

index.js
trisapeace: That makes sense for `request.data.data`, but I just don't see where we are using the other object keys.

index.js
risatronic: Okay sounds good haha

index.js
risatronic: @mariocimet mentioned to look inside data for I think `deviceid` specifically so I checked what we were using in the tests to determine what a valid request would entail was my thinking

trisapeace: @mariocimet , it doesn't look like our code would break if the door didn't send us the deviceid (or the others). What are you hoping to catch with this?

risatronic: Yeah now that it's mentioned isn't deviceid one of the columns we were wanting to remove too? 😅 I think it was more a point of consistency in any case re checking for things not explicitly mentioned in the route handlers

index.js
risatronic: In this case would I need to explicitly check for data as if it didn't exist the other two wouldn't either?

index.js
risatronic: Oh, now that I'm looking at this in my text editor I remember why I did this, it was for consistency since the other things for 400 and 500 are formatted this way in the else & catch

```
} else {
  helpers.log(`Bad request, parameters missing ${JSON.stringify(validationErrors)}`)
  res.status(400).send()
  }
} catch (err) {
   helpers.log(err)
   res.status(500).send()
}
```

index.js
risatronic: mm ok, I am looking at this too and those fields aren't included in the tests so it would make tests fail and I'm not sure what the values of them are supposed to be to be valid if I were to add them 🤔

from `serverTest.js`:

```
it('should return 400 to a device vitals signal with an unregistered coreID', async () => {
  let response = await chai.request(server).post('/api/devicevitals').send({coreid: 'unregisteredID',data:{deviceid:'5A:2B:3C',data: 'closed',control: 'AA'}})
  expect(response).to.have.status(400)
})
```

index.js
trisapeace: That also makes sense. I'd just say that taking out the `.json("OK")` will require more testing because you've changed the conversation between the devices. You're choice :smile:

index.js
trisapeace: I'd prefer the philosophy that we want to be as strict as we need to be and no stricter. If they want to change their API to return different values, we should only complain if they were values that we actually care about.

index.js
trisapeace: To find out some valid values, you could try asking @sampathsatti . Or just try it out - you could add a log message, deploy to dev, and then watch what comes in.

trisapeace: It would also be a good opportunity to write some tests that will fail when those values aren't provided.

index.js
risatronic: Yeah I will probably just put it back for now just on the off chance 😅

index.js
mariocimet: yeah it's just data.data

mariocimet: those values are in the IM21 Door Status enum, we already vary it in the tests using the enum for open vs. closed signals

`'use strict'

const IM21_DOOR_STATE = {
    OPEN: '02',
    CLOSED: '00',
    HEARTBEAT_OPEN: '0A',
    HEARTBEAT_CLOSED: '08',
    LOW_BATT: '04'
}`

index.js
mariocimet: I was just referring to the request.data.data bit which we actually use to tell us the door status.

risatronic: oh ok, so should I change it just to look for that one?

risatronic: My question more specifically too I think is that you can't look at `request.data` as far as I can see from the validator documentation, which is why I was looking for the data.data bit of `request.body`, is that incorrect?

"Creates a validation chain for one or more fields. They may be located in any of the following request objects:

req.body
req.cookies
req.headers
req.params
req.query"

trisapeace (approved):
Approving this with the assumption that the changes that we just went over in review/pairing with Mario will be included. Good job!

PR #68, opened 2021-01-15
Md try linting

risatronic:
test


risatronic:
> > I wonder how GitHub deals with these comments on a particular commit. It would have been good for this to be in PR even if it never gets merged just so we have record of the comments even after the branch is gone.
> 
> they might stay, I will take a look 🤔

I'm replying from the PR right now where I can see the comments :)

PR #69, opened 2021-01-15
Remove incorrect commas

risatronic:
Very small fix as this was noticed when trying out the proposed eslint configuration and it formatted it really hinkily. There is no task for clickup in this & not sure that there really needs to be??

### DIFF
--- a/Location.js
+++ b/Location.js
@@ -1,21 +1,23 @@
 class Location {
+    
     constructor(locationid, displayName, deviceid, phonenumber, detectionzoneMin, detectionzoneMax, sensitivity, led, noisemap, movThreshold, durationThreshold, stillThreshold, rpmThreshold, xethruSentAlerts, xethruHeartbeatNumber) {
-        this.locationid = locationid,
-        this.displayName = displayName,
-        this.deviceid = deviceid,
-        this.phonenumber = phonenumber,
-        this.detectionzoneMin = detectionzoneMin,
-        this.detectionzoneMax = detectionzoneMax,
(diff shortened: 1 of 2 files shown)


trisapeace (approved):
Looks good to me.

risatronic:
I am really unclear on why this is failing when all I did was merge in what was updated in `devenv` and remove more commas :(

trisapeace:
I suggest that you rebase your branch off of `master` and then you might have a better view into what the tests are trying to do. I agree that the failure messages don't seem to make sense if the tests are passing on `master`.

risatronic:
> I suggest that you rebase your branch off of `master` and then you might have a better view into what the tests are trying to do. I agree that the failure messages don't seem to make sense if the tests are passing on `master`.

Sorry, they are passing on `devenv` with @mariocimet's merge and I did merge that into this and they're not passing here so I'm not sure I fully understand. I'm tempted to just close this and re-do it since it's such a small fix if that's what you mean?

trisapeace:
Haha, you're right that I did mean `devenv`, not `master`. I was meaning actually use `git rebase devenv` locally on your branch and then force pushing. But you're right, just redoing it on a new branch would be even easier :smile:

risatronic:
> Haha, you're right that I did mean `devenv`, not `master`. I was meaning actually use `git rebase devenv` locally on your branch and then force pushing. But you're right, just redoing it on a new branch would be even easier 😄

Okay then that is what I will do since it'll take approximately 2 minutes haha, thanks! :)

PR #71, opened 2021-01-25
Implement Linting (CU-kcxpaw)

risatronic:
Lots of small changes/manual fixes, most straightforward but I've disabled a couple of things that I either didn't understand or which made tests fail, please check commit history

It looks like I can also centralise this with a separate prettier repo like the eslint one and just import both in the same way, so I will update that if needed

I haven't connected this to the ClickUp task because it is in the Buttons Kanban and I did this one first, so I will either connect it to that one, or make a second task to link to this one in the Sensor Kanban depending on what people think?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,47 +1,5 @@
 {
     "extends": [
-        "airbnb",
-        "prettier"
-    ],
-    "plugins": [
-        "prettier"
-    ],
-    "rules": {
-        "prettier/prettier": [
-            "error"
-        ],
-        "func-style": [
-            "error",
-            "declaration"
-        ],
-        "function-call-argument-newline": [
-            "error",
-            "consistent"
-        ],
-        "prefer-destructuring": [
-            "error",
-            {
-                "array": false,
-                "object": false
-            },
-            {
-                "enforceForRenamedProperties": false
-            }
-        ],
-        "no-restricted-syntax": [
-            "error",
-            "ForInStatement",
-            "LabeledStatement",
-            "WithStatement"
-        ],
-        "no-await-in-loop": [
-            0
-        ],
-        "camelcase": [
-            0
-        ],
-        "consistent-return": [
-            0
-        ]
-    }
+        "eslint-config-brave"
+    ]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,22 +1,32 @@
 {
-    "env": {
-        "node": true,
-        "es6": true
-    },
-    "extends": "eslint:recommended",
-    "globals": {
-        "Atomics": "readonly",
-        "SharedArrayBuffer": "readonly"
-    },
-    "parserOptions": {
-        "ecmaVersion": 2018
-    },
+    "extends": [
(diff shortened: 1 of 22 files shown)


trisapeace:
Task linked: [CU-kcxpaw Review & Update ESLint](https://app.clickup.com/t/kcxpaw)

risatronic:
I am not sure why having the file names in the `db` folder be capitalised to match the class names is apparently a problem for this despite updating imports when changing the file names, but I made them lower-case again and just left the class names within the files capitalised and it's now happy ¯\_(ツ)_/¯

trisapeace:
I kind of think that we have too many open pull requests right now for this to be particularly useful to code review yet since it'll likely be easiest to do it again from scratch once the others are merged rather than trying to resolve conflicts in all the changed files.

Could you point out just the things that you'd like review on? Like the things that you're unsure about that might result in changes to the ESLint or Prettier config for everyone?

risatronic:
> I kind of think that we have too many open pull requests right now for this to be particularly useful to code review yet since it'll likely be easiest to do it again from scratch once the others are merged rather than trying to resolve conflicts in all the changed files.
> 
> Could you point out just the things that you'd like review on? Like the things that you're unsure about that might result in changes to the ESLint or Prettier config for everyone?

sorry are you saying I will have to manually change everything again that I changed in this? I reeeally hope not because especially this repo in particular took a very long time 😬 I'll do it if I have to obviously but I really hope I'm misunderstanding 😅

trisapeace:
Yeah, that is what I mean. Unless we decide that all the other PRs will need to deal with the conflicts instead. But I really want to get my Brave Alert Lib one in sooner rather than later because it's already need to be redone too many times.

risatronic:
> Yeah, that is what I mean. Unless we decide that all the other PRs will need to deal with the conflicts instead. But I really want to get my Brave Alert Lib one in sooner rather than later because it's already need to be redone too many times.

ok if I'm possibly going to have to do that I feel like I may as well not update anything else until that's decided on because I don't want to add more things I'll have to do over :/ I kind of had assumed the point of me doing all the manual changes now was to integrate them, otherwise I would've waited

risatronic:
> > Yeah, that is what I mean. Unless we decide that all the other PRs will need to deal with the conflicts instead. But I really want to get my Brave Alert Lib one in sooner rather than later because it's already need to be redone too many times.
> 
> ok if I'm possibly going to have to do that I feel like I may as well not update anything else until that's decided on because I don't want to add more things I'll have to do over :/ I kind of had assumed the point of me doing all the manual changes now was to integrate them, otherwise I would've waited

I can hopefully preserve most of it I just don't want to do more things to it until that's figured out if that makes sense, sorry. Your thing has been in the works for a good while though so I get where you're coming from, I just think I maybe misunderstood what I was supposed to be doing

trisapeace:
Yes, the Buttons ones should be able to integrate. It's just this one has too much waiting in the wings that would need changing and hence retesting. I don't think it's a good time yet for the Sensor repo.

PR #76, opened 2021-02-13
Tell Helm to get the PG_DATABASE value from the secrets file

trisapeace:
@mariocimet This is what the problem was yesterday. It seems as though it's OK to build with a `.env` in my local, as long as all the environment variables are declared in the deployment.

Maybe we also need a section on updating environment variables since it's now a number of steps...
- Updating your local `.env`
- Letting people know that it's changed
- Updating the `*.env` files on ODetect Admin
- Updating the `sensor-*` secrets on ODetect Admin
- Updating `.travis.yml`
... is there anything else?

### DIFF
--- a/sensor-helm-chart/templates/sensor-deployment.yaml
+++ b/sensor-helm-chart/templates/sensor-deployment.yaml
@@ -60,6 +60,11 @@ spec:
               secretKeyRef:
                 name: {{.Values.secretName}}
                 key: PG_USER
+          - name: PG_DATABASE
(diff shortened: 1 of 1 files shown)


mariocimet:
doh! that would do it

mariocimet:
re: the section, you're referring to changing the schema right? I think it's worthwhile to add that yeah, and differentiate it from updating the values therein.

mariocimet (approved):

PR #78, opened 2021-02-23
Added a smoke testing script with setup and teardown endpoints. (CU-kcxndt)

mariocimet:
Check it out, this is running on the staging deployment. To run the smoke test, use the following command:

`node smokeTest 'https://staging.odetect.brave.coop/'  '<your phone>' '<valid twilio number for our account>'`

The desired behavior is the creation of a session which gets closed, and then a session which generates an alert. The procedure I have in mind is checking the phone that you gave as a parameter, as well as the front-end for the sessions.

I also like to look at the logs on the server with 

`kubectl logs -f deploy/staging-sensor-server`.

I will note that I get 'bad request sender is not twilio' for this 🤔.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,24 @@ https://docs.docker.com/engine/install/linux-postinstall/)
     ./build.sh
     ```
 
+## Running Smoke Tests on the Kubernetes Cluster
+
+Once a build is deployed to the cluster, you can test basic aspects of it's functionality by running the smoketest.js script. The scrip takes three parameters, a url for a deployment, a recipient phone number, and a Twilio number:
+
+    `npm run smoketest <server url> <your phone number> <valid Twilio number>`
+
+So, for example, if I wanted to run the smoke test with +1778228537 as the recipient phone number, and I knew +17786083684 was a valid twilio number for the deployment I'm testing, I could run the following command:
+
+    `npm run smoketest 'https://staging.odetect.brave.coop' '+17782285537' '+17786083684'`
+
+The script will take a few minutes to conclude. The expected behavior is for a location to be created, and for two sessions to be started - one which closes without generating an alert, and one which results in a 'Stillness' alert to the phone number provided. These two sessions should also be reflected in the frontend dashboard. If you do not answer the alert on your phone, you should notice a notification about a session being restarted. If you do, you should complete the chatbot flow.
+
+You can get further details about the behavior of the build by watching logs for the application with: 
+`kubectl logs -f deployment/dev-sensor-server`
+and by watching the behavior of redis with: 
+`kubectl exec deploy/redis-dev redis-cli monitor`
+
+
 # Local Development
 
 ## Dependencies
--- a/README.md
+++ b/README.md
@@ -63,7 +63,19 @@ If there are changes to the schema of the environment variables (like the additi
 
 ## 4. Deploy changes to the Kubernetes Cluster
 
-We use [helm](https://helm.sh/docs/) to manage our deployments. Helm allows us to deploy different instances of the application corresponding to dev, staging, and production environments as different 'releases'.
+We use [helm](https://helm.sh/docs/) to manage our deployments. Helm allows us to deploy different instances of the application corresponding to dev, staging, and production environments as different 'releases'. Before updating production, we run smoke tests to verify that there's nothing obviously wrong with a release candidate.
+
+### Smoke testing
+
+1. Run `ssh brave@odetect-admin.brave.coop`
+
+1. cd into the ~/BraveSensor-Server/sensor-helm-chart/ directory and run `git pull origin master` to get the latest version of the helm chart
+
+1. Run the command `helm upgrade staging --set secretName=sensor-dev --set image.tag=<your new container tag>` to deploy the latest version to staging
+
+1. Run the smoke test script as described [below](#running-smoke-tests-on-the-kubernetes-cluster)
+
+1. Verify that the deployment shows the expected behavior before proceeding
 
 ### Updating production
 
@@ -103,18 +115,18 @@ https://docs.docker.com/engine/install/linux-postinstall/)
 
 ## Running Smoke Tests on the Kubernetes Cluster
 
-Once a build is deployed to the cluster, you can test basic aspects of it's functionality by running the smoketest.js script. The scrip takes three parameters, a url for a deployment, a recipient phone number, and a Twilio number:
+Once a build is deployed to the cluster, you can test basic aspects of its functionality by running the smoketest.js script. The script takes three parameters, a url for a deployment, a recipient phone number, and a Twilio number.:
 
     `npm run smoketest <server url> <your phone number> <valid Twilio number>`
 
-So, for example, if I wanted to run the smoke test with +1778228537 as the recipient phone number, and I knew +17786083684 was a valid twilio number for the deployment I'm testing, I could run the following command:
+The twilio number must be a valid number for the account, and must have its SMS webhook pointing at the server you're running the smoketest on. So, for example, if I wanted to run the smoke test with +17781234567 as the recipient phone number, and I knew +17786083684 was a valid twilio number for the deployment I'm testing, I could run the following command:
 
-    `npm run smoketest 'https://staging.odetect.brave.coop' '+17782285537' '+17786083684'`
+    `npm run smoketest 'https://staging.odetect.brave.coop' '+17781234567' '+17786083684'`
 
 The script will take a few minutes to conclude. The expected behavior is for a location to be created, and for two sessions to be started - one which closes without generating an alert, and one which results in a 'Stillness' alert to the phone number provided. These two sessions should also be reflected in the frontend dashboard. If you do not answer the alert on your phone, you should notice a notification about a session being restarted. If you do, you should complete the chatbot flow.
 
 You can get further details about the behavior of the build by watching logs for the application with: 
-`kubectl logs -f deployment/dev-sensor-server`
+`kubectl logs -f deployment/staging-sensor-server`
 and by watching the behavior of redis with: 
 `kubectl exec deploy/redis-dev redis-cli monitor`
 
--- a/README.md
+++ b/README.md
@@ -79,12 +79,7 @@ We use [helm](https://helm.sh/docs/) to manage our deployments. Helm allows us t
 
 ### Updating production
 
-1. Run `ssh brave@odetect-admin.brave.coop`
-
-1. cd into the ~/BraveSensor-Server/sensor-helm-chart/ directory and run `git pull origin master` to get the latest version of the helm chart
-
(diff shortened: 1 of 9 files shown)


smokeTest.js
```diff
@@ -0,0 +1,112 @@
+process.argv.forEach((val, index) => {
```
trisapeace, line 1: What is this doing? The `npm run smoketest` script doesn't seem to be passing any arguments to it.

smokeTest.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function getRandomArbitrary(min, max) {
+    return Math.random() * (max - min) + min
+}
+
+
+let destinationURL = process.argv[2];
+let recipientNumber = process.argv[3];
+let twilioNumber = process.argv[4];
```
trisapeace, line 26: Can these be `const`s?

smokeTest.js
```diff
@@ -0,0 +1,112 @@
...
+
+
+  
+const axios = require('axios').default;
```
trisapeace, line 30: I think I removed the dependency on axios in https://app.clickup.com/t/j6yuzk. You'll probably need to put it back.

smokeTest.js
```diff
@@ -0,0 +1,112 @@
...
+    try{
+        await axios.post('/smokeTest/setup', {
+            recipientNumber: recipientNumber,
+            twilioNumber: twilioNumber,
+            locationid: testLocation1Id
+        })
+    }catch(e){helpers.log(e)}
+}
+
+async function silence(locationid){
```
trisapeace, line 71: Is this exactly the same as in serverTest.js? If so, maybe a testHelper.js file would be useful for these sorts of things?

trisapeace (commented):
Thanks for doing this. On the right track!

smokeTest.js
mariocimet: Just logging the arguments that have been passed in, I think it can be removed

smokeTest.js
mariocimet: they are yeah, I like that idea and will refactor

mariocimet: they aren't the same right now, but they are very similar. The differences are the use of chai in tests vs axios in the smoketests. These are parameterized to access a provided URL, the serverTest.js ones have a hardcoded reference to the `server` object.

mariocimet: looking at it now it seems chai can be used just fine for SmokeTests, so I think I'll still do a helper file with that

mariocimet:
@trisapeace @risatronic this is ready for another review/pass :)

smokeTest.js
mariocimet: so I ended up not doing a testHelpers file. I actually tried but kep trunning into issues using chai for url testing with a remote server. It seems like it really only works when you have an actual app/server object (like from express) to pass to it. Which is the case in the unit tests, but not in the smoke tests

mariocimet: could still see an argument for a testHelpers file, but it would only have the getRandomInt and getRandomArbitrary functions :/ :shrug:

trisapeace: Would probably be a good idea, and would also make it clear that such a place does exist for when other temptations to copy-and-paste crop up ;-)
But it's not a critical change, for me.

README.md
```diff
@@ -101,6 +101,24 @@ https://docs.docker.com/engine/install/linux-postinstall/)
     ./build.sh
     ```
 
+## Running Smoke Tests on the Kubernetes Cluster
```
trisapeace, line 104: Thanks for the instructions. I think this should be a level up in headers since it's not really a subsection of `Build and push a new container`.

I'm wondering if we should add a final step to `How to release a new version of BraveSensor-Server` to run the smoke tests. Do you feel confident that this could run safely in production?

mariocimet: I am pretty confident yeah, but was envisioning this as something that happens in staging, so to speak, before you upgrade the production development. So it would actually be more a pre-step than a final one?

trisapeace: That sounds good to me too. Is the staging environment ready to be used?

README.md
```diff
@@ -101,6 +101,24 @@ https://docs.docker.com/engine/install/linux-postinstall/)
...
+
+Once a build is deployed to the cluster, you can test basic aspects of it's functionality by running the smoketest.js script. The scrip takes three parameters, a url for a deployment, a recipient phone number, and a Twilio number:
```
trisapeace, line 106: Typo: `scrip` should be `script`.

trisapeace, line 106: Typo: `it's` should be `its`.

README.md
mariocimet: it's ready, it could use some enhancements like its own redis instance and its own database, but I've been using it to test this already without issues

smokeTest.js
```diff
@@ -0,0 +1,112 @@
+const XETHRU_STATE = require('./SessionStateXethruEnum.js')
+const MOV_THRESHOLD = 17
+const IM21_DOOR_STATUS = require('./IM21DoorStatusEnum')
+const testLocation1Id = 'TestLocation1'
```
trisapeace, line 4: I think by having this hardcoded we can only ever run one smoketest at a time, right? I think that's OK, but it might be worth noting in the README.

smokeTest.js
```diff
@@ -0,0 +1,112 @@
...
+    for(let i = 0; i<15; i++){
+        await movement(testLocation1Id, getRandomInt(0,MOV_THRESHOLD), getRandomInt(0,MOV_THRESHOLD))
+    }
+    for(let i = 0; i<15; i++){
+        await movement(testLocation1Id, getRandomInt(MOV_THRESHOLD + 1,100), getRandomInt(MOV_THRESHOLD + 1,100))
+    }
+    for(let i = 0; i<85; i++){
+        await silence(testLocation1Id)
+    }
+    await teardown()
```
trisapeace, line 49: It would probably be a good idea to put the `teardown` in a `finally` block since we'll want to clear it out of the DB even if something else goes wrong.

db/db.js
```diff
@@ -464,6 +464,11 @@ async function clearLocations() {
     await pool.query("DELETE FROM locations")
 }
 
+async function clearTestLocation(locationid) {
+    await pool.query("DELETE FROM locations where locationid=$1", [locationid])
```
trisapeace, line 468: I think this should also clear out the sessions that were created by this location. Is there a reason to leave it behind?

smokeTest.js
trisapeace: Hmmm, did you not need to do this? I don't see it in `package.json`.

README.md
trisapeace: Then that makes sense. A pre-final step to deploy to staging and run the smoke tests there seems like a great idea.

trisapeace:
I tried this out and staging and it worked! So cool! So much better than needing to have a test sensor set up or waiting for people to do what we want in production and grabbing that data. LOOOOOVE IT! :heart:

smokeTest.js
mariocimet: yes! I like this suggestion.

README.md
```diff
@@ -101,6 +101,24 @@ https://docs.docker.com/engine/install/linux-postinstall/)
...
+
+    `npm run smoketest <server url> <your phone number> <valid Twilio number>`
+
+So, for example, if I wanted to run the smoke test with +1778228537 as the recipient phone number, and I knew +17786083684 was a valid twilio number for the deployment I'm testing, I could run the following command:
+
+    `npm run smoketest 'https://staging.odetect.brave.coop' '+17782285537' '+17786083684'`
+
+The script will take a few minutes to conclude. The expected behavior is for a location to be created, and for two sessions to be started - one which closes without generating an alert, and one which results in a 'Stillness' alert to the phone number provided. These two sessions should also be reflected in the frontend dashboard. If you do not answer the alert on your phone, you should notice a notification about a session being restarted. If you do, you should complete the chatbot flow.
```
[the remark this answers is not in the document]
mariocimet: true. What do you think about instead removing this limitation? It could be as simple as making it so that we create a location with "TestLocation" + getRandomInt

although I guess you could get collisions in that case

smokeTest.js
mariocimet: True, do you think it's worth generating a unique id here to avoid this and allow arbitrary smoke tests?

db/db.js
mariocimet: It should, you're right

smokeTest.js
trisapeace: I think it's OK just having the one. Although maybe change the name to something more obviously smoketest-related so that if it ever does end up in the DB then it'll be clear that it can just be deleted.

trisapeace:
Task linked: [CU-kcxndt create smoke test documentation and testing script](https://app.clickup.com/t/kcxndt)

README.md
risatronic: I also think this makes sense!

index.js
```diff
@@ -462,6 +462,28 @@ app.post(
         }
     })
 
+app.post('/smokeTest/setup', async(request, response) => {
+    const {locationid,recipientNumber,twilioNumber} = request.body
+    try{
+        await db.createLocation(locationid, '0', recipientNumber, 17, 15, 150, 30000, 120000, 3000, recipientNumber, twilioNumber, recipientNumber, 9000, locationid, 'door_coreID', 'radar_coreID')
```
risatronic, line 468: Out of curiosity are those thresholds 17 & 15 for a specific reason or just to have something to put in there? (I had to check the db function to see what they were is why I'm wondering)

smokeTest.js
risatronic: I can see both angles of this too 😅  More helpery type functions will probably crop up over time so I feel like it will eventually exist regardless of when that happens

index.js
mariocimet: the movement threshold could be anything, the time related thresholds I set for the purposes of convenience and enabling a short test period

risatronic: That makes sense, thanks! :)

risatronic: I just wasn't sure if those numbers were important haha

smokeTest.js
```diff
@@ -0,0 +1,129 @@
...
+const destinationURL = process.argv[1]
+const recipientNumber = process.argv[2]
+const twilioNumber = process.argv[3]
+
+function sleep(millis) {
+  return new Promise(resolve => {
+    setTimeout(resolve, millis)
+  })
+}
+axios.defaults.headers.post['Content-Type'] = 'application/x-www-form-urlencoded'
```
risatronic, line 31: I feel like there should perhaps be another line break above these axios guys but that won't affect it working or not 😅

risatronic (approved):
The only things I have to say were about the one line break and also do we want the empty `testHelpers.js` file?

trisapeace (approved):

PR #79, opened 2021-02-27
Implement Linting (CU-kcxpaw)

risatronic:
- Add Brave-wide linting to Sensor repo & update code to conform
- Add repository field to `package.json`
- Update build file name in readme

I have a couple of clarifying questions/points of interest for people to go over:

1. In `serverTest.js` I disabled a rule for the first line with the sleep function, not sure how people feel about that vs me refactoring it
2. In `BraveAlerterConfigurator.js` I disabled a couple of rules and had done a similar thing in Buttons, was looking for feedback specifically from @trisapeace on this one (the "class methods use this" rule and "no await in return" which doesn't like `return await functionCall()`)
3. I have disabled the eslint "eqeqeq" rule in a few places but it looks like (from my assessment) the only place it would cause problems to update comparisons to be using `===` instead of `==` is in OD-flag related areas because its values seem to be strings that are just single digits where the other things being compared are generally longer non-numeric strings or `undefined` - there is also one interesting case where values are `1 * [boolean expression]` and then being used in what appears to be addition to be compared to numbers higher than 0/1 so I assume that's forcing boolean true/false into 1/0 and can be updated to be `===` with no issues but wanted to check
4. I disabled the complaint about `console.error` for the same reason discussed when it came up in Buttons

I think that's it? I also added `// prettier-ignore` in a few places similar to where I added it in buttons to avoid 17 different things all being on their own lines in `db.js` but that can easily be changed :)


trisapeace:
Task linked: [CU-kcxpaw Review & Update ESLint](https://app.clickup.com/t/kcxpaw)

index.js
```diff
@@ -33,176 +33,383 @@ const app = express()
...
+        // Checks if no session exists for this location yet.
+        if (latestSession.end_time == null) {
+          // Checks if session is open.
+          await db.updateSessionState(latestSession.sessionid, currentState)
+        }
+      }
+    }
+
+    // Checks if current state belongs to the session triggerStates
+    else if (TRIGGERSTATES.includes(currentState)) {
```
trisapeace, line 231: I'm not a huge fan of how this put the `else` a couple lines after the closing brace on the `if`. Did it not like it when the line started with `}`? Or was this something that was there before?

index.js
```diff
@@ -33,176 +33,383 @@ const app = express()
...
+        }
+      } else {
+        const currentSession = await db.createSession(location.phonenumber, currentLocationId, currentState, client)
+        start_times[currentLocationId] = currentSession.start_time
+      }
+      await db.commitTransaction(client)
+    }
+
+    // Checks if current state belongs to the session closingStates
+    else if (CLOSINGSTATES.includes(currentState)) {
```
trisapeace, line 253: Same here with the newlines that I'm not a fan of.

index.js
```diff
@@ -33,176 +33,383 @@ const app = express()
...
+            fallbackToPhoneNumber: location.fallback_phonenumber,
+            fallbackFromPhoneNumber: location.twilio_number,
+          }
+          braveAlerter.startAlertSession(alertInfo)
+        }
+      }
+    } else {
+      helpers.log('Current State does not belong to any of the States groups')
     }
-}));
```
trisapeace, line 89: In this ridiculously long file, I'm not sure what these brackets were there for. Is it OK that they've been taken out now?

test/serverTest.js
```diff
@@ -1,14 +1,18 @@
-const sleep = (millis) => new Promise(resolve => setTimeout(resolve, millis))
+// eslint-disable-next-line func-style
+const sleep = millis => new Promise(resolve => setTimeout(resolve, millis))
```
trisapeace, line 2: I'm OK with this. But was it just weird looking to have it a normal `function` declaration? It also doesn't need to be at the very top of the file if that's the part that would make it look funny.

test/serverTest.js
```diff
@@ -19,462 +23,557 @@ const testLocation1Id = 'TestLocation1'
...
 
-async function im21Door(door_coreID, signal){
-    try{
-        await chai.request(server).post('/api/door').send(
-            {
-                coreid: door_coreID,
-                data: `{ "deviceid": "FA:E6:51", "data": "${signal}", "control": "86"}`
-            })
-    }catch(e){helpers.log(e)}
+async function im21Door(coreid, signal) {
```
trisapeace, line 85: Did `door_codeid` need to change because of the underscore? How about leaving the word `door` in there, so something like `doorCoreId`?

trisapeace (commented):
Looking good.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ the code was deployed.
 ### Changed
 - Use Brave Alert Lib for Twilio communication with the Responder (CU-bar1zy)
 
+### Updated
```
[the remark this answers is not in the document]
risatronic: Can "update" that, haha

db/DoorData.js
```diff
@@ -0,0 +1,8 @@
+class DoorData {
```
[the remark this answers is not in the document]
risatronic: yeah I think I should have just force pushed after changing them but this also worked in a roundabout way 😅

db/redis.js
```diff
@@ -1,123 +1,117 @@
...
+async function addDoorTestSensorData(locationid, signal) {
+  await client.xadd(`door:${locationid}`, '*', 'signal', signal)
 }
 
-async function addIM21DoorSensorData(locationid, doorSignal){
-    if(doorSignal==SESSIONSTATE_DOOR.CLOSED||doorSignal==SESSIONSTATE_DOOR.OPEN){
-        await client.xadd("door:"+locationid,  "*","signal", doorSignal)
-    }
+async function addIM21DoorSensorData(locationid, doorSignal) {
+  if (doorSignal === SESSIONSTATE_DOOR.CLOSED || doorSignal === SESSIONSTATE_DOOR.OPEN) {
```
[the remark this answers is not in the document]
risatronic: can do!

index.js
risatronic: I am 99% sure it's because of the comment line, it doesn't like them wedged in-between things. I could make them in-line comments maybe?

index.js
risatronic: After investigating, it looks like this is now split over a couple of lines to add trailing commas

```
app.use(session({
    key: 'user_sid',
    secret: helpers.getEnvVar('SECRET'),
    resave: false,
    saveUninitialized: false,
    cookie: {
        expires: 10*60*1000
    }
}));
```
has become:
```
app.use(
  session({
    key: 'user_sid',
    secret: helpers.getEnvVar('SECRET'),
    resave: false,
    saveUninitialized: false,
    cookie: {
      expires: 10 * 60 * 1000,
    },
  }),
)
```

risatronic: I'm glad you asked though because I had a momentary panic since this was initially aligned with a line corresponding to `handleSensorRequest` in the review preview and I did move that beast of a function 😆

test/serverTest.js
risatronic: I don't think it would look particularly weird, no, I just figured it might have been done like this for a reason since there are other things that are actually written out functions going on lower down 😅 I just changed it to this and put it at the top of function declarations and all the tests still pass, so I can update it if that looks nicer!

```
function sleep(millis) {
  return new Promise(resolve => {
    setTimeout(resolve, millis)
  })
}
```

db/redis.js
```diff
@@ -1,123 +1,117 @@
...
-const addXeThruSensorData = (locationid, state, rpm, distance, mov_f, mov_s) => {
-    client.xadd("xethru:" + locationid,  "*", 
-        "state", state,
-        "distance", distance,
-        "rpm", rpm, 
-        "mov_f", mov_f,
-        "mov_s", mov_s, 
-    );
+function addXeThruSensorData(locationid, state, rpm, distance, mov_f, mov_s) {
+  client.xadd(`xethru:${locationid}`, '*', 'state', state, 'distance', distance, 'rpm', rpm, 'mov_f', mov_f, 'mov_s', mov_s)
```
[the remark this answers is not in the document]
risatronic: This actually apparently comes from eslint's `function-call-argument-newline` as well as prettier so I think I would have to disable that for the entire file (which I personally think is ok since it's happy with everything else as is), each individual line of that block in-line, orrrr probably the best option which is still not very pretty is below 😬 

```
// prettier-ignore
/* eslint-disable function-call-argument-newline */
function addXeThruSensorData(locationid, state, rpm, distance, mov_f, mov_s) {
  client.xadd(
    `xethru:${locationid}`, '*', 
    'state', state, 
    'distance', distance, 
    'rpm', rpm, 
    'mov_f', mov_f, 
    'mov_s', mov_s)
}
/* eslint-enable function-call-argument-newline */
```

I tried to see if that disable/enable can be after `//` instead as it looks slightly nicer but alas, no

test/serverTest.js
risatronic: This is just because it's "already declared in the outer scope" so that update is fine by me!

BraveAlerterConfigurator.js
```diff
@@ -1,110 +1,109 @@
+/* eslint-disable no-return-await */
```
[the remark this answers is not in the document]
risatronic: Nope that fixes it! :)

index.js
risatronic: It doesn't like them on the same line so the options left are to add a `// prettier-ignore` but that goes above the line so I think it'd just cause the same problem since that's also a comment, or to put the explanation inside the block like this (it actually already is that way above, which I included):

```
      if (typeof latestSession !== 'undefined') {
        // Checks if no session exists for this location yet.
        if (latestSession.end_time == null) {
          // Checks if session is open.
          await db.updateSessionState(latestSession.sessionid, currentState)
        }
      }
    } else if (TRIGGERSTATES.includes(currentState)) {
      // Checks if current state belongs to the session triggerStates
      const client = await db.beginTransaction()
      const latestSession = await db.getMostRecentSession(currentLocationId, client)
```

db/redis.js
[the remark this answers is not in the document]
trisapeace: Sure, I think that looks OK. It would be nice to have a comment explaining why all the ignores.

index.js
trisapeace: It's actually not a very useful comment. Let's get rid of it. Same with `Checks if current state belongs to voidStates` and `Checks if current state belongs to the session closingStates`

index.js
trisapeace: Thanks for finding that!

test/serverTest.js
trisapeace: Top of the function declarations seems good to me. Thanks!
It was probably just put there like that because it was copied as a one-liner from a Stackoverflow answer :smile:

test/serverTest.js
trisapeace: Oh interesting. OK, I'm not too fussed. Either way is good for me. Thanks!

test/serverTest.js
```diff
@@ -1,14 +1,18 @@
-const sleep = (millis) => new Promise(resolve => setTimeout(resolve, millis))
+// eslint-disable-next-line func-style
+// const sleep = millis => new Promise(resolve => setTimeout(resolve, millis))
```
trisapeace, line 2: This commented out line can be removed now.

trisapeace (approved):
Looks good to me. Just those last couple things that all have to do with comments in some way, shape, or form. Thanks!

index.js
risatronic: Works for me! 😅

risatronic:
Those have all been updated :)

trisapeace:
Fantastic. Thanks!

mariocimet (approved):

PR #80, opened 2021-03-03
Saving IM21 request control field for debugging door sensor issues (CU-n70k94)

mariocimet:
Why?

We are having trouble with false positives at Covenant_2 aka 'Youth #111' and suspect that door sensor problems are to blame. Tracking the "control bytes" can help us diagnose it. The control number should increase by one with each door event, so tracking these would reveal if we're missing any requests server side (as opposed to the sensor or argon itself missing events).

See our conversation at https://bravely.slack.com/archives/C0153GVBNGG/p1614800310001600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ the code was deployed.
 ### Added
 - Script, endpoints, and instructions for smoke tests (CU-kcxndt)
 - Update to use Brave-wide linting standard (CU-kcxpaw)
+- Saving control field in door sensor requests (CU-n70k94)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


risatronic (approved):
Approved per discussion during debugging with Heidi

trisapeace (approved):
Fine by me. Do we have any tests for this that need to be extended to include `control`?

trisapeace:
Fair enough.

Can you please add it to the CHANGELOG though and also include the ClickUp number in the name of the PR so that it is linked?
Thanks!

trisapeace:
Task linked: [CU-n70k94 Fix Covenant_2, Youth 111 False Positives](https://app.clickup.com/t/n70k94)

trisapeace (approved):

PR #83, opened 2021-03-07
checkHeartbeat running every second independently of incoming sensor data (CU-p34wcw)

mariocimet:
This will result in us getting notifications when they go down, not just when they come back on.

### DIFF
--- a/index.js
+++ b/index.js
@@ -148,23 +148,26 @@ async function sendBatteryAlert(locationid, signal) {
 }
 
 // Heartbeat Helper Functions
-async function checkHeartbeat(locationid) {
-  const location = await db.getLocationData(locationid)
-  // Query raw sensor data to transmit to the FrontEnd
-  const XeThruData = await redis.getLatestXeThruSensorData(location.locationid)
-  // Check the XeThru Heartbeat
-  const currentTime = moment()
-  const latestXethru = moment(XeThruData.timestamp, 'x')
-  const XeThruDelayMillis = currentTime.diff(latestXethru)
+async function checkHeartbeat() {
+  let locations = await db.getLocations()
 
-  if (XeThruDelayMillis > XETHRU_THRESHOLD_MILLIS && !location.xethru_sent_alerts) {
-    helpers.log(`XeThru Heartbeat threshold exceeded; sending alerts for ${location.locationid}`)
-    await db.updateSentAlerts(location.locationid, true)
-    sendAlerts(location.locationid)
-  } else if (XeThruDelayMillis < XETHRU_THRESHOLD_MILLIS && location.xethru_sent_alerts) {
-    helpers.log(`XeThru at ${location.locationid} reconnected`)
-    await db.updateSentAlerts(location.locationid, false)
-    sendReconnectionMessage(location.locationid)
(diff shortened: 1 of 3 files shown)


mariocimet:
tested this with a location on dev, it's up there if you want to disconnect and reconnect your XeThrus to check.

index.js
```diff
@@ -475,9 +475,6 @@ async function handleSensorRequest(currentLocationId){
     let location = await db.getLocationData(currentLocationId);
 
     // Check the XeThru Heartbeat
```
trisapeace, line 477: Should this comment be removed as well?

test/serverTest.js
```diff
@@ -159,6 +159,7 @@ describe('Brave Sensor server', () => {
             let sessions = await db.getAllSessionsFromLocation(testLocation1Id)
             expect(sessions.length).to.equal(1)
             expect(sessions[0].od_flag).to.equal(1)
+            await sleep(1000)
```
trisapeace, line 162: What are the new `sleep` s for?

trisapeace (commented):
I tried this with mine, but didn't get any notifications.

test/serverTest.js
mariocimet: my tests hang locally exactly in between these two, and if solving w a sleep i figured it's preferable to be one-offs rather than in the before/after eachs

index.js
mariocimet: yuppers

test/serverTest.js
trisapeace: OK, that's too bad, but if it works, then it works :+1:

index.js
```diff
@@ -653,6 +653,9 @@ async function sendBatteryAlert(locationid, signal) {
 
 const server = app.listen(8080);
 helpers.log('brave server listening on port 8080')
+if(!helpers.isTestEnvironment()){
+    setInterval(checkHeartbeat, 1000)
```
trisapeace, line 657: Is this once a second? That seems extreme.

mariocimet: yeah. The actual alert threshold is 60 seconds right now, we could do the same for the interval that we check at? or 30 seconds, any suggestion?

this is also the same interval as for Buttons

trisapeace: Ah! That makes sense about the threshold being different than how often we check.

I guess we don't want the maximum time it could be down to be longer than our stillness threshold (which is 60 seconds, right?). So that's probably why we did it this way. What if we did this check every 15 seconds and changed the alert threshold to 45 seconds?

Good point about Buttons. Maybe it's overkill there too, but it does mean that we get more accurate numbers for `heartbeatDashboard` and we have fewer things to check on than we're hopefully going to have for Sensors.

It's up to you :smile:

trisapeace (approved):
I suggest you wait until after the linting has been merged before merging this.

trisapeace:
Can you please update the CHANGELOG and put the ClickUp number in the PR name for this one as well?
Thanks!

trisapeace:
Task linked: [CU-p34wcw alter heartbeat so that it actually notifies when something goes down, not when it comes back on](https://app.clickup.com/t/p34wcw)

PR #85, opened 2021-03-09
Update dependencies to their latest minor version (CU-kry4g7)

trisapeace:
- Did not update any major versions because of the potential risk of breaking something
- This updates the version of minimist, which was what triggered the need for a security update
- Also added date of the last deployment because we forgot


trisapeace:
Task linked: [CU-kry4g7 Update dependencies to avoid known minimist security issue](https://app.clickup.com/t/kry4g7)

package.json
```diff
@@ -16,29 +16,29 @@
   ],
   "dependencies": {
     "@smartthings/smartapp": "^1.0.3",
-    "body-parser": "*",
+    "body-parser": "^1.19.0",
```
risatronic, line 19: I actually just noticed in my local repo that body parser was crossed out cause it's apparently deprecated, is that recent or did I just not see it before?

risatronic (approved):
Looks fine to me! :)

package.json
risatronic: yeah we used it in school recently as well as here, and it's never been crossed out before so I thought it was weird too!

mariocimet (approved):

PR #86, opened 2021-03-11
added max length to function that populates xeThru redis stream CU-p34wbb

mariocimet:
the goal with this is bounded and predictable memory usage, compared to the unpredictable behavior that results in exceeding our memory limits and for example, the door stream for a location being deleted

### DIFF
--- a/db/redis.js
+++ b/db/redis.js
@@ -77,7 +77,7 @@ async function addVitals(locationid, signalStrength, cloudDisconnects) {
 /* eslint-disable function-call-argument-newline */
 function addXeThruSensorData(locationid, state, rpm, distance, mov_f, mov_s) {
   client.xadd(
-    `xethru:${locationid}`, 'MAXLEN', '604800', '*',
+    `xethru:${locationid}`, 'MAXLEN', '~', '604800', '*',
     'state', state, 
     'distance', distance, 
     'rpm', rpm, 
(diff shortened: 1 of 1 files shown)


trisapeace:
Task linked: [CU-p34wbb Configure redis to never evict door data](https://app.clickup.com/t/p34wbb)

db/redis.js
```diff
@@ -64,7 +64,7 @@ async function addDoorTestSensorData(locationid, signal) {
 async function addIM21DoorSensorData(locationid, doorSignal) {
   // eslint-disable-next-line eqeqeq
   if (doorSignal == SESSIONSTATE_DOOR.CLOSED || doorSignal == SESSIONSTATE_DOOR.OPEN) {
-    await client.xadd(`door:${locationid}`, '*', 'signal', doorSignal)
+    await client.xadd(`door:${locationid}`, 'MAXLEN', '~', '10000', '*', 'signal', doorSignal)
```
trisapeace, line 67: Are these maximums for the single location or for the whole redis instance?

mariocimet: for a single location. The way it works is that it limits the length of a key, in this case a single key would be door:TestLocation

trisapeace: Gotcha. Thanks for clarifying.

db/redis.js
```diff
@@ -64,7 +64,7 @@ async function addDoorTestSensorData(locationid, signal) {
...
   }
 }
```
trisapeace, line 70: I see that you didn't add a MAXLEN to `addVitals`, `addDoorSensorData`, or `addDoorTestSensorData`. I could see leaving the door ones out because they are being removed in the INS PR soon, but why not do `addVitals`?

mariocimet: oversight!

trisapeace (approved):

PR #87, opened 2021-03-13
Use Travis to push Docker containers to DO (CU-mrxren)

trisapeace:
- This will ensure that our images are built in a clean and
  consistent environemnt. No chance to anything weird getting
  bundled in from our local dev environments

- Also used the prettier format for the README and CHANGELOG. Because of this, you might want to turn on "Hide whitespace changes" in order to see what really changed in thie PR.

- Because I didn't want the push to Digital Ocean to happen unless
  the tests and linting pass, I refactored Travis to use staged jobs.
  To make this work, I couldn't use the matrix anymore (which we didn't
  really need anyways) and I needed to move all the env variables to be
  global variables so that they will all be used for every run (in this
  case, we only have one run anyways).

- Updated production deployment instructions to use the container that
  was deployed by Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,56 +1,67 @@
 dist: xenial
 
-matrix:
+language: node_js
+node_js: 'node' # node version specified in .nvmrc
+
+# install dependencies
+install:
+  - npm ci
+  - npm install pg@latest
+
+# install Redis
+services:
+  - redis-server
+
+addons:
+  # install PostgreSQL
+  postgresql: '12'
+  apt:
+    packages:
+      - postgresql-12
+      - postgresql-client-12
+  # install Digital Ocean API CLI tool
+  snaps:
+    - doctl
+
+# setup environment variables because Travis doesn't have a .env file
+env:
+  global:
+    - WEB_USERNAME_TEST=TravisDashboardUsername
+    - PASSWORD_TEST=TravisDashboardPassword
+    - SECRET_TEST=TravisCookieSecret
+    - PG_USER_TEST=brave
+    - PG_DATABASE_TEST=brave
+    - PG_PASSWORD_TEST=travispassword
+    - PG_PORT_TEST=5433
+    - PG_HOST_TEST=localhost
+    - PG_USER=brave
+    - PG_DATABASE=brave
+    - PG_PASSWORD=travispassword
+    - PG_PORT=5433
+    - PG_HOST=localhost
+    - REDIS_CLUSTER_IP_TEST=localhost
+    # TWILIO_SID_TEST (from https://www.twilio.com/console/voice/project/test-credentials)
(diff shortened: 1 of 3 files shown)


trisapeace:
Task linked: [CU-mrxren replace 'build.sh' scripts with pushing container images directly from travis](https://app.clickup.com/t/mrxren)

trisapeace:
The version up on `dev.odetect.brave.coop` right now was generated by this branch.

mariocimet:
You mentioned you wanted to possibly reduce the access to Digital Ocean that this has? I'm ok with this as is but curious what you had in mind

trisapeace:
There are apparently three different ways to authenticate with DO (https://www.digitalocean.com/docs/container-registry/how-to/set-up-ci-cd/), and I'm wondering if the first one they describe, the Docker Configuration file is more restrictive since it comes directly from the registry page. But I don't know that for sure yet.

trisapeace:
Nope, I give up and have spent enough time on it. I can't get the registry docker config file style of authentication to work, let alone see whether it gives reduced permissions. 

So this has the Digital Ocean access token as an encrypted environment variable in Travis. This means that any Travis run from a branch in our repo will be able to use the  `doctl` commands to do whatever the Digital Ocean Service User is allowed to do.

Right now, the only people who have permission to create branches on our repo are us (Mario, Sampath, Marisa, and me) because we have Write permission (ref: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request#:~:text=Anyone%20with%20read%20permissions%20to,can%20fork%20the%20repository%20first.). 

Anyone else is allowed to create pull requests on our repo, but must create a fork to do so, and forks do not have access to our encrypted Travis variables (ref: https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions).

So the risk is just that one of us does something bad by accident. Which I think is pretty unlikely because the `doctl` commands seems pretty long and hard to get right. How do you feel about this, @mariocimet ?

risatronic:
> Nope, I give up and have spent enough time on it. I can't get the registry docker config file style of authentication to work, let alone see whether it gives reduced permissions.
> 
> So this has the Digital Ocean access token as an encrypted environment variable in Travis. This means that any Travis run from a branch in our repo will be able to use the `doctl` commands to do whatever the Digital Ocean Service User is allowed to do.
> 
> Right now, the only people who have permission to create branches on our repo are us (Mario, Sampath, Marisa, and me) because we have Write permission (ref: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request#:~:text=Anyone%20with%20read%20permissions%20to,can%20fork%20the%20repository%20first.).
> 
> Anyone else is allowed to create pull requests on our repo, but must create a fork to do so, and forks do not have access to our encrypted Travis variables (ref: https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions).
> 
> So the risk is just that one of us does something bad by accident. Which I think is pretty unlikely because the `doctl` commands seems pretty long and hard to get right. How do you feel about this, @mariocimet ?

To clarify that I understand this, this is the same idea as the other things that say `secure` in the current `.travis.yml` on `devenv` but it is fine for them all to be there because no one except us can do "internal pull requests" as referenced in the Travis documentation you linked so it won't let them actually use those?

trisapeace:
@risatronic , yes that's right. However, the other things that are currently `secure` are just Twilio credentials. Digital Ocean is much more important than Twilio because it's where our servers and DB are.

mariocimet:
It feels reasonably safe to me. It's hard enough to do doctl things on purpose!

mariocimet (approved):

risatronic (approved):
Looks good to me! :)

PR #88, opened 2021-03-15
Add INS Radar Endpoint (CU-m0wdfn)

mariocimet:
Looking for feedback on the implementation which is basically just a fork in the road at handleSensorRequest


trisapeace:
Task linked: [CU-m0wdfn create API endpoint for INS radar data](https://app.clickup.com/t/m0wdfn)

db/db.js
```diff
@@ -418,10 +492,10 @@ async function getLocations() {
...
+async function updateLocationData(deviceid, phonenumber, detection_min, detection_max, sensitivity, noisemap, led, rpm_threshold, still_threshold, duration_threshold, mov_threshold, door_particlecoreid, radar_particlecoreid, location, radar_type) {
   try {
     const results = await pool.query(
-      'UPDATE locations SET deviceid = $1, phonenumber = $2, detectionzone_min = $3, detectionzone_max = $4, sensitivity = $5, noisemap = $6, led = $7, rpm_threshold = $8, still_threshold = $9, duration_threshold = $10, mov_threshold = $11, door_particlecoreid = $12, radar_particlecoreid = $13 WHERE locationid = $14 returning *',
+      'UPDATE locations SET deviceid = $1, phonenumber = $2, detectionzone_min = $3, detectionzone_max = $4, sensitivity = $5, noisemap = $6, led = $7, rpm_threshold = $8, still_threshold = $9, duration_threshold = $10, mov_threshold = $11, door_particlecoreid = $12, radar_particlecoreid = $13, radar_type = $14 WHERE locationid = $14 returning *',
```
sampathsatti, line 498: locationid = $15, not $14 ...

sampathsatti (approved):
Looks great! There’s a minor edit that we might need to make in db.js with the update location function but we might not need to do that at all since we don’t call that function ever...

mariocimet:
> Looks great! There’s a minor edit that we might need to make in db.js with the update location function but we might not need to do that at all since we don’t call that function ever...

good catch. I think it might be better to delete it altogether if its not in use?

risatronic:
> > Looks great! There’s a minor edit that we might need to make in db.js with the update location function but we might not need to do that at all since we don’t call that function ever...
> 
> good catch. I think it might be better to delete it altogether if its not in use?

@mariocimet @sampathsatti I am adding the ability to add/edit locations from the dashboard aren't I? I feel like this would be useful if so even if it isn't used right now!

risatronic (approved):
Looks good to me! :)

InnosentStateMachine.js
```diff
@@ -0,0 +1,128 @@
...
+          // eslint-disable-next-line eqeqeq
+          if (door.signal == DOOR_STATE.CLOSED) {
+            if (doorDelay < DOOR_THRESHOLD_MILLIS) {
+              state = STATE.DOOR_OPENED_START
+            }
+            if (doorDelay >= DOOR_THRESHOLD_MILLIS && inPhase_avg > mov_threshold) {
+              state = STATE.DOOR_CLOSED_START
+            } else {
+              state = STATE.NO_PRESENCE_NO_SESSION
+            }
```
trisapeace, line 74: This seems strange to me. If we happens to have doorDelay < DOOR_THRESHOLD_MILLIS, then we set `state = STATE.DOOR_OPENED_START` then we will also fall into the `else` of on line 72, thus immediately setting `state = STATE.NO_PRESENCE_NO_SESSION`. So is the first `if` on lines 67-69 unnecessary? Or should something change here?

InnosentStateMachine.js
```diff
@@ -0,0 +1,128 @@
...
+          // eslint-disable-next-line eqeqeq
+          if (door.signal == DOOR_STATE.OPEN && session.od_flag == OD_FLAG_STATE.NO_OVERDOSE) {
+            state = STATE.DOOR_OPENED_CLOSE
+          } else if (inPhase_avg < mov_threshold) {
+            state = STATE.STILL
+          }
+          // eslint-disable-next-line eqeqeq
+          if (session.od_flag == OD_FLAG_STATE.NO_OVERDOSE && (await db.isOverdoseSuspectedInnosent(session, location_data))) {
+            state = STATE.SUSPECTED_OD
+          }
```
trisapeace, line 98: Here too, I find it strange that we would potentially overwrite `state` multiple times before leaving this `case` statement. Could the second `if` on line 96 be integrated into the `if-else`  in lines 90-94, potentially as the first `if`, thus making line 90 the first `else if(...)`?

InnosentStateMachine.js
```diff
@@ -0,0 +1,128 @@
...
+            state = STATE.DOOR_OPENED_CLOSE
+          } else if (inPhase_avg > mov_threshold) {
+            state = STATE.MOVEMENT
+          }
+          // eslint-disable-next-line eqeqeq
+          if (session.od_flag == OD_FLAG_STATE.NO_OVERDOSE && (await db.isOverdoseSuspectedInnosent(session, location_data))) {
+            state = STATE.SUSPECTED_OD
+          }
+
+          break
```
trisapeace, line 113: I think that these two `case` statements could be made into a single one:

```
case STATE.MOVEMENT:
case STATE.STILL:
  if (session.od_flag...) {
     state = STATE.SUSPECTED_OD
  } else if (door.signal...) {
     state = STATE.DOOR_OPENED_CLOSE
  } else if (inPhase_avg < mov_threshold) {
     state = STATE.STILL
  } else if (inPhase_avg > mov_threshold) {
    state = STATE.MOVEMENT
  }
  break;
```

Also, should one of `else if (inPhase_avg < mov_threshold) ` or `(inPhase_avg > mov_threshold)` be `<=` or `>=`?

db/011-add-radar-type-to-locations.sql
```diff
@@ -0,0 +1,21 @@
...
+    migrationId := 11;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        -- ADD SCRIPT HERE
+        ALTER TABLE locations ADD COLUMN radar_type text default 'XeThru';
```
trisapeace, line 15: What do you think about adding a line to update the `locations` table and set the `radar_type` column to `XeThru` whenever it's `NULL`.

db/db.js
```diff
@@ -294,6 +294,52 @@ async function updateSessionStillCounter(stillcounter, sessionid, locationid) {
   }
 }
 
+async function isOverdoseSuspectedInnosent(session, location) {
```
trisapeace, line 297: I suggest that we add `clientParam` to all our new functions in `db.js` so that we'll be able to use them in transactions without thinking about it.

db/db.js
trisapeace: Good catch.

index.js
```diff
@@ -166,8 +167,13 @@ async function checkHeartbeat() {
   }
 }
 
-async function handleSensorRequest(currentLocationId) {
-  const statemachine = new StateMachine(currentLocationId)
+async function handleSensorRequest(currentLocationId, radarType) {
+  let statemachine
+  if (radarType === 'XeThru') {
```
trisapeace, line 172: Can you please make `XeThru` and `Innosent` into an Enum and use it wherever you have these strings hardcoded?

db/db.js
```diff
@@ -294,6 +294,52 @@ async function updateSessionStillCounter(stillcounter, sessionid, locationid) {
...
+  const now = new Date()
```
trisapeace, line 298: What do you think about adding a DB call to get the database's time like @risatronic did here (https://github.com/bravetechnologycoop/BraveButtons/pull/88/files#diff-0e0c962c17d2dbb509faf124ae28f7a93341ee047ca1ebace5e1d0ee6f9d831eR846) and using it to get the value of `now`? In general, it can be risky to compare times that were generated on different servers. If you use the DB call, then both the "current time" and the `session.start_time` would have come from the same place.

db/db.js
```diff
@@ -294,6 +294,52 @@ async function updateSessionStillCounter(stillcounter, sessionid, locationid) {
...
+
+  // threshold values for the various overdose conditions
+  const still_threshold = location.still_threshold
+  const sessionDuration_threshold = location.duration_threshold
+
+  // number in front represents the weighting
+  const condition2 = 1 * (session.still_counter > still_threshold) // seconds
+  const condition3 = 1 * (sessionDuration > sessionDuration_threshold)
+
+  let alertReason = 'NotSet'
```
trisapeace, line 312: Can you please make these alertReasons into an Enum and then use them everywhere they are hardcoded?

db/db.js
```diff
@@ -401,6 +447,34 @@ async function getLocationIDFromParticleCoreID(coreID) {
   }
 }
 
+async function getRadarTypeFromLocationID(locationid) {
```
trisapeace, line 450: I don't see this function being used anywhere.

Also, I think that you don't need it. You could just use the `db.getLocationData` method and looks at its `radar_type`, right?

index.js
```diff
@@ -424,17 +430,36 @@ app.post('/api/xethru', Validator.body(['locationid', 'state', 'rpm', 'mov_f', '
   }
 })
 
+app.post('/api/innosent', async (request, response) => {
```
trisapeace, line 433: Can you please add an Express Validator to check for the `request.body.coreid` and `request.body.data`? Is it OK if the `inPhase` and `quadrature` are `undefined`?

index.js
```diff
@@ -424,17 +430,36 @@ app.post('/api/xethru', Validator.body(['locationid', 'state', 'rpm', 'mov_f', '
...
+  try {
+    const coreID = JSON.parse(request.body.coreid)
+    const location = await db.getLocationFromParticleCoreID(coreID)
+    const locationid = location.locationid
+    const data = JSON.parse(request.body.data)
+    const inPhase = data.inPhase
+    const quadrature = data.quadrature
+    await redis.addInnosentRadarSensorData(locationid, inPhase, quadrature)
+    await handleSensorRequest(locationid, 'Innosent')
+    response.status(200).json('Ok')
```
trisapeace, line 443: Normally the `OK` is in all caps, I think.

test/testStateMachine.js
```diff
@@ -7,7 +7,7 @@ const STATE = require('../SessionStateEnum.js')
 const XETHRU_STATE = require('../SessionStateXethruEnum.js')
```
trisapeace, line 7: Can you please change the name of this file to `testXeThruStateMachine.js`?

Also, can you please add a corresponding `testInnosentStateMachine.js` file?

Also, do you think we should extend the smoke tests to include cases that use the Innosent radar?

trisapeace (commented):
Nice. Looking good. I think that quite a few of my comments are things that were inherited from copy-and-pasting XeThru code.

db/db.js
```diff
@@ -401,6 +447,34 @@ async function getLocationIDFromParticleCoreID(coreID) {
...
+// Retrieves the locationid corresponding to a particle device coreID
+async function getLocationFromParticleCoreID(coreID) {
+  try {
+    const results = await pool.query('SELECT * FROM locations WHERE door_particlecoreid = $1 OR radar_particlecoreid = $1', [coreID])
+    if (results === undefined) {
+      helpers.log('Error: No location with associated coreID exists')
+      return null
+    }
+
+    return results.rows[0]
```
trisapeace, line 472: Might be nice to use `createLocationFromRow` to return a `Location` object instead. Up to you.

test/serverTest.js
```diff
@@ -200,6 +235,124 @@ describe('Brave Sensor server', () => {
     })
   })
 
+  describe('POST request radar and door events with mock im21 door sensor', () => {
```
trisapeace, line 238: I think this is the same name as above, can you change them to specify XeThru vs Innosent?

CHANGELOG.md
```diff
@@ -26,6 +32,7 @@ the code was deployed.
 
 - checkHeartbeat to run every 15 seconds to fix lack of alerts when sensors go down (CU-p34wcw)
 
+
```
trisapeace, line 35: For me, prettier removes these extra lines. I wonder if we have it set up differently on our machines.

InnosentStateMachine.js
```diff
@@ -0,0 +1,112 @@
...
+          return entry.inPhase
+        })
+        .reduce((a, b) => a + b) / innosentHistory.length
+
+    if (states === undefined) {
+      // In case the DB states table is empty create a RESET entry
+      await redis.addStateMachineData(STATE.RESET, this.location)
+      state = STATE.RESET
+    } else {
+      state = states.state // Takes current state in case that none of the state conditions get meet for a state transition.
```
trisapeace, line 37: Typo: "get meet" should be "are met"

db/db.js
```diff
@@ -302,55 +388,67 @@ async function updateSessionStillCounter(stillcounter, sessionid, locationid) {
...
+    }
 
-  // eslint-disable-next-line eqeqeq
-  if (condition2 == 1 || condition1 == 1) {
-    alertReason = 'Stillness'
-  }
+    // eslint-disable-next-line eqeqeq
+    if (condition2 == 1 || condition1 == 1) {
+      alertReason = ALERT_REASON.STILLNESS
+    }
```
trisapeace, line 420: These `==`s can be changed to `===`s like in the `isOverdoseSuspectedInnosent` function, right?

db/db.js
```diff
@@ -294,6 +321,65 @@ async function updateSessionStillCounter(stillcounter, sessionid, locationid) {
...
+
+        return true
+      } catch (err) {
+        helpers.log(`Error running od_flag update query`)
+      }
+    }
+
+    return false
+  } catch (e) {
+    helpers.log(`Error running isOverdoseSuspected: ${e}`)
```
trisapeace, line 379: This should be the `isOverdoseSuspectedInnosent` to match the name of the function.

db/db.js
```diff
@@ -294,6 +321,65 @@ async function updateSessionStillCounter(stillcounter, sessionid, locationid) {
...
+          alertReason,
+          session.sessionid,
+        ])
+        if (!transactionMode) {
+          client.release()
+        }
+
+        return true
+      } catch (err) {
+        helpers.log(`Error running od_flag update query`)
```
trisapeace, line 373: Do you want to output `${err}` here too?

db/db.js
```diff
@@ -302,55 +388,67 @@ async function updateSessionStillCounter(stillcounter, sessionid, locationid) {
...
+          OD_FLAG_STATE.OVERDOSE,
+          alertReason,
+          session.sessionid,
+        ])
+        if (!transactionMode) {
+          client.release()
+        }
+        return true
+      } catch (err) {
+        helpers.log(`Error running od_flag update query`)
```
trisapeace, line 445: Do you want to output `${err}` here too?

db/db.js
```diff
@@ -401,6 +499,21 @@ async function getLocationIDFromParticleCoreID(coreID) {
   }
 }
 
+// Retrieves the locationid corresponding to a particle device coreID
+async function getLocationFromParticleCoreID(coreID) {
```
trisapeace, line 503: Can you please add the `clientParam` stuff to this function too?

db/redis.js
```diff
@@ -79,6 +87,29 @@ function addXeThruSensorData(locationid, state, rpm, distance, mov_f, mov_s) {
...
+  for (let i = 0; i < inPhaseArray.length; i += 1) {
+    pipeline.xadd(`innosent:${locationid}`, '*', 'inPhase', inPhaseArray[i], 'quadrature', quadratureArray[i])
+  }
+  await pipeline.exec(err => {
+    helpers.log(err)
+  })
+}
+
+async function getInnosentWindow(locationID, startTime, endTime, windowLength) {
+  const rows = await client.xrevrange(`innosent:${locationID}`, startTime, endTime, 'count', windowLength)
```
trisapeace, line 103: Are these `await`s in this file also susceptible to the unhandle promise rejection warning? Maybe they need a try-catch.

index.js
```diff
@@ -192,27 +194,38 @@ async function checkHeartbeat() {
...
 
   for (const location of locations) {
+    let latestRadar
     // Query raw sensor data to transmit to the FrontEnd
-    const XeThruData = await redis.getLatestXeThruSensorData(location.locationid)
+    if (location.radarType === RADAR_TYPE.XETHRU) {
+      const XeThruData = await redis.getLatestXeThruSensorData(location.locationid)
+      latestRadar = Date(XeThruData.timestamp)
+    } else if (location.radarType === RADAR_TYPE.INNOSENT) {
+      const InnosentData = await redis.getLatestInnosentSensorData(location.locationid)
```
trisapeace, line 203: This looks strange to me to have these values start with a capital letter. They look like class names.

index.js
```diff
@@ -231,7 +244,7 @@ async function handleSensorRequest(currentLocationId) {
 
   if (location_start_time !== null && location_start_time !== undefined) {
     const start_time_sesh = new Date(location_start_time)
-    const now = new Date()
+    const now = new Date(await db.getCurrentTime())
```
trisapeace, line 247: Don't forget the try-catch around the db function call to avoid unhandle promise rejection warnings.

index.js
```diff
@@ -477,17 +490,61 @@ app.post('/api/xethru', Validator.body(['locationid', 'state', 'rpm', 'mov_f', '
...
+  Validator.check('data')
+    .custom(dataString => {
+      const data = JSON.parse(dataString)
+
+      const inPhase = (data || {}).inPhase
+      const quadrature = (data || {}).quadrature
+
+      return inPhase !== undefined && quadrature !== undefined
+    })
+    .withMessage('missing radar values, check for firmware or device integration errors'),
```
trisapeace, line 505: This is cool.

smokeTest.js
```diff
@@ -87,23 +117,49 @@ async function im21Door(coreID, signal) {
 async function smokeTest(recipientPhoneNumber, twilioPhoneNumber) {
   helpers.log('Smoke testing - please do not terminate this script before it finishes running')
   try {
-    await setup(recipientPhoneNumber, twilioPhoneNumber, testLocation1Id, door_coreID, radar_coreID)
+    helpers.log('Running XeThru Smoke Tests')
```
trisapeace, line 120: Could you please update the README to include what we expect when we run the smoke tests?

smokeTest.js
```diff
@@ -25,18 +26,47 @@ async function teardown() {
...
       twilioNumber: twilioPhoneNumber,
+      radarType: type,
     })
   } catch (e) {
     helpers.log(e)
   }
 }
 
-async function silence(locationid) {
+async function innosentMovement(coreID, min, max) {
```
trisapeace, line 41: Yay! Thanks for updating this!

test/serverTest.js
```diff
@@ -200,6 +236,124 @@ describe('Brave Sensor server', () => {
...
+        5000,
+        0,
+        '+15005550006',
+        '+15005550006',
+        '+15005550006',
+        1000,
+        'locationName',
+        door_coreID,
+        radar_coreID,
+        'Innosent',
```
trisapeace, line 261: You could use the `RADAR_TYPE` enum for this.

test/testInnosentStateMachine.js
```diff
@@ -0,0 +1,275 @@
...
+      return innosent
+    },
+    getLatestDoorSensorData() {
+      return door
+    },
+    addStateMachineData() {},
+    getInnosentWindow() {
+      return innosentHistory
+    },
+  }
```
trisapeace, line 41: Hahaha, old school. I know you just copied this from the old state machine tests, but I find it amusing because these were essentially mocks before we added the `sinon` library. Not worth changing, but funny to see.

test/testInnosentStateMachine.js
```diff
@@ -0,0 +1,275 @@
...
+}
+
+describe('test getNextState', () => {
```
trisapeace, line 44: Since this is copied from `testXeThruStateMachine.js`, it has the same top level name. I see that you've specified innosent in a few of the lower-level test names, but I think it should be done here too otherwise it might not be clear which `test getNextState` --> `when no initial state` --> `should return the RESET state` test was run, for example.

trisapeace (approved):
My comments are just for minor changes. Looks great.

PR #94, opened 2021-03-29
Taking absolute value of INS radar readings for computing state (CU-qv7qqq)

mariocimet:


trisapeace:
Task linked: [CU-qv7qqq Use absolute value in INS average](https://app.clickup.com/t/qv7qqq)

trisapeace (approved):
This makes sense to me. Although I'm still seeing more Stills with the INS than my XeThru. But maybe that's due to the threshold on my XeThru?

PR #96, opened 2021-04-14
Do transaction-handling and use Location/Session objects consistently across all DB functions (CU-kry5t5)

trisapeace:
In the end, I'm pretty happy with this `runQuery` function. I'd tried a few different ways of making this boilerplate in the past but didn't like any of them. This feels pretty straightforward though, so I'm happy. The only annoying part was still needing to put the `await runQuery(...)` call inside a `try/catch` because of the `await` in order to avoid potential `Unhandled Promise Rejection Warning`s. Oh well, it's still less copy-and-pasting than before. What do you think?

In line with the transaction boilerplate, I added `try/catch` blocks around all of our uses of actual transactions in the code, including a call to `rollbackTransaction` in the `catch`.

I also took this opportunity to make sure that we are consistently using the `Location` and `Session` object for all DB functions that return locations or sessions. This meant quite a few changes all over the place, so please let me know if you see anywhere that I've missed!

### DIFF
--- a/db/db.js
+++ b/db/db.js
@@ -25,46 +25,15 @@ pool.on('error', err => {
   helpers.logError(`unexpected database error: ${JSON.stringify(err)}`)
 })
 
-async function getCurrentTime(clientParam) {
-  let client = clientParam
-  const transactionMode = typeof client !== 'undefined'
-
-  try {
-    if (!transactionMode) {
-      client = await pool.connect()
-    }
-
-    const { rows } = await client.query('SELECT NOW()')
-    const time = rows[0].now
-
-    return time
-  } catch (e) {
-    helpers.log(`Error running the getCurrentTime query: ${e}`)
-  } finally {
-    if (!transactionMode) {
-      try {
-        client.release()
-      } catch (err) {
-        helpers.log(`getCurrentTime: Error releasing client: ${err}`)
-      }
-    }
-  }
-}
-
-// Functions added to facilitate moving to Mustache template from angular front end
-// prettier-ignore
 function createSessionFromRow(r) {
+  // prettier-ignore
   return new Session(r.locationid, r.start_time, r.end_time, r.od_flag, r.state, r.phonenumber, r.notes, r.incidenttype, r.sessionid, r.duration, r.still_counter, r.chatbot_state, r.alert_reason)
 }
 
-// prettier-ignore
 function createLocationFromRow(r) {
+  // prettier-ignore
   return new Location(r.locationid, r.display_name, r.phonenumber, r.sensitivity, r.led, r.noisemap, r.mov_threshold, r.duration_threshold, r.still_threshold, r.rpm_threshold, r.heartbeat_sent_alerts, r.heartbeat_alert_recipient, r.door_particlecoreid, r.radar_particlecoreid, r.radar_type, r.reminder_timer, r.fallback_timer, r.auto_reset_threshold, r.twilio_number, r.fallback_phonenumbers, r.door_stickiness_delay)
 }
-
-// The following functions will route HTTP requests into database queries
-
-// GET all XeThru data
 
 async function beginTransaction() {
   const client = await pool.connect()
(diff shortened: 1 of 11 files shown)


db/db.js
```diff
@@ -78,262 +47,275 @@ async function commitTransaction(client) {
...
     }
 
-    return results.rows
-  } catch (e) {
-    helpers.log(`Error running the getAllSessionsFromLocation query: ${e}`)
-  }
-}
-
-// Gets the last session data from an unclosed session for a specified location
-async function getLastUnclosedSession(locationid) {
```
trisapeace, line 182: I couldn't see anywhere where this function was being used. So I remove it.

index.js
```diff
@@ -273,61 +281,96 @@ async function handleSensorRequest(currentLocationId, radarType) {
...
+        const latestSession = await db.getMostRecentSession(currentLocationId, client)
+
+        if (typeof latestSession !== 'undefined') {
+          // Checks if no session exists for this location yet.
+          if (latestSession.endTime == null) {
+            // Checks if session is open.
+            await db.updateSessionState(latestSession.sessionid, currentState, client)
+          }
+        }
+        await db.commitTransaction(client)
```
trisapeace, line 296: This didn't have a transaction before, but I think it makes sense for it to have one.

trisapeace:
@mariocimet I just deployed this to dev, so it's there if you want to try it out.

trisapeace:
Task linked: [CU-kry5t5 Refactor `client.release()` calls into a `finally` block in `db.js`](https://app.clickup.com/t/kry5t5)

trisapeace:
@schwarrrtz I added you as a reviewer to this Sensors PR because it's a change that I'd like to make in Buttons as well and I'd like your opinion on it. It's factoring out the transaction boilerplate stuff in db.js and using defined objects for the results from DB queries.
If it's good for both projects, I'll probably move the `runQuery` function into `helpers` in `brave-alert-lib`.

index.js
mariocimet: agree

db/db.js
```diff
@@ -78,262 +47,275 @@ async function commitTransaction(client) {
...
-    }
-
-    await client.query('UPDATE sessions SET end_time = CURRENT_TIMESTAMP WHERE sessionid = $1', [sessionid])
-    client.query("UPDATE sessions SET duration = TO_CHAR(age(end_time, start_time),'HH24:MI:SS') WHERE sessionid = $1", [sessionid]) // Sets the duration to the difference between the end and start time
-    if (!transactionMode) {
-      client.release()
-    }
-  } catch (e) {
-    helpers.log(`Error running the updateSessionEndTime query: ${e}`)
+    await runQuery('updateSessionEndTime', 'UPDATE sessions SET end_time = CURRENT_TIMESTAMP WHERE sessionid = $1', [sessionid], clientParam)
```
mariocimet, line 204: this is really a lot nicer. I don't think even the try/catch is really a strike against it because we need to handle errors for the queries anyways so its not even a net addition. Quality improved 🌟

index.js
```diff
@@ -273,61 +281,96 @@ async function handleSensorRequest(currentLocationId, radarType) {
...
+      let client
+      try {
+        client = await db.beginTransaction()
+        const latestSession = await db.getMostRecentSession(currentLocationId, client)
+
+        if (typeof latestSession !== 'undefined') {
+          // Checks if session exists
+          if (latestSession.endTime == null) {
+            // Checks if session is open for this location
+            const currentSession = await db.updateSessionState(latestSession.sessionid, currentState, client)
```
mariocimet, line 316: Here and in the updateSessionState function we're still using the column name and returning a results.rows[0] without running it through createSessionFromRow

index.js
```diff
@@ -273,61 +281,96 @@ async function handleSensorRequest(currentLocationId, radarType) {
...
+            start_times[currentLocationId] = currentSession.start_time
+          } else {
+            const currentSession = await db.createSession(location.phonenumber, currentLocationId, currentState, client)
+            start_times[currentLocationId] = currentSession.start_time
+          }
         } else {
           const currentSession = await db.createSession(location.phonenumber, currentLocationId, currentState, client)
           start_times[currentLocationId] = currentSession.start_time
         }
```
mariocimet, line 325: Here in the updateSessionState and createSession functions we're still using the column name and returning a results.rows[0] without running it through createSessionFromRow

trisapeace: Good catch! I'll fix those up.
That's what happens when I search for `FROM sessions` and `FROM locations` and forget to search for `RETURNING *` :smile:

mariocimet: other than this one, all the other db functions returning raw results.rows are either getCurrentTime, or they're not assigning or doing anything with the return value, so I think they're fine.

mariocimet: ah! yeah, my search was for results.rows[0] in db.js, with a space prefix to differentiate it from when we feed it to the object constructor

mariocimet:
pre-approving for when you correct the ones in lines 316-325 of index :)

mariocimet (approved):

index.js
trisapeace: Fixed.

trisapeace:
Deployed the changes to dev.sensors.brave.coop and ran the smoketests. Seems fine.

schwarrrtz:
I'm not familiar enough with the Sensors server codebase to fully review this, but I like the `runQuery` refactor and would be happy to see this in the Buttons server as well. My only comment would be that the `catch` blocks that just contain `// Errors handled in runQuery` could have oneline log message instead - probably all the errors will indeed be handled in `runQuery` but if not then it would be good to know about it.

trisapeace:
Thanks for looking @schwarrrtz . That's great.

I was really hoping to get the boilerplate in each function down as little as possible. But I see your point. I'll add log messages to the catches.

schwarrrtz (approved):
thanks for updating the `runQuery` function @trisapeace - to me a oneline log message is no more boilerplate than the oneline comment it replaces.

PR #98, opened 2021-04-20
Add Sentry (CU-px7k6e)

mariocimet:


trisapeace:
Task linked: [CU-px7k6e Add Sentry](https://app.clickup.com/t/px7k6e)

index.js
```diff
@@ -38,6 +40,12 @@ const locationFormCSSPartial = fs.readFileSync(`${__dirname}/mustache-templates/
 // Start Express App
 const app = express()
 
+// Start Sentry
+Sentry.init({
+  dsn: helpers.getEnvVar('SENTRY_DSN'),
+  integrations: [new Tracing.Integrations.Postgres(), new Sentry.Integrations.Http({ tracing: true }), new Tracing.Integrations.Express({ app })],
+})
+
```
trisapeace, line 48: Can this be done in Brave Alert Lib? Maybe a new Sentry module that is exported in `index.js` that is then used here by calling a `sentryHelper.setup()` or something.

index.js
```diff
@@ -62,6 +70,10 @@ const CLOSINGSTATES = [STATE.DOOR_OPENED_CLOSE]
 // These states will start a Brave Alert session for a location
 const ALERTSTARTSTATES = [STATE.SUSPECTED_OD]
 
+// Sentry monitoring Middleware
+app.use(Sentry.Handlers.requestHandler())
+app.use(Sentry.Handlers.tracingHandler())
+
```
trisapeace, line 76: What do you think about putting this also into a `sentryHelper.setup()` function in Brave Alert Lib?

index.js
```diff
@@ -867,10 +879,13 @@ app.post('/smokeTest/teardown', async (request, response) => {
...
     response.status(200).send()
   } catch (error) {
-    helpers.log(`Smoke test setup error: ${error}`)
+    helpers.logError(`Smoke test setup error: ${error}`)
   }
 })
 
+// The error handler must be before any other error middleware and after all controllers
+app.use(Sentry.Handlers.errorHandler())
+
```
trisapeace, line 888: Hmmm, this too would be nice to have in Brave Alert Lib, although by your comment here it should like maybe it's a function separate from `setupSentry`, maybe `setupSentryErrorHandler`. Funny requirement for its placement.

package.json
```diff
@@ -15,6 +15,8 @@
     "lib/*"
   ],
   "dependencies": {
+    "@sentry/node": "^6.2.5",
+    "@sentry/tracing": "^6.2.5",
```
trisapeace, line 19: If we can move everything into Brave Alert Lib, then we won't need to do this explicitly, it'll be there automatically as a dependency of `brave-alert-lib`. Like how I was able to remove the direct `dotenv` dependency from Buttons and Sensors.

mariocimet:
This is now up on https://staging.sensors.brave.coop for poking and for creating error conditions. On the other side you can look at the Sentry console here https://sentry.io/organizations/brave-technology-coop/issues/?project=3012816. At first I had this set up with the beta test database but nothing in redis, so it was spitting out errors. It showed up on the console and also sent me a nicely labelled email about it.

trisapeace (approved):
Looks good to me :smile:

PR #100, opened 2021-04-23
Fix alertSessionChangedCallback to not overwrite session with nulls (CU-tnarab)

trisapeace:
- After the fallback message is sent by brave-alert-lib, it calls
  alertSessionChangedCallback with an alertSession object containing
  ONLY the session ID and the results from calling Twilio. The code
  was taking that and blindly setting the chatbot state and
  incidenttype based on that alertSession object, thus setting those
  values to null every time the fallback message is sent. Now it tries
  to be smarter and will re-save the current values if given undefined

### DIFF
--- a/BraveAlerterConfigurator.js
+++ b/BraveAlerterConfigurator.js
@@ -55,13 +55,17 @@ class BraveAlerterConfigurator {
   }
 
   async alertSessionChangedCallback(alertSession) {
-    const incidentType = incidentTypes[incidentTypeKeys.indexOf(alertSession.incidentCategoryKey)]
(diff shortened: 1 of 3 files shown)


trisapeace:
Task linked: [CU-tnarab Should not clear the chatbot state after sending the fallback message](https://app.clickup.com/t/tnarab)

trisapeace:
This is up on `dev.sensors.brave.coop` and I've looked over:
- Smoke tests
- Manually sat in my bathroom a few times and responded or didn't respond at various points in the chatbot flow
   - Received fallback messages and saw that it didn't change the chatbot state or incidenttype for both my INS and XeThru

mariocimet:
Looks good

mariocimet (approved):

PR #104, opened 2021-05-10
Add ability to set Locations to inactive (CU-rk8axg)

trisapeace:
- Inactive locations will not send out duration, stillness, or heartbeat
  alerts or change state, but they will still record sensor and vitals
  data and will still notify Sentry of low batteries in door sensors
- Locations created through the UI will be ~active~ inactive by default. But existing Locations will be active.
- Added a column on the Landing Dashboard to indicate if a Location
  is inactive
- Added a note on the Add A Location page to warn users that newly created Locations will be inactive.

Things I tested:
- Given is_active set to false and no open sessions. When I closed the door and moved around for longer than the duration_timer. Then no session was started, no alerts were sent, and no state changes were logged.
- Given is_active set to false and there is a session and the od_flag is false. When I open and close the door. Then the session stays open, no alerts are sent, and no state changes were logged.
- Given is_active set to false and there is a session and the od_flag is false. When I receive the alert. Then I am able to finish the chatbot session as normal to close the session.
- Given is_active set to false and there is a session and the od_flag is false. When I wait longer than the auto-reset threshold. Then the session is not auto-reset and no auto-reset message is sent.
- Toggling is_active between true and false at various points to see if I could get it into a weird state, but it all seemed to make sense.
- Given is_active set to false. When I disconnect the device. Then no disconnection alert is sent to me and no disconnection log is logged in Sentry.
- Given is_active set to false. When I reconnect the device. Then no reconnection alert is sent to me and no reconnection log is logged in Sentry.

Didn't know how to test:
- Low door battery alert

### DIFF
--- a/db/017-add-is-active-to-locations.sql
+++ b/db/017-add-is-active-to-locations.sql
@@ -1,0 +1,26 @@
+DO $migration$
+    DECLARE migrationId INT;
+    DECLARE lastSuccessfulMigrationId INT;
+BEGIN
+    -- The migration ID of this file
+    migrationId := 17;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        -- ADD SCRIPT HERE
+        ALTER TABLE locations 
+        ADD COLUMN is_active boolean
+        DEFAULT false;
+
+        UPDATE locations
+        SET is_active = true;
+
+        -- Update the migration ID of the last file to be successfully run to the migration ID of this file
+        INSERT INTO migrations (id)
+        VALUES (migrationId);
+    END IF;
+END $migration$;
--- a/db/db.js
+++ b/db/db.js
@@ -32,7 +32,7 @@ function createSessionFromRow(r) {
 
 function createLocationFromRow(r) {
   // prettier-ignore
-  return new Location(r.locationid, r.display_name, r.phonenumber, r.sensitivity, r.led, r.noisemap, r.mov_threshold, r.duration_threshold, r.still_threshold, r.rpm_threshold, r.heartbeat_sent_alerts, r.heartbeat_alert_recipients, r.door_particlecoreid, r.radar_particlecoreid, r.radar_type, r.reminder_timer, r.fallback_timer, r.auto_reset_threshold, r.twilio_number, r.fallback_phonenumbers, r.door_stickiness_delay, r.alert_api_key)
+  return new Location(r.locationid, r.display_name, r.phonenumber, r.sensitivity, r.led, r.noisemap, r.mov_threshold, r.duration_threshold, r.still_threshold, r.rpm_threshold, r.heartbeat_sent_alerts, r.heartbeat_alert_recipients, r.door_particlecoreid, r.radar_particlecoreid, r.radar_type, r.reminder_timer, r.fallback_timer, r.auto_reset_threshold, r.twilio_number, r.fallback_phonenumbers, r.door_stickiness_delay, r.alert_api_key, r.is_active)
 }
 
 async function beginTransaction() {
@@ -164,12 +164,12 @@ async function getHistoryOfSessions(locationid, clientParam) {
   }
 }
 
-async function getAllSessionsFromLocation(location, clientParam) {
+async function getAllSessionsFromLocation(locationid, clientParam) {
   try {
     const results = await runQuery(
       'getAllSessionsFromLocation',
       'SELECT * FROM sessions WHERE locationid = $1 order by start_time desc',
-      [location],
+      [locationid],
       clientParam,
     )
 
@@ -245,12 +245,12 @@ async function updateSessionState(sessionid, state, clientParam) {
 }
 
 // Updates the value of the alert flag in the location database
-async function updateSentAlerts(location, sentalerts, clientParam) {
+async function updateSentAlerts(locationid, sentalerts, clientParam) {
   try {
     const results = await runQuery(
       'updateSentAlerts',
       'UPDATE locations SET heartbeat_sent_alerts = $1 WHERE locationid = $2 RETURNING *',
-      [sentalerts, location],
+      [sentalerts, locationid],
       clientParam,
     )
     if (results === undefined) {
@@ -482,6 +482,20 @@ async function getLocationFromParticleCoreID(coreID, clientParam) {
   }
 }
 
+async function getActiveLocations(clientParam) {
+  try {
+    const results = await runQuery('getLocations', 'SELECT * FROM locations WHERE is_active', [], clientParam)
+
+    if (typeof results === 'undefined') {
+      return null
+    }
+
+    return results.rows.map(r => createLocationFromRow(r))
+  } catch (err) {
+    helpers.log(JSON.stringify(err))
+  }
+}
+
 // Retrieves the locations table
 async function getLocations(clientParam) {
   try {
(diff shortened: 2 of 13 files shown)


trisapeace:
Task linked: [CU-rk8axg Sensors should be able to deactivate a location without removing it from the database](https://app.clickup.com/t/rk8axg)

db/db.js
```diff
@@ -164,12 +164,12 @@ async function getHistoryOfSessions(locationid, clientParam) {
...
 }
 
-async function getAllSessionsFromLocation(location, clientParam) {
+async function getAllSessionsFromLocation(locationid, clientParam) {
   try {
     const results = await runQuery(
       'getAllSessionsFromLocation',
       'SELECT * FROM sessions WHERE locationid = $1 order by start_time desc',
-      [location],
+      [locationid],
```
trisapeace, line 172: Changed this variable name to make it clearer that what we're passing is just the Location's ID, not the whole Location object.

db/db.js
```diff
@@ -245,12 +245,12 @@ async function updateSessionState(sessionid, state, clientParam) {
...
 
 // Updates the value of the alert flag in the location database
-async function updateSentAlerts(location, sentalerts, clientParam) {
+async function updateSentAlerts(locationid, sentalerts, clientParam) {
   try {
     const results = await runQuery(
       'updateSentAlerts',
       'UPDATE locations SET heartbeat_sent_alerts = $1 WHERE locationid = $2 RETURNING *',
-      [sentalerts, location],
+      [sentalerts, locationid],
```
trisapeace, line 253: Changed this variable name to make it clearer that what we're passing is just the Location's ID, not the whole Location object.

index.js
```diff
@@ -208,7 +208,7 @@ async function sendResetAlert(locationid) {
 
 // Heartbeat Helper Functions
 async function checkHeartbeat() {
-  const locations = await db.getLocations()
+  const locations = await db.getActiveLocations()
```
trisapeace, line 211: No need to check heartbeat for inactive Locations. In fact, that something is known-down might be the reason why we set a Location to inactive for a period of time.

index.js
```diff
@@ -151,7 +151,7 @@ async function autoResetSession(locationid) {
 // This function seeds the state table with a RESET state in case there was a prior unresolved state discrepancy
 if (!helpers.isTestEnvironment()) {
   setInterval(async () => {
-    const locations = await db.getLocations()
+    const locations = await db.getActiveLocations()
```
trisapeace, line 154: Doesn't check inactive sessions for the auto reset conditions because we don't want to send out any auto-reset messages.

index.js
```diff
@@ -238,25 +238,24 @@ async function checkHeartbeat() {
...
   const currentState = await statemachine.getNextState(db, redis)
-  const stateobject = await redis.getLatestLocationStatesData(currentLocationId)
+  const stateobject = await redis.getLatestLocationStatesData(location.locationid)
   let prevState
   if (!stateobject) {
     prevState = STATE.RESET
   } else {
     prevState = stateobject.state
   }
-  const location = await db.getLocationData(currentLocationId)
```
trisapeace, line 256: Moved this outside of the function `handleSensorRequest` and made it the responsibility of the caller to get the Location and pass it in because that way the caller could check whether the Location was active and only call this function if it is. Also, two out of the three places where `handleSensorRequest` was being called already had the whole Location object, so this is more efficient to not have this second DB call.

test/integration/serverTest.js
```diff
@@ -864,8 +1024,8 @@ describe('Brave Sensor server', () => {
 
     describe('POST /locations/:locationId', () => {
       beforeEach(async () => {
-        sinon.stub(helpers, 'log')
-        sinon.stub(helpers, 'logError')
+        sinon.spy(helpers, 'log')
+        sinon.spy(helpers, 'logError')
```
trisapeace, line 1028: I found it useful when debugging these tests to have the logs still be printed out to the console, so I changed this from stub to spy.

trisapeace:
@mariocimet This one is up on Dev and ready for review. When you go through it, could you please also try out a few things on Dev? It has a lot of moving pieces and I want to make sure that it doesn't get us in any weird situations. Thanks!

mariocimet (commented):
As requested I will think through this as if I was implementing from scratch. So I'll start by listing what my todo list would be.

- a column that's a boolean flag (and cascading db changes for creating, updating, and turning locations into objects)
- a field/toggle in the screen for creating and updating locations
- conditionals that check this feature of the location at the relevant points
	- data handlers - xethru, door sensor, ins endpoint - run the "main loop" or not
	- displaying data (should mark inactive)
	- running the heartbeat or not

mariocimet:
I would probably test low battery alert with curl/postman/axios and sending the 'LOW_BATT: '04' data byte, and watch for the Sentry message that should come in for that.

index.js
```diff
@@ -677,6 +677,7 @@ app.post(
           data.reminderTimer,
           data.fallbackTimer,
           newAlertApiKey,
+          data.isActive === 'true',
```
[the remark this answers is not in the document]
mariocimet: ah, I see, this is for new locations. So I'll punt the question to design, whats the thought with defaulting to active? What about the use case of say, queuing up a location?

mariocimet:
I request that we change from defaulting to false, to giving the option of setting inactive from the get go

index.js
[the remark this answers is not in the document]
trisapeace: This is just changing the string `data.isActive` into a real JS boolean value from the UI form.

db/017-add-is-active-to-locations.sql
```diff
@@ -0,0 +1,23 @@
...
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        -- ADD SCRIPT HERE
+        ALTER TABLE locations 
+        ADD COLUMN is_active boolean
+        DEFAULT true;
```
trisapeace, line 17: Change this to `DEFAULT false`

mustache-templates/landingPage.mst
```diff
@@ -23,13 +23,15 @@
...
                             <th scope="col"><h5>Last Session Started</h5></th>
                         </tr>
                     </thead>
                     <tbody>
                     {{#locations}}
                         <tr class="table-data">
+                            <td>{{^isActive}}[INACTIVE]{{/isActive}}</td>
                             <th scope="row"><a href="/locations/{{id}}">{{name}}</a></th>
                             <td>{{sessionStart}}</td>
                         </tr>
```
trisapeace, line 37: Add note about inactive directly above the Submit button.

db/017-add-is-active-to-locations.sql
trisapeace: Done. 

And I also added a query to set the existing Locations' is_active flag to true.

mustache-templates/landingPage.mst
trisapeace: Done.

mariocimet:
Tested the low battery warning in both active and inactive cases. For reference I just curled a JSON blob that I copied from the Particle console (on the page with the IM21 Data integration), replacing the coreid and the signal for the device I wanted to spoof and to have the low battery message, like so:

`curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"data": "{ \"deviceid\": \"36:F1:51\", \"data\": \"04\", \"control\": \"2C\" }","coreid": "e00fce68e1ee9f139bf40656"}' \
  https://dev.sensors.brave.coop/api/door`

mariocimet:
repeated the tests with the heartbeat messages and confirmed expected behavior

index.js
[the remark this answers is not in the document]
mariocimet: ah ok, I finally understood what this actually does, it supplies a boolean argument (i.e. "true" in a javascript sense if the string is === to 'true' and "false" as a javascript boolean otherwise). Got it.

mariocimet (approved):

trisapeace:
Thanks @mariocimet . Really appreciate the thorough review and testing!

PR #108, opened 2021-05-14
Only update start_times array while we have a lock on sessions (CU-vf7z0f)

trisapeace:
- This will prevent a race condition that was causing never-ending
  sessions when a first session (S1) was auto-reset while having
  a lock on the sessions table, but then it gave up its lock before
  setting start_times[locationid] to null for that location. This
  allowed a new session to be created between those two operations
  that then had its start_times[locationid] set to null essentially
  right after it started. This left us in a position where the
  sessionDuration calcuation would return `undefined` and we would
  never auto-reset the new session.

Tested this by:
- Running the smoke tests and replying to all the texts
- Running the smoke tests and not replying to any of the texts in order to see the auto-reset happen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ the code was deployed.
 
 ### Fixed
 - Helm chart RELEASE variable error when the image tag happened to be an integer.
+- Race condition in auto-reset (CU-vf7z0f).
 
 ## [3.2.0] - 2021-04-26
 
--- a/index.js
+++ b/index.js
@@ -135,6 +135,7 @@ async function autoResetSession(locationid) {
     await db.closeSession(currentSession.sessionid, client)
     await db.updateSessionResetDetails(currentSession.sessionid, 'Auto reset', 'Reset', client)
     redis.addStateMachineData('Reset', locationid)
+    start_times[locationid] = null
     await db.commitTransaction(client)
   } catch (e) {
     helpers.log('Could not reset open session')
@@ -270,10 +271,8 @@ async function handleSensorRequest(currentLocationId, radarType) {
   }
 
   // If session duration is longer than the threshold (20 min), reset the session at this location, send an alert to notify as well.
-
   if (sessionDuration * 1000 > location.autoResetThreshold) {
     autoResetSession(location.locationid)
-    start_times[currentLocationId] = null
     helpers.log(`${currentLocationId}: autoResetSession has been called`)
     sendResetAlert(location.locationid)
   }


trisapeace:
Task linked: [CU-vf7z0f Troubleshoot never-ending Sensor Sessions on Dev](https://app.clickup.com/t/vf7z0f)

trisapeace:
I just tried this with a modified smoke test to show that new sessions can still be created and finished successfully after a reset. This doesn't check that the race condition doesn't happen (because it's hard to make it end up in a race condition), but it does show that regular sessions still behave the way we expect.

```
try {
    helpers.log('Running XeThru Smoke Tests')
    await teardown()
    await setup(recipientPhoneNumber, twilioPhoneNumber, RADAR_TYPE.XETHRU)
    for (let i = 0; i < 15; i += 1) {
      await xeThruMovement(testLocation1Id, getRandomInt(MOV_THRESHOLD + 1, 100), getRandomInt(MOV_THRESHOLD + 1, 100))
    }
    await im21Door(door_coreID, IM21_DOOR_STATUS.OPEN)
    for (let i = 0; i < 5; i += 1) {
      await xeThruMovement(testLocation1Id, getRandomInt(0, MOV_THRESHOLD), getRandomInt(0, MOV_THRESHOLD))
    }
    await im21Door(door_coreID, IM21_DOOR_STATUS.CLOSED)
    for (let i = 0; i < 15; i += 1) {
      await xeThruMovement(testLocation1Id, getRandomInt(0, MOV_THRESHOLD), getRandomInt(0, MOV_THRESHOLD))
    }
    for (let i = 0; i < 15; i += 1) {
      await xeThruMovement(testLocation1Id, getRandomInt(MOV_THRESHOLD + 1, 100), getRandomInt(MOV_THRESHOLD + 1, 100))
    }
    for (let i = 0; i < 85; i += 1) {
      await xeThruSilence(testLocation1Id)
    }
    for (let i = 0; i < 15; i += 1) {
      await xeThruMovement(testLocation1Id, getRandomInt(MOV_THRESHOLD + 1, 100), getRandomInt(MOV_THRESHOLD + 1, 100))
    }
    await im21Door(door_coreID, IM21_DOOR_STATUS.OPEN)
    for (let i = 0; i < 5; i += 1) {
      await xeThruMovement(testLocation1Id, getRandomInt(0, MOV_THRESHOLD), getRandomInt(0, MOV_THRESHOLD))
    }
    await im21Door(door_coreID, IM21_DOOR_STATUS.CLOSED)
  } catch (error) {
    helpers.log(`Error running smoke test: ${error}`)
  } finally {
    await teardown()
  }
  ```

mariocimet (approved):

mariocimet:
looks good

PR #109, opened 2021-05-22
Always send 200 to Particle and reduce thrown errors in DB functions (CU-m0we6c)

trisapeace:
**Change all endpoints that Particle uses to always return 200**
- Apparently Particle throttles sending messages when it returns too
  many 400s and 500s. So to make sure that we're getting all the
  message, and since we don't actually do anything with the return
  status/message anyways, it seems to make sense to return 200 and
  not worry about it.
- Ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits

**Reduce thrown errors in DB functions**
- A few of our DB functions are supposed to return a single object (Location or Session), which means that they use `results.rows[0]`. But if there were no results from the query, there is no `results.rows[0]`, thus throwing an error. Which is  then immediately caught and the very un-helpful error message `{}` is output to the logs.
- Now all of these functions check for this case and consistently return `null`
- Had to update some of the checks for the return value to check for `null` now instead of `undefined`

What I did to test this on `dev.sensors.brave.coop`:
- Ran smoke tests
- Watched the logs while there is an INS sending messages that is not in the `Locations` table. They are now all coming in at a 2-second interval as expected.
- Started a real session in my bathroom with my INS while there is another INS sending messages that are not in the `Locations` table. Saw that my movement and stillness were coming in frequently at 2-second intervals (we don't expect this always since it should only log when there is a change of state)
- Looked at the Cluster Insights in Digital Ocean and didn't see anything strange. But also didn't see anything strange while it was working incorrectly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,16 @@ the code was deployed.
 - Concept of active/inactive locations (CU-rk8axg)
 
 ### Changed
+
 - Improved some API error messages.
 - Heartbeat text now contains troubleshooting instructions and contact email for follow-up. (CU-rx46w0)
 - Do not change state or send alerts for inactive locations (CU-rk8axg)
+- Always return 200 to Particle when it calls our APIs (CU-m0we6c)
 
 ### Fixed
-- Helm chart RELEASE variable error when the image tag happened to be an integer.
-- Race condition in auto-reset (CU-vf7z0f).
+
+- Helm chart RELEASE variable error when the image tag happened to be an integer
+- Race condition in auto-reset (CU-vf7z0f)
 
 ## [3.2.0] - 2021-04-26
 
--- a/db/db.js
+++ b/db/db.js
@@ -63,7 +63,7 @@ async function rollbackTransaction(client) {
 
 async function runQuery(functionName, queryString, queryParams, clientParam) {
   let client = clientParam
-  const transactionMode = typeof client !== 'undefined'
+  const transactionMode = client !== undefined
 
   try {
     if (!transactionMode) {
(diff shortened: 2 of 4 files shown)


trisapeace:
Task linked: [CU-m0we6c Real life testing of state machine to work with INS data](https://app.clickup.com/t/m0we6c)

CHANGELOG.md
```diff
@@ -19,13 +19,16 @@ the code was deployed.
...
 - Heartbeat text now contains troubleshooting instructions and contact email for follow-up. (CU-rx46w0)
 - Do not change state or send alerts for inactive locations (CU-rk8axg)
+- Always return 200 to Particle when it calls our APIs (CU-m0we6c)
 
 ### Fixed
-- Helm chart RELEASE variable error when the image tag happened to be an integer.
-- Race condition in auto-reset (CU-vf7z0f).
+
+- Helm chart RELEASE variable error when the image tag happened to be an integer
+- Race condition in auto-reset (CU-vf7z0f)
```
trisapeace, line 31: All I did here was remove the periods at the end since we haven't been using periods in Sensors.

db/db.js
```diff
@@ -63,7 +63,7 @@ async function rollbackTransaction(client) {
 
 async function runQuery(functionName, queryString, queryParams, clientParam) {
   let client = clientParam
-  const transactionMode = typeof client !== 'undefined'
+  const transactionMode = client !== undefined
```
trisapeace, line 66: For consistency with how we do this comparison elsewhere in the codebase. A good discussion here: https://stackoverflow.com/questions/3390396/how-can-i-check-for-undefined-in-javascript#answer-3390635

mariocimet (approved):

mariocimet:
Looks good, I like the simplified null checks as well.

PR #110, opened 2021-05-27
Storing IM21 heartbeat messages and checking for IM21 heartbeat (CU-wf97dy)

mariocimet:
now up on dev-sensor-server


trisapeace:
Task linked: [CU-wf97dy check IM21 heartbeat](https://app.clickup.com/t/wf97dy)

index.js
```diff
@@ -24,6 +24,7 @@ const IM21_DOOR_STATUS = require('./IM21DoorStatusEnum')
 const SESSIONSTATE_DOOR = require('./SessionStateDoorEnum')
 
 const RADAR_THRESHOLD_MILLIS = 30 * 60 * 1000
+const DOOR_THRESHOLD_MILLIS = 30 * 60 * 1000
```
trisapeace, line 27: Can these be environment variables?

db/db.js
```diff
@@ -249,11 +249,11 @@ async function updateSessionState(sessionid, state, clientParam) {
 }
 
 // Updates the value of the alert flag in the location database
-async function updateSentAlerts(locationid, sentalerts, clientParam) {
+async function updateSentAlerts(locationid, sentalerts, component, clientParam) {
   try {
     const results = await runQuery(
       'updateSentAlerts',
-      'UPDATE locations SET heartbeat_sent_alerts = $1 WHERE locationid = $2 RETURNING *',
+      `UPDATE locations SET ${component}_heartbeat_sent_alerts = $1 WHERE locationid = $2 RETURNING *`,
```
trisapeace, line 256: This feels like opening us up to SQL injection. I'd prefer an if-else here or two separate DB functions.

index.js
```diff
@@ -190,15 +191,15 @@ async function sendSingleAlert(locationid, message) {
   })
 }
 
-async function sendAlerts(locationid, displayName) {
+async function sendAlerts(locationid, displayName, component) {
   await sendSingleAlert(
     locationid,
-    `The Brave Sensor at ${displayName} (${locationid}) has disconnected. \nPlease press the reset buttons on either side of the sensor box.\nIf you do not receive a reconnection message shortly after pressing both reset buttons, contact your network administrator.\nYou can also email contact@brave.coop for further support.`,
+    `The Brave Sensor ${component} at ${displayName} (${locationid}) has disconnected. \nPlease press the reset buttons on either side of the sensor box.\nIf you do not receive a reconnection message shortly after pressing both reset buttons, contact your network administrator.\nYou can also email contact@brave.coop for further support.`,
```
trisapeace, line 197: Any thoughts about a different message for door heartbeat alerts?

index.js
```diff
@@ -223,15 +225,26 @@ async function checkHeartbeat() {
...
+        sendAlerts(location.locationid, location.displayName, 'Radar')
+      } else if (radarDelay < RADAR_THRESHOLD_MILLIS && location.radarHeartbeatSentAlerts) {
         helpers.logSentry(`Radar at ${location.locationid} reconnected`)
-        await db.updateSentAlerts(location.locationid, false)
-        sendReconnectionMessage(location.locationid, location.displayName)
+        await db.updateSentRadarAlerts(location.locationid, false, 'radar')
+        sendReconnectionMessage(location.locationid, location.displayName, 'Radar')
+      }
+
+      if (doorDelay > DOOR_THRESHOLD_MILLIS && !location.doorHeartbeatSentAlerts) {
```
trisapeace, line 240: How about an 'else-if' here

.env.example
```diff
@@ -45,4 +45,12 @@ REDIS_CLUSTER_IP_TEST=localhost
...
\ No newline at end of file
+SENTRY_DSN_TEST=localhost
+
+# Threshold for triggering a Radar Heartbeat Alert
+RADAR_THRESHOLD_MILLIS=
+RADAR_THRESHOLD_MILLIS_TEST=1000
+
+# Threshold for triggering a Door Heartbeat Alert
+DOOR_THRESHOLD_MILLIS=
+DOOR_THRESHOLD_MILLIS_TEST=1000
```
trisapeace: Can you please make a note in the next deployment task that we'll need to add these to the secrets?

index.js
```diff
@@ -797,19 +811,17 @@ app.post('/api/door', Validator.body(['coreid', 'data']).exists(), async (reques
...
-        } else if (signal === IM21_DOOR_STATUS.CLOSED) {
+        } else if (signal === IM21_DOOR_STATUS.CLOSED || signal === IM21_DOOR_STATUS.HEARTBEAT_CLOSED) {
           doorSignal = SESSIONSTATE_DOOR.CLOSED
         } else if (signal === IM21_DOOR_STATUS.LOW_BATT) {
           doorSignal = 'LowBatt'
-          helpers.logSentry(`Received a low battery alert for ${locationid}: ${message}`)
-        } else if (signal === IM21_DOOR_STATUS.HEARTBEAT_OPEN || signal === IM21_DOOR_STATUS.HEARTBEAT_CLOSED) {
-          doorSignal = 'HeartBeat'
+          helpers.logSentry(`Received a low battery alert for ${locationid}`)
+          BraveAlerter.sendSingleAlert(locationid, `The battery for the ${location.displayName} door sensor is low, and needs replacing.`)
```
trisapeace, line 821: There is a `sendSingleAlert` function declared locally in `index.js` that takes locationid and a message as parameters. I think you want to use this instead.

The `sendSingleAlert` method in BraveAlerter is more generic and takes the toPhoneNumber, fromPhoneNumber, and message as parameters.

index.js
```diff
@@ -221,15 +222,28 @@ async function checkHeartbeat() {
         const innosentData = await redis.getLatestInnosentSensorData(location.locationid)
         latestRadar = moment(innosentData.timestamp, 'x')
       }
+
+      const doorData = await redis.getLatestDoorSensorData(location.locationid)
+      latestDoor = moment(doorData.timestamp, 'x')
       const currentTime = moment()
```
trisapeace, line 228: What do you think about getting the current time for this comparison from Redis: https://redis.io/commands/TIME ? This will ensure that we're comparing apples to apples in case our server times get out of sync.

index.js
```diff
@@ -221,15 +222,28 @@ async function checkHeartbeat() {
...
       const radarDelay = currentTime.diff(latestRadar)
+      const doorDelay = currentTime.diff(latestDoor)
+
+      const doorHeartbeatExceeded = doorDelay > helpers.getEnvVar('DOOR_THRESHOLD_MILLIS')
+      const radarHeartbeatExceeded = radarDelay > helpers.getEnvVar('RADAR_THRESHOLD_MILLIS')
 
-      if (radarDelay > RADAR_THRESHOLD_MILLIS && !location.heartbeatSentAlerts) {
-        helpers.logSentry(`Radar Heartbeat threshold exceeded; sending alerts for ${location.locationid}`)
+      if ((doorHeartbeatExceeded || radarHeartbeatExceeded) && !location.heartbeatSentAlerts) {
+        helpers.logSentry(`System at ${location.locationid} is disconnected`)
```
trisapeace, line 236: Do you need this log considering you have the more specific logs directly below?

sensor-helm-chart/templates/sensor-deployment.yaml
```diff
@@ -89,6 +89,16 @@ spec:
...
+          - name: RADAR_THRESHOLD_MILIS
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.secretName | quote }}
+                key: RADAR_THRESHOLD_MILIS
+          - name: DOOR_THRESHOLD_MILIS
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.secretName | quote }}
+                key: DOOR_THRESHOLD_MILIS
```
trisapeace: `MILIS` should have two "L"s

setup_postgresql.sh
```diff
@@ -15,4 +15,4 @@ sudo PGPASSWORD=$PG_PASSWORD psql -U $PG_USER -h $PG_HOST -p $PG_PORT -d $PG_DAT
 sudo PGPASSWORD=$PG_PASSWORD psql -U $PG_USER -h $PG_HOST -p $PG_PORT -d $PG_DATABASE --set=sslmode=require -f ./db/014-alterfallbackphonenumbertobeanarray.sql
 sudo PGPASSWORD=$PG_PASSWORD psql -U $PG_USER -h $PG_HOST -p $PG_PORT -d $PG_DATABASE --set=sslmode=require -f ./db/015-add-api-key-to-locations.sql
 sudo PGPASSWORD=$PG_PASSWORD psql -U $PG_USER -h $PG_HOST -p $PG_PORT -d $PG_DATABASE --set=sslmode=require -f ./db/016-alterheatbeackphonenumbertobeanarray.sql
-sudo PGPASSWORD=$PG_PASSWORD psql -U $PG_USER -h $PG_HOST -p $PG_PORT -d $PG_DATABASE --set=sslmode=require -f ./db/017-add-is-active-to-locations.sql
+sudo PGPASSWORD=$PG_PASSWORD psql -U $PG_USER -h $PG_HOST -p $PG_PORT -d $PG_DATABASE --set=sslmode=require -f ./db/017-add-is-active-to-locations.sql
```
trisapeace, line 18: I think you need to rebase and change this to migration 018. 
017 should be the change to add `is_active` to the `locations` table.

index.js
```diff
@@ -213,6 +212,8 @@ async function checkHeartbeat() {
```
trisapeace, line 212: Hmmm, what do you think about the WATCHDOG stuff starting on line 152? Should it run if `heartbeatSentAlerts` is `true`? I don't think that it matters, since that WATCHDOG section doesn't actually send any messages to anyone. Which actually makes me doubt the need for me to run that loop over only active locations. I think it would be OK to run over all of them. What do you think? I'm not entirely clear on the reasons for the WATCHDOG state discrepency/reset seeding thing.

index.js
```diff
@@ -221,15 +222,28 @@ async function checkHeartbeat() {
...
+      if ((doorHeartbeatExceeded || radarHeartbeatExceeded) && !location.heartbeatSentAlerts) {
+        helpers.logSentry(`System at ${location.locationid} is disconnected`)
+        if (doorHeartbeatExceeded) {
+          helpers.logSentry(`Door sensor down at ${location.locationid}`)
+        }
+        if (radarHeartbeatExceeded) {
+          helpers.logSentry(`Radar sensor down at ${location.locationid}`)
+        }
         await db.updateSentAlerts(location.locationid, true)
         sendAlerts(location.locationid, location.displayName)
```
trisapeace, line 244: While you're in here, could you please change the name of this function? How about something like `sendDisconnectionMessage` because it would be more descriptive and match the name of the `sendReconnectionMessage` function.

index.js
trisapeace: Hmmm, on the other hand, the local `sendSingleAlert` function will send the message to the heartbeatRecipients. Is this what we want?

Maybe we want to call `braveAlerter.sendSingleAlert` (note the small "b" to use the already-defined `braveAlerter` object rather than the class), with the responderPhone as the toPhoneNumber.

mariocimet: ah, yeah this was a mistake, I want the local sendSingleAlert to text the heartbeat numbers. I'd prefer this over just the responder phone since Heartbeat numbers includes them but will also notify us of the issue which can contextualize response etc

sensor-helm-chart/templates/sensor-deployment.yaml
mariocimet: yes this is what I botched when deploying this

setup_postgresql.sh
mariocimet: actually 018 is gone, the way we envision it I don't think we require a migration for it, since we are only tracking one composite heartbeat

index.js
mariocimet: I like it, and moment is deprectaed anyways

index.js
mariocimet: I intend to scrap the Watchdog altogether in the state machine refactor so my intention was to leave it untouched for now.

.env.example
mariocimet: yes, done

mariocimet:
@trisapeace made the requested changes and pushed to dev again. I made the thresholds 20 and 5 minutes for door and radar respectively and tested the following:

received a disconnection message after unplugging door sensor and waiting for the threshold to be reached. Used redis-cli to confirm that the state machine wasn't running for it (wasn't doing the xrevrange commands from the state machine loop). Disconnected radar. Reconnected door Argon, waited for it to connect, it and cycled the door and received no message. Reconnected radar as well and shortly after I got a reconnection message and state machine started running as expected

.env.example
```diff
@@ -45,4 +45,8 @@ REDIS_CLUSTER_IP_TEST=localhost
 
 # Data Source URL for Sentry
 SENTRY_DSN=
-SENTRY_DSN_TEST=localhost
\ No newline at end of file
+SENTRY_DSN_TEST=localhost
+
+# Threshold for triggering a Radar Heartbeat Alert
+RADAR_THRESHOLD_MILLIS=1200000
+DOOR_THRESHOLD_MILLIS=1200000
```
trisapeace, line 52: You decided against have `_TEST` versions of these?

db/redis.js
```diff
@@ -22,6 +22,11 @@ function connect() {
   }
 }
 
+async function getCurrentTimeinSeconds() {
+  const time = await client.time()
+  return time[0]
+}
```
trisapeace, line 28: Cool! Glad this worked.

index.js
```diff
@@ -85,6 +84,10 @@ app.use(
   }),
 )
 
+function convertToSeconds(microseconds) {
```
trisapeace, line 87: Do you mean `milliseconds` instead of `microseconds`?

index.js
```diff
@@ -213,23 +216,36 @@ async function checkHeartbeat() {
...
-      if (radarDelay > RADAR_THRESHOLD_MILLIS && !location.heartbeatSentAlerts) {
-        helpers.logSentry(`Radar Heartbeat threshold exceeded; sending alerts for ${location.locationid}`)
+      const doorData = await redis.getLatestDoorSensorData(location.locationid)
+      latestDoor = convertToSeconds(doorData.timestamp)
+      const redisTime = await redis.getCurrentTimeinSeconds()
+      const radarDelay = redisTime - latestRadar
+      const doorDelay = redisTime - latestDoor
+
+      const doorHeartbeatExceeded = doorDelay > helpers.getEnvVar('DOOR_THRESHOLD_SECONDS')
+      const radarHeartbeatExceeded = radarDelay > helpers.getEnvVar('RADAR_THRESHOLD_SECONDS')
```
trisapeace, line 236: I like that you changed this to seconds. I think it makes more sense. However, the names of the environment variables in `.env.example` also need to change to `_SECONDS` then.

trisapeace (approved):
Sounds good to me. Just some minor changes in my comments. Fix them up and I think it's good to go.

index.js
mariocimet: yes good catch

index.js
mariocimet: I do whoops

.env.example
mariocimet: I didn't explicitly. That said, we don't run heartbeat code in test mode anyways.

PR #111, opened 2021-05-29
Add mass notification feature to Dashboard (CU-w9bcb5)

trisapeace:
- Added a "New Notification" screen to the Dashboard. Using it will send a message to all the Responder Phones using the `sendSingleAlert` function in `brave-alert-lib`. 
- By default, it will only send to Responder Phones of Active locations. But if you check the checkbox on the New Notification screen, it will also send to Inactive locations.
- The Subject field is not required. If it is provided and if this will be a text notification, it will be pre-pended to the message with a colon and newline. This is here because when we move to Notifications in the Alert App, the design expects there to be a Subject field (see the [Figma](https://www.figma.com/file/GBEb8H6wEuUGZgvx0Gylzt/BRAVE))
- Moved dashboard-related config and routes to own file in an attempt to start splitting up index.js into smaller, more descriptive files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ the code was deployed.
 
 ### Added
 
-- Storing and checking for door heartbeat messages (CU-wf97dy).
+- Storing and checking for door heartbeat messages (CU-wf97dy)
+- New Notification page on the Dashboard (CU-w9bcb5)
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


trisapeace:
Task linked: [CU-w9bcb5 Add a way to send a custom message to all active responder phones](https://app.clickup.com/t/w9bcb5)

trisapeace:
Closing for now because I think this will be easier after DB schema changes to add an `installations`-like table to Sensors.

PR #113, opened 2021-06-01
Refactor backend state machine and add firmware state machine API endpoint CU-v9ae26

mariocimet:
A host of changes:

- Refactored backend state machine to use firmware state machine topology, and to use a single State Machine file instead of one for each radar type
- Decoupled state machine processing from Session creation, except for generating alerts. The system will now generate subsequent alerts even if a session has not been responded to.
- Added endpoint and handler for alert messages from firmware state machine
- Added heartbeat checking for firmware state machine heartbeats
- Removed several unused database columns
- Removed watchdog timer
- Renamed other columns to conform to new shared Brave Devices schema

Thank you so much @trisapeace for the guidance on testing!!

**Test plan:**
- :heavy_check_mark: Deploy to `dev-sensor-server` and generate alerts from both firmware state machine and legacy devices
- :heavy_check_mark: Hit `POST https://chatbot-dev.brave.coop/api/StartChatbot` for a test location and see that
   - :heavy_check_mark: A session is created and an alert sent to my phone with the correct alert type
   - :heavy_check_mark: additional requests before the SESSION_RESET_THRESHOLD do not generate a new session
   - :heavy_check_mark: additional requests after the SUBSEQUENT_ALERT_MESSAGE_THRESHOLD do generate new sms messages
- :heavy_check_mark: Disconnecting the device and waiting for RADAR_ALERT_THRESHOLD seconds results in an alert about system disconnection
- :heavy_check_mark: Reconnecting the device results in a reconnetion message 
- :heavy_check_mark: Smoke tests display normal behavior for legacy API endpoints


StateMachine.js
```diff
@@ -0,0 +1,87 @@
+const helpers = require('brave-alert-lib')
```
mariocimet, line 1: can make these just functions in index.js or even handleSensorRequest.js()

stateMachine/StateMachine.js
```diff
@@ -0,0 +1,77 @@
...
+          await redis.addStateMachineData(STATE.DURATION_TIMER, location.locationid)
+        }
+        break
+      case STATE.DURATION_TIMER:
+        if (door.signal === DOOR_STATUS.OPEN) {
+          await redis.addStateMachineData(STATE.IDLE, location.locationid)
+        } else if (!movementOverThreshold) {
+          await redis.addStateMachineData(STATE.STILLNESS_TIMER, location.locationid)
+        } else if (
+          await stateMachineHelpers.timerExceeded(location.locationid, +location.durationTimer + +location.initialTimer, STATE.INITIAL_TIMER)
```
mariocimet, line 44: replace with something more explicit like Math.toInt() or something

trisapeace:
Task linked: [CU-v9ae26 add local state machine endpoint to backend](https://app.clickup.com/t/v9ae26)

db/018-refactor-sessions.sql
```diff
@@ -0,0 +1,67 @@
...
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        ALTER TABLE sessions
+            DROP still_counter,
+            DROP od_flag,
```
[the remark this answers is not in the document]
trisapeace: FYI here is a Slack thread where Gordon indicates the possible desire to continue to track old-style sessions: https://bravely.slack.com/archives/CAN15J43F/p1624662155122500?thread_ts=1624658926.121400&cid=CAN15J43F

I'm trying to convince him otherwise :smile:

index.js
```diff
@@ -579,7 +407,7 @@ app.post(
   '/locations',
   [
     Validator.body(['locationid', 'displayName', 'doorCoreID', 'radarCoreID', 'radarType', 'twilioPhone']).notEmpty(),
```
trisapeace, line 409: Add `firmware_state_machine` here.

index.js
```diff
@@ -608,7 +436,7 @@ app.post(
           data.doorCoreID,
           data.radarCoreID,
           data.radarType,
-          data.phone,
+          data.responderPhoneNumber,
           data.twilioPhone,
           data.alertApiKey,
```
trisapeace, line 441: Add `firmware_state_machine` here.

index.js
```diff
@@ -637,20 +465,15 @@ app.post(
...
-      'stillThreshold',
-      'autoResetThreshold',
-      'doorDelay',
+      'movementThreshold',
+      'durationTimer',
+      'stillnessTimer',
+      'initialTimer',
       'reminderTimer',
       'fallbackTimer',
       'isActive',
```
trisapeace, line 474: Add `firmware_state_machine` here.

index.js
```diff
@@ -678,15 +501,10 @@ app.post(
...
-          data.stillThreshold,
-          data.autoResetThreshold,
-          data.doorDelay,
+          data.movementThreshold,
+          data.durationTimer,
+          data.stillnessTimer,
+          data.initialTimer,
           data.reminderTimer,
           data.fallbackTimer,
           newAlertApiKey,
```
trisapeace, line 510: Add `firmware_state_machine` here.

index.js
```diff
@@ -722,7 +540,7 @@ app.post('/api/xethru', Validator.body(['locationid', 'state', 'rpm', 'mov_f', '
 
       const location = await db.getLocationData(locationid)
       if (location.isActive && !location.heartbeatSentAlerts) {
-        await handleSensorRequest(location, RADAR_TYPE.XETHRU)
+        await StateMachine.getNextState(location, handleAlert)
```
trisapeace, line 543: I think this worked out well passing in the `handleAlert` function. Good way to factor out the StateMachine.js file but not need to completely tear apart `index.js` yet.

index.js
```diff
@@ -793,6 +611,40 @@ app.post(
   },
 )
 
+app.post('/api/startChatbot', Validator.body(['coreid', 'event']).exists(), async (request, response) => {
+  try {
+    const validationErrors = Validator.validationResult(request).formatWith(helpers.formatExpressValidationErrors)
+
+    if (validationErrors.isEmpty()) {
+      const coreId = request.body.coreid
+      const alertType = request.body.event.split(' ')[0]
```
trisapeace, line 620: This event-split-thing seems pretty ugly. Why was it decided that this would be how the firmware and webhook would pass the data to this endpoint?

index.js
```diff
@@ -793,6 +611,40 @@ app.post(
...
+      helpers.log(alertType)
```
trisapeace, line 621: Do you still want to keep this log? For it to be useful, it could use some more info about the location, I think.

index.js
```diff
@@ -889,17 +740,52 @@ app.post(
...
+      if (!location) {
+        const errorMessage = `Bad request to ${request.path}: no location matches the coreID ${coreId}`
+        helpers.logError(errorMessage)
+        // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
+        response.status(200).json(errorMessage)
+      } else {
+        const message = JSON.parse(request.body.data)
+        const doorStatus = message.door_status
+        const doorTime = message.door_time
+        const insTime = message.ins_time
```
trisapeace, line 759: Should this be an error if any of these values are empty? If so, maybe they should have a Validator.

index.js
```diff
@@ -889,17 +740,52 @@ app.post(
...
+        response.status(200).json('OK')
+      }
+    } else {
+      const errorMessage = `Bad request to ${request.path}: ${validationErrors.array()}`
+      helpers.logError(errorMessage)
+      // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
+      response.status(200).json(errorMessage)
+    }
+  } catch (err) {
+    const errorMessage = `Error calling ${request.path}: ${JSON.stringify(err)}`
```
trisapeace, line 772: Better not to use `JSON.stringify` for error objects. Just `${err}` actually gives better output.

index.js
```diff
@@ -889,17 +740,52 @@ app.post(
...
   const { recipientNumber, twilioNumber, radarType } = request.body
   try {
     await db.createLocation(
       'SmokeTestLocation',
       recipientNumber,
       17,
-      15,
-      150,
+      15000,
+      150000,
```
trisapeace, line 787: On the UI, these are still in seconds,  but these look like milliseconds. I think that something needs to be updated.

index.js
```diff
@@ -909,10 +795,6 @@ app.post('/smokeTest/setup', async (request, response) => {
       'door_coreID',
       'radar_coreID',
       radarType,
-      2,
-      0,
-      2,
-      8,
       'alertApiKey',
       true,
```
trisapeace, line 799: Add a value for `firmware_state_machine` here.

stateMachine/stateMachineHelpers.js
```diff
@@ -0,0 +1,64 @@
+const { helpers } = require('brave-alert-lib')
```
trisapeace, line 1: I'd love to see some unit tests for the methods in this file.

stateMachine/stateMachineHelpers.js
```diff
@@ -0,0 +1,64 @@
...
+const redis = require('../db/redis')
+const RADAR_TYPE = require('../RadarTypeEnum')
+
+async function movementAverageOverThreshold(radarType, locationid, movementThreshold) {
+  if (radarType === RADAR_TYPE.XETHRU) {
+    try {
+      const xeThruhistory = await redis.getXethruWindow(locationid, '+', '-', 15) // Array of last 15 readings from this location
+      if (xeThruhistory.length === 0) {
+        return false
```
trisapeace, line 10: Maybe add an error log when this happens because it shouldn't, right? Or will it, the first few heartbeats after the sensor is added?

stateMachine/StateMachine.js
```diff
@@ -0,0 +1,85 @@
...
+const stateMachineHelpers = require('./stateMachineHelpers')
+const redis = require('../db/redis')
+const ALERT_REASON = require('../AlertReasonEnum')
+const DOOR_STATUS = require('../SessionStateDoorEnum')
+const STATE = require('./SessionStateEnum')
+
+async function getNextState(location, handleAlert) {
+  try {
+    let state
+    // await redis.beginTransaction()
```
trisapeace, line 11: I think this comment is here to remind you to put this into a redis transaction.

test/integration/serverTest.js
```diff
@@ -1,29 +1,58 @@
...
-const MOV_THRESHOLD = 17
+const braveAlerter = imports.braveAlerter
+const StateMachine = require('../../stateMachine/StateMachine')
+const XETHRU_STATE = require('../../SessionStateXethruEnum')
+const ALERT_REASON = require('../../AlertReasonEnum')
+
+const MOVEMENT_THRESHOLD = 40
+const STILLNESS_TIMER = 300
+const DURATION_TIMER = 600
+const INITIAL_TIMER = 200
```
trisapeace, line 24: Please confirm seconds vs milliseconds for these values.

test/integration/serverTest.js
```diff
@@ -1,29 +1,58 @@
...
+const door_coreID = 'door_particlecoreid1'
+const radar_coreID = 'radar_particlecoreid1'
+
+async function firmwareAlert(coreID, alertReason) {
+  try {
+    await chai
+      .request(server)
+      .post('/api/startChatbot')
+      .send({
+        event: `${alertReason} Alert`,
```
trisapeace, line 45: Ah, I see.

OK, so since we're eventually going to be moving to a new enum for the alert reason when we have the combined Buttons/Sensors DB, what I suggest to avoid the ugly `split` here is  to just create a new enum for `FirmwareAlertType` or something like that which will have the exact two strings that we expect the firmware to send. Then instead of doing a `split` to get the `alertReason` part of it, we would figure out which of the Firmware enum value it is and  based on that , figure out which AlertReason Enum value it should have in our DB. That will give us the most freedom.

But I'm slowing down today in the heat so I'm not sure if I explained that well. Let me know if I didn't make sense!

test/integration/serverTest.js
```diff
@@ -1,29 +1,58 @@
...
+        data: 'test-event',
+        ttl: 60,
+        published_at: '2021-06-14T22:49:16.091Z',
+        coreid: coreID,
+      })
+    await helpers.sleep(10)
```
trisapeace, line 51: Darn, you needed to put `sleep`s back in?

trisapeace (commented):
Alrighty. I still need to finish looking at `serverTest.js` and `testStateMachine`. But this is looking really good.

test/unit/testStateMachine.js
```diff
@@ -0,0 +1,760 @@
...
+        sandbox
+          .stub(redis, 'getStates')
+          .withArgs(testLocationId, redis.getCurrentTimeinMilliseconds() - 150000, '+')
+          .returns([{ state: STATE.DURATION_TIMER }, { state: STATE.STILLNESS_TIMER }, { state: STATE.DURATION_TIMER }])
+        const result = await stateMachineHelpers.timerExceeded(testLocationId, 150000, STATE.INITIAL_TIMER)
+        expect(result).to.be.true
+      })
+    })
+  })
+})
```
trisapeace, line 760: Great set of tests :smile:

trisapeace (commented):
Done looking this over. Great work! What a monster this ended up being. This will make things so much better, thank you for doing it.

db/018-refactor-sessions.sql
```diff
@@ -0,0 +1,67 @@
...
+            DROP duration,
+            ALTER start_time TYPE timestamptz,
+            ALTER start_time SET DEFAULT now(),
+            ALTER start_time SET NOT NULL;
```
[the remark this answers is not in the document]
mariocimet: yes!

db/018-refactor-sessions.sql
```diff
@@ -0,0 +1,67 @@
...
+        ALTER TABLE sessions ADD id UUID NULL;
+        UPDATE sessions SET id = CAST(LPAD(TO_HEX(sessionid), 32, '0') AS UUID);
+        ALTER TABLE sessions 
+            DROP sessionid,
+            ALTER id SET NOT NULL,
+            ALTER id SET DEFAULT gen_random_uuid(),
+            ADD CONSTRAINT sessions_pkey PRIMARY KEY (id);
+
+        ALTER TABLE sessions RENAME start_time TO created_at;
+        ALTER TABLE sessions RENAME incidenttype TO incident_type;
```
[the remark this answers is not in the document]
mariocimet: I would prefer to do a separate task where we do all the renamings to conform to that schema, just because I don't want to add blocks to this one

db/018-refactor-sessions.sql
```diff
@@ -0,0 +1,67 @@
...
+        ALTER TABLE sessions RENAME phonenumber to phone_number;
+        ALTER TABLE sessions drop column end_time;
+        ALTER TABLE sessions ADD COLUMN updated_at timestamptz DEFAULT now() NOT NULL;
```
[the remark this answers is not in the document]
mariocimet: good question, I actually don't have a good reason. Do you have a suggestion?

db/018-refactor-sessions.sql
```diff
@@ -0,0 +1,67 @@
...
+        DROP TRIGGER IF EXISTS set_sessions_timestamp ON sessions;
+
+        CREATE TRIGGER set_sessions_timestamp
+        BEFORE UPDATE ON sessions
+        FOR EACH ROW EXECUTE PROCEDURE trigger_set_timestamp();
+
+        ALTER TABLE locations RENAME mov_threshold TO movement_threshold;
+        ALTER TABLE locations RENAME still_threshold TO stillness_timer;
+        ALTER TABLE locations RENAME duration_threshold TO duration_timer;
+        ALTER TABLE locations RENAME door_stickiness_delay TO initial_timer;
```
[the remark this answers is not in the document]
mariocimet: I prefer a separate task where we do the renamings all at once - just to not drag out this PR further

db/db.js
```diff
@@ -488,7 +313,26 @@ async function getLocationFromParticleCoreID(coreID, clientParam) {
 
 async function getActiveLocations(clientParam) {
```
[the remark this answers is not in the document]
mariocimet: yes, done.

db/db.js
```diff
@@ -488,7 +313,26 @@ async function getLocationFromParticleCoreID(coreID, clientParam) {
...
   try {
-    const results = await runQuery('getLocations', 'SELECT * FROM locations WHERE is_active', [], clientParam)
+    const results = await runQuery('getLocations', 'SELECT * FROM locations WHERE is_active AND firmware_state_machine = false', [], clientParam)
```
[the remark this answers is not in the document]
mariocimet: yes, done

test/integration/serverTest.js
mariocimet: Just to make reasoning about thresholds etc easier, not for debugging flakey tests or anything.

mustache-templates/locationsDashboard.mst
```diff
@@ -27,10 +27,8 @@
                     <thead>
                         <tr class="table-header" scope="row">
                             <th scope="col">Session ID</th>
-                            <th scope="col">Start Time</th>
-                            <th scope="col">End Time</th>
-                            <th scope="col">End State</th>
-                            <th scope="col">Duration</th>
```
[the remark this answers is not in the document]
mariocimet: Duration in this case would just be the difference between session beggining and completion? I don't know if its the most valuable.

index.js
mariocimet: I don't think there's any reason in particular. The firmware could just as easily send just "Duration" or "Stillness" - but that's a change across the other repo as well.

stateMachine/stateMachineHelpers.js
mariocimet: I actually do have some in testStateMachine, towards the bottom.

db/db.js
```diff
@@ -134,7 +132,7 @@ async function getMostRecentSessionPhone(twilioNumber, clientParam) {
   try {
     const results = await runQuery(
       'getMostRecentSessionPhone',
-      "SELECT s.* FROM sessions AS s LEFT JOIN locations AS l ON s.locationid = l.locationid WHERE l.twilio_number = $1  AND s.start_time > (CURRENT_TIMESTAMP - interval '7 days') ORDER BY s.start_time DESC LIMIT 1",
+      'SELECT s.* FROM sessions AS s LEFT JOIN locations AS l ON s.locationid = l.locationid WHERE l.twilio_number = $1 ORDER BY s.created_at DESC LIMIT 1',
```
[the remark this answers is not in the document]
mariocimet: What about doing this in steps?
 add to the migration
 
 DELETE from sessions where od_flag  is 0; (or null? just to be sure)
... do the rest as normal.

index.js
```diff
@@ -129,62 +106,56 @@ async function generateCalculatedTimeDifferenceString(timeToCompare) {
...
+      }
+      braveAlerter.startAlertSession(alertInfo)
+    } else if (currentTime - currentSession.updatedAt >= helpers.getEnvVar('SUBSEQUENT_ALERT_MESSAGE_THRESHOLD')) {
+      helpers.log('handleAlert: sending singleAlert')
+      await db.updateSession(currentSession.id, client)
+      braveAlerter.sendSingleAlert(
+        location.responderPhoneNumber,
+        location.twilioNumber,
+        `An additional ${alertReason} alert was generated at ${location.displayName}`,
+      )
```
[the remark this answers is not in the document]
mariocimet: we like /api/sensorEvent

stateMachine/StateMachine.js
mariocimet: punting on this.

test/integration/serverTest.js
```diff
@@ -119,6 +153,97 @@ describe('Brave Sensor server', () => {
...
+      await sleep(1000)
+      await firmwareAlert(radar_coreID, ALERT_REASON.STILLNESS)
+      const newSessions = await db.getAllSessionsFromLocation(testLocation1Id)
+      const newUpdatedAt = newSessions[0].updatedAt
+      expect(newUpdatedAt).to.be.afterTime(oldUpdatedAt)
+    })
+
+    it('should create additional sessions for alerts that come in after the session reset timeout has expired', async () => {
+      await firmwareAlert(radar_coreID, ALERT_REASON.STILLNESS)
+      await sleep(+helpers.getEnvVar('SESSION_RESET_THRESHOLD') + 50)
```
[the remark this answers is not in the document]
mariocimet: parseInt too instead of +

test/integration/serverTest.js
```diff
@@ -1,29 +1,58 @@
...
 const IM21_DOOR_STATUS = require('../../IM21DoorStatusEnum')
-const { getRandomArbitrary, getRandomInt, printRandomIntArray } = require('../../testingHelpers.js')
+const { getRandomArbitrary, getRandomInt, printRandomIntArray } = require('../../testingHelpers')
+const STATE = require('../../stateMachine/SessionStateEnum')
 
 chai.use(chaiHttp)
 chai.use(sinonChai)
+chai.use(chaiDateTime)
 
+const sandbox = sinon.createSandbox()
```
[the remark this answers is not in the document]
mariocimet: I do!

db/redis.js
```diff
@@ -52,9 +62,40 @@ async function getDoorWindow(locationID, startTime, endTime, windowLength) {
...
+  }
+}
+
+async function getLatestState(locationid) {
+  const singleitem = await getStatesWindow(locationid, '+', '-', 1)
+
+  if (!singleitem) {
+    return null
+  }
+  return singleitem[0]
```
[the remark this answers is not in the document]
mariocimet: getStatesWindow already creates the StateData object, so this is fine as is.

mustache-templates/locationsDashboard.mst
```diff
@@ -39,11 +37,9 @@
...
-                                <td>{{startTime}}</td>
-                                <td>{{endTime}}</td>
-                                <td>{{state}}</td>
-                                <td>{{duration}}</td>
+                                <th scope="row">{{id}}</th>
+                                <td>{{createdAt}}</td>
+                                <td>{{updatedAt}}</td>
                                 <td>{{chatbotState}}</td>
                                 <td>{{incidentType}}</td>
                                 <td>{{notes}}</td>
```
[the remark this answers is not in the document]
mariocimet: I like it but I'm punting it to another task.

db/018-refactor-sessions.sql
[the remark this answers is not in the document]
trisapeace: How about using `end_time`?
There are rows in the DB that have `end_time is null`, but they are all ones where `od_flag = 0`, which will be deleted earlier in this file. So that should give us a better idea of when it was last updated, don't you think?

```
backend-replacement=> select count(*) from sessions where end_time is null and od_flag=1;
 count 
-------
     0
(1 row)
```

mustache-templates/locationsDashboard.mst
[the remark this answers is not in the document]
mariocimet: I like it, but I'm punting to a different task.

index.js
```diff
@@ -40,32 +35,14 @@ const app = express()
...
-// These states will close an ongoing session for a certain location
-const CLOSINGSTATES = [STATE.DOOR_OPENED_CLOSE]
+const braveAlerter = new BraveAlerterConfigurator().createBraveAlerter()
 
-// These states will start a Brave Alert session for a location
-const ALERTSTARTSTATES = [STATE.SUSPECTED_OD]
-
-// Body Parser Middleware
+// // Body Parser Middleware
+// app.use(express.urlencoded({ extended: true }))
```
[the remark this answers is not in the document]
mariocimet: I was just testing replacing it with the non-deprecated version, but then reverted and copy pasted it back in, so it was a duplicate. all cleaned up now. That said, we should replace the deprecated express body parser usage.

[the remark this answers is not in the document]
trisapeace: Yeah, I agree. When you tried to use a non-deprecated version, did it seem difficult?

db/018-refactor-sessions.sql
```diff
@@ -0,0 +1,67 @@
...
+        $t$ LANGUAGE plpgsql;
+
+        DROP TRIGGER IF EXISTS set_sessions_timestamp ON sessions;
+
+        CREATE TRIGGER set_sessions_timestamp
+        BEFORE UPDATE ON sessions
+        FOR EACH ROW EXECUTE PROCEDURE trigger_set_timestamp();
+
+        ALTER TABLE locations RENAME mov_threshold TO movement_threshold;
+        ALTER TABLE locations RENAME still_threshold TO stillness_timer;
```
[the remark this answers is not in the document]
mariocimet: yes! Initial timer is changing from milliseconds to seconds, I divided by 1000 in the db script

index.js
```diff
@@ -129,62 +106,56 @@ async function generateCalculatedTimeDifferenceString(timeToCompare) {
...
+        message: `This is a ${alertReason} alert. Please check on the bathroom at ${location.displayName}. Please respond with 'ok' once you have checked on it.`,
+        reminderTimeoutMillis: location.reminderTimer,
+        fallbackTimeoutMillis: location.fallbackTimer,
+        reminderMessage: `This is a reminder to check on the bathroom`,
+        fallbackMessage: `An alert to check on the bathroom at ${location.displayName} was not responded to. Please check on it`,
+        fallbackToPhoneNumbers: location.fallbackNumbers,
+        fallbackFromPhoneNumber: location.twilioNumber,
+      }
+      braveAlerter.startAlertSession(alertInfo)
+    } else if (currentTime - currentSession.updatedAt >= helpers.getEnvVar('SUBSEQUENT_ALERT_MESSAGE_THRESHOLD')) {
```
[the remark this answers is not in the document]
mariocimet: is the idea to make SUBSEQUENT_ALERT_MESSAGE_THRESHOLD redundant, or to check it inside the else block?

I guess it depends on whether we want to update updated_at with each new alert, regardless of the threshold?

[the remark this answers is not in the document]
trisapeace: Hmmm, no I think you're right that it should only be after the threshold. I don't know what I was thinking about this one. Sorry 'bout that.

test/unit/testStateMachine.js
```diff
@@ -0,0 +1,760 @@
...
+
+      it('should not issue an alert', async () => {
+        const handleAlert = sinon.stub()
+        await StateMachine.getNextState(new Location(testLocationId), handleAlert)
+        expect(handleAlert).to.not.be.called
+      })
+    })
+  })
+
+  describe('and the door is closed, and movement is over the threshold, and initial timer has not elapsed', async () => {
```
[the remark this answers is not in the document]
mariocimet: so this one is actually necessary, but I did duplicate the one below

test/unit/testStateMachine.js
```diff
@@ -0,0 +1,760 @@
...
+    })
+
+    it('should issue a stillness alert', async () => {
+      const handleAlert = sinon.stub()
+      await StateMachine.getNextState(new Location(testLocationId), handleAlert)
+      expect(handleAlert).to.be.calledWith(new Location(testLocationId), ALERT_REASON.STILLNESS)
+    })
+  })
+
+  describe('and the door is closed, and movement is under the threshold and only the stillness timer has elapsed', async () => {
```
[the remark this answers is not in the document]
mariocimet: I don't know if that makes sense as a test - because if the duration timer has elapsed, then the stillness timer would have already elapsed and it would have gone back to idle.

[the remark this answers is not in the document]
mariocimet: nothing says the stillness timer has to be shorter, but I think the same is true of the other case, where the duration timer has gone off, but not stillness - and in that case we also go back to idle

SensorEventEnum.js
```diff
@@ -0,0 +1,6 @@
+const ALERT_REASON = {
```
trisapeace, line 1: Did you mean to call this `SENSOR_EVENT`?

SensorEventEnum.js
```diff
@@ -0,0 +1,6 @@
...
+  DURATION: 'Duration Alert',
+  STILLNESS: 'Stillness Alert',
+}
+
+module.exports = Object.freeze(ALERT_REASON)
```
trisapeace, line 6: Did you mean to call this `SENSOR_EVENT`?

mustache-templates/updateLocation.mst
```diff
@@ -110,19 +110,19 @@
...
                     <div class="col-sm-5">
-                        <input type="text" class="form-control" name="reminderTimer" placeholder="Reminder Timer (milliseconds)" value="{{reminderTimer}}" required>
+                        <input type="text" class="form-control" name="reminderTimer" placeholder="Reminder Timer (seconds)" value="{{reminderTimer}}" required>
                     </div>
                 </div>
                 <div class="form-group row justify-content-start row-no-gutters">
                     <label for="fallbackTimer" class="col-sm-3 col-form-label">Fallback Message Timer (ms):</label>
                     <div class="col-sm-5">
-                        <input type="text" class="form-control" name="fallbackTimer" placeholder="Fallback Message Timer (milliseconds)" value="{{fallbackTimer}}" required>
+                        <input type="text" class="form-control" name="fallbackTimer" placeholder="Fallback Message Timer (seconds)" value="{{fallbackTimer}}" required>
```
trisapeace, line 125: I think these two are still in ms. Unless you want to change them now too?

test/integration/serverTest.js
```diff
@@ -136,7 +135,7 @@ async function im21Door(doorCoreId, signal) {
         coreid: doorCoreId,
         data: `{ "data": "${signal}", "control": "86"}`,
       })
-    await helpers.sleep(10)
+    await helpers.sleep(50)
```
trisapeace, line 138: Hmmm, how long do the tests take now? With all these extended `sleep` calls, it looks like this file will take like 5 minutes more to run. That feels long, doesn't it?

trisapeace (commented):
Looks good! Just a few minor things...

test/integration/serverTest.js
mariocimet: They take just under 30 seconds to run on my machine

mustache-templates/updateLocation.mst
mariocimet: ah, yeah you're right. Switching them back, I think changing them is for another task.

test/integration/serverTest.js
mariocimet: 2 minutes, 28 secs on travis

mariocimet: the reason I chose this for tests is because my implementation uses "parseInt" (also our eventual db schema calls for integers in these) so the smallest threshold I can have is 1 second, for the initial timer.

trisapeace: Sweet. Thanks for checking for me.

mariocimet:
> Thanks for all your hard work on this. I think it's much improved.

thank *you* for all *your* hard work on this

PR #114, opened 2021-06-14
Add VSCode launch config for debugging Mocha tests

trisapeace:
- So that those of us who use VS Code will be able to use it to
  debug Mocha tests without any extra work

@mariocimet , after our code review / pair programming session last week, I started thinking that we _shoud_ be able to have a debugger working in VS code. And it didn't actually take much effort to make this work. Wish I'd done it much earlier. 

This adds two run configurations. One which runs all the tests in the `/test` folder and the other which runs all the tests in the currently-open file.

All you need to do is open up the "Run and Debug" View, by clicking on the 
image
icon in the left menu, then you can select which of those Run configurations you want to use in this dropdown menu:
image
and push the "Play" button.

Then, it will run the tests and stop at any breakpoints that you've set in the code and then you can explore the current value of all the in-scope variables, you can create Watch expressions, and explore the call stack:
image

Honestly, why didn't I do this earlier??????

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,31 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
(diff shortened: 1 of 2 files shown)


mariocimet:
neat!!

mariocimet:
I too have abused console logs

mariocimet (approved):

PR #115, opened 2021-06-17
Use `coreid` instead of `location` to determine Location for XeThrus (CU-mt15y9)

trisapeace:
- This is a breaking change because sending `location` will no longer
  work
- `location` was being set during device provisioning. This means that
  it is one more thing to remember and to get right as part of device
  provisioning. Better to have it flexible, changeable, and also
  more consistent with the INS and IM21 webhooks
- Note that it doesn't use `data` like the INS and IM21 webhooks
  because it was set up with Custom form fields in the Particle
  Console whereas the INS and IM21 webhooks use the Default form
  fields which put all the custom parts as a JSON string under `data`
- Updated tests accordingly
- Also updated smoke test setup to use seconds instead of milliseconds where needed

Test Plan:
- :heavy_check_mark: Deploy to `dev.sensors.brave.coop`
- :heavy_check_mark: Given the correct radar core ID is in the DB for the XeThru in my bathroom (and a shortened `still_threshold`), trigger a Stillness alert
- :heavy_check_mark: Given the correct radar core ID is in the DB for the XeThru in my bathroom (and a shortened `still_threshold`), trigger a Duration alert
- :heavy_check_mark: Given an incorrect radar core ID is in the DB for the XeThru in my bathroom (and a shortened `RADAR_THRESHOLD_SECONDS`), see that 
   - :heavy_check_mark: Can see error logs about a radar that is posting without a matching Location
   - :heavy_check_mark: Get a disconnection message because my Location isn't receiving radar data anymore


trisapeace:
Task linked: [CU-mt15y9 make radar endpoints identify radar locationid based on core_id](https://app.clickup.com/t/mt15y9)

index.js
```diff
@@ -534,29 +534,34 @@ app.post(
...
-      res.status(200).json('OK')
     } else {
       const errorMessage = `Bad request to ${req.path}: ${validationErrors.array()}`
       helpers.logError(errorMessage)
       // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
       res.status(200).json(errorMessage)
     }
   } catch (err) {
-    const errorMessage = `Error calling ${req.path}: ${JSON.stringify(err)}`
+    const errorMessage = `Error calling ${req.path}: ${err.toString()}`
```
mariocimet, line 564: woops! This is why we get those Error calling /api/xethru {} messages isn't it.

trisapeace: Yup, exactly.

index.js
```diff
@@ -534,29 +534,34 @@ app.post(
 app.use(braveAlerter.getRouter())
 
 // Handler for incoming XeThru POST requests
-app.post('/api/xethru', Validator.body(['locationid', 'state', 'rpm', 'mov_f', 'mov_s']).exists(), async (req, res) => {
+app.post('/api/xethru', Validator.body(['coreid', 'state', 'rpm', 'mov_f', 'mov_s']).exists(), async (req, res) => {
```
mariocimet, line 537: Just wondering if this is actually breaking? On the one hand if you tried to set up a new radar with old instructions and didn't set the right coreID in the creation page, it wouldn't work, at the same time, it's still backwards compatible with the old firmware

mariocimet (approved):
looks good!

mariocimet:
Before deploying let's make sure to double check what's in the radar coreid column for all our production locations (adding this to the clickup task as well).

index.js
mariocimet: ah nvm, I just noticed in the clickup task that there's a webhook component to the change as well

trisapeace: Yes, I've already update the production webhook to add the `coreid` (so that we don't get in trouble by forgetting to do it). Then we can take our time remembering to remove `locationid` from it. It looks like there were a couple of other things (like `device` in there that we don't use that we could clean up if we really wanted to.

trisapeace:
Thanks! Good idea. I did check the radar IDs in the production database when I first started working on this code, but I agree that it would be a very good idea to check again right before we go into production.

PR #116, opened 2021-07-14
Fixed bug where stringifying errors removes all details about the error, querying radar redis streams with time instead of # of entries

mariocimet:


index.js
```diff
@@ -216,7 +216,7 @@ async function checkHeartbeat() {
         sendReconnectionMessage(location.locationid, location.displayName)
       }
     } catch (err) {
-      helpers.logError(`Error checking heartbeat: ${err}`)
```
trisapeace, line 219: These ones that are just printing `${err}` should be find. I think it's only the `${JOSN.stringify(err)}` ones that need to change.

trisapeace (approved):

db/redis.js
```diff
@@ -162,6 +163,34 @@ async function getInnosentWindow(locationID, startTime, endTime, windowLength) {
   return radarStream
 }
 
+async function getInnosentTimeWindow(locationID, windowSize) {
+  const windowSizeinMilliseconds = windowSize * 1000
+  const currentTime = await getCurrentTimeinMilliseconds()
+  let startTime = currentTime - windowSizeinMilliseconds
+
+  if (isTestEnvironment) {
```
mariocimet, line 171: so both of my new functions have this. I only added this because of weirdness in Travis with testing. Without this, I get the error:

`	Error computing Innosent Moving Average: ReplyError: ERR Invalid stream ID specified as stream command argument
`
for each time this function is called in the tests. Weirdly this did not occur when just running tests locally with npm test (before I added the flag).

.travis.yml
```diff
@@ -42,6 +42,7 @@ env:
     - REDIS_CLUSTER_IP_TEST=localhost
     - SESSION_RESET_THRESHOLD_TEST=2000
     - SUBSEQUENT_ALERT_MESSAGE_THRESHOLD_TEST=1000
+    - RADAR_WINDOW_SIZE_SECONDS=2
```
trisapeace, line 45: This might be your bug. You need `_TEST` at the end of this. Otherwise, when running the tests on Travis, you'll end up with `undefined` when it calls `helpers.getEnvVar('RADAR_WINDOW_SIZE_SECONDS')`.

mariocimet:
doh! @trisapeace that's definitely it

.env.example
```diff
@@ -58,4 +58,8 @@ SESSION_RESET_THRESHOLD_TEST=2000
 
 # Threshold for sending out a new message for a new alert during a session.
 SUBSEQUENT_ALERT_MESSAGE_THRESHOLD=120
-SUBSEQUENT_ALERT_MESSAGE_THRESHOLD_TEST=1000
\ No newline at end of file
+SUBSEQUENT_ALERT_MESSAGE_THRESHOLD_TEST=1000
+
+# How large a window we use to calculate moving average of radar values
+RADAR_WINDOW_SIZE_SECONDS=15
```
trisapeace, line 64: Could you please pre-add these values to the staging and prod secrets or add a subtask to do it in the deploy task?

stateMachine/stateMachineHelpers.js
```diff
@@ -3,9 +3,10 @@ const redis = require('../db/redis')
 const RADAR_TYPE = require('../RadarTypeEnum')
 
 async function movementAverageOverThreshold(radarType, locationid, movementThreshold) {
+  const windowSize = helpers.getEnvVar('RADAR_WINDOW_SIZE_SECONDS')
   if (radarType === RADAR_TYPE.XETHRU) {
     try {
-      const xethruHistory = await redis.getXethruWindow(locationid, '+', '-', 15) // Array of last 15 readings from this location
+      const xethruHistory = await redis.getXethruTimeWindow(locationid, windowSize) // Array of last 15 readings from this location
```
trisapeace, line 9: This comment is no longer accurate.

trisapeace (approved):
Cool. Just a coupla things. Approved anyways :smile:

PR #118, opened 2021-07-19
Add GET /alert/historicAlerts endpoint and `sessions.responded_at` column (CU-hjwfx2)

trisapeace:
- Use alert_type_enum in DB and from Alert Lib instead of strings:
   - To better conform with our [desired shared DB schema](https://docs.google.com/document/d/1k-xkWnR7xoWRi6gEiN8p-nda5bhxDvOqNYCtPNi7OwY/edit)
   - So that we don't need to translate between the AlertReasonEnum.js values here and the alertTypeEnum.js values in BraveAlertLib
- Add `responded_at` to Sessions
   - So that we can display it on the Alert App
   - Also changed the `db.js` functions around saving sessions to be more like the way that Buttons does it for more consistency between the two code bases and so that we only need one DB function for saving sessions instead of two.
- Add `GET /alert/historicAlerts` endpoint
   - For use in the Alert App
- Add corresponding tests

## Test Plan
- :heavy_check_mark: Deploy to dev and set the same API key for both `Theresa` (XeThru + server state machine) and `Theresa3` (Boron + firmware state machine) locations
- :heavy_check_mark: Create a real Duration Alert in my bathroom
   - :heavy_check_mark: See Alert Reason "Duration" in the dashboard
   - :heavy_check_mark: See Responded At value in the dashboard
- :heavy_check_mark: Create a real Stillness Alert in my bathroom
   - :heavy_check_mark: See Alert Reason "Stillness" in the dashboard
   - :heavy_check_mark: See Responded At value in the dashboard
- :heavy_check_mark; Hit `GET https://dev.sensors.brave.coop/alert/historicAlerts` for my API key and see that
   - :heavy_check_mark: Not-completed alerts in the last `SESSION_RESET_THRESHOLD` seconds are not returned
   - :heavy_check_mark: Completed alerts in the last `SESSION_RESET_THRESHOLD` seconds are returned
   - :heavy_check_mark: All alerts longer ago than `SESSION_RESET_THRESHOLD` are  returned
   - :heavy_check_mark: Different values for `X-API-KEY` (missing, empty string, no matching API key, matching API key)
   - :heavy_check_mark: Different values for `maxHistoricAlerts` (missing, null, empty string, 0, -1, 1, 99999)
   - :heavy_check_mark: HistoricAlerts which are returned have the values I'd expect based on the Sessions they were created from
-  :heavy_check_mark: Hit the `GET https://chatbot-dev.brave.coop/alert/historicAlerts` endpoint from the Alert App and render its results
- :heavy_check_mark: Verify that the `sessions` table is updated correctly
   - :heavy_check_mark: State transitions
   - :heavy_check_mark: Additional alerts

### DIFF
--- a/BraveAlerterConfigurator.js
+++ b/BraveAlerterConfigurator.js
@@ -1,6 +1,6 @@
 /* eslint-disable class-methods-use-this */
 // In-house dependencies
-const { BraveAlerter, AlertSession, ALERT_STATE, helpers, Location, SYSTEM } = require('brave-alert-lib')
+const { BraveAlerter, AlertSession, CHATBOT_STATE, helpers, Location, SYSTEM } = require('brave-alert-lib')
 const db = require('./db/db')
 
 const incidentTypes = ['No One Inside', 'Person responded', 'Overdose', 'None of the above']
@@ -14,6 +14,8 @@ class BraveAlerterConfigurator {
       this.getAlertSessionByPhoneNumber.bind(this),
       this.alertSessionChangedCallback.bind(this),
       this.getLocationByAlertApiKey.bind(this),
+      this.getHistoricAlertsByAlertApiKey.bind(this),
+      () => {},
       false,
       this.getReturnMessage.bind(this),
     )
@@ -96,26 +98,30 @@ class BraveAlerterConfigurator {
     return new Location(locations[0].locationid, SYSTEM.SENSOR)
   }
 
+  async getHistoricAlertsByAlertApiKey() {
+    return null
+  }
+
   getReturnMessage(fromAlertState, toAlertState) {
     let returnMessage
 
     switch (fromAlertState) {
-      case ALERT_STATE.STARTED:
-      case ALERT_STATE.WAITING_FOR_REPLY:
+      case CHATBOT_STATE.STARTED:
+      case CHATBOT_STATE.WAITING_FOR_REPLY:
         returnMessage =
           'Please respond with the number corresponding to the incident. \n1: No One Inside\n2: Person Responded\n3: Overdose\n4: None of the Above'
         break
 
-      case ALERT_STATE.WAITING_FOR_CATEGORY:
-        if (toAlertState === ALERT_STATE.WAITING_FOR_CATEGORY) {
+      case CHATBOT_STATE.WAITING_FOR_CATEGORY:
+        if (toAlertState === CHATBOT_STATE.WAITING_FOR_CATEGORY) {
           returnMessage =
             'Invalid category, please try again\n\nPlease respond with the number corresponding to the incident. \n1: No One Inside\n2: Person Responded\n3: Overdose\n4: None of the Above'
-        } else if (toAlertState === ALERT_STATE.COMPLETED) {
+        } else if (toAlertState === CHATBOT_STATE.COMPLETED) {
           returnMessage = 'Thank you!'
         }
         break
 
-      case ALERT_STATE.COMPLETED:
+      case CHATBOT_STATE.COMPLETED:
         returnMessage = 'Thank you'
         break
 
--- a/BraveAlerterConfigurator.js
+++ b/BraveAlerterConfigurator.js
@@ -68,8 +68,19 @@ class BraveAlerterConfigurator {
       const session = await db.getSessionWithSessionId(alertSession.sessionId, client)
 
       if (session) {
-        const incidentType = incidentTypes[incidentTypeKeys.indexOf(alertSession.incidentCategoryKey)]
-        await db.saveAlertSession(alertSession.alertState, incidentType, alertSession.sessionId, client)
+        if (alertSession.alertState) {
+          session.chatbotState = alertSession.alertState
+        }
+
+        if (alertSession.incidentCategoryKey) {
+          session.incidentType = incidentTypes[incidentTypeKeys.indexOf(alertSession.incidentCategoryKey)]
+        }
+
+        if (alertSession.alertState === CHATBOT_STATE.WAITING_FOR_CATEGORY) {
+          session.respondedAt = await db.getCurrentTime(client)
+        }
+
+        await db.saveSession(session, client)
       } else {
         helpers.logError(`alertSessionChangedCallback was called for a non-existent session: ${alertSession.sessionId}`)
       }
--- a/BraveAlerterConfigurator.js
+++ b/BraveAlerterConfigurator.js
@@ -1,6 +1,6 @@
 /* eslint-disable class-methods-use-this */
 // In-house dependencies
-const { BraveAlerter, AlertSession, CHATBOT_STATE, helpers, Location, SYSTEM } = require('brave-alert-lib')
+const { BraveAlerter, AlertSession, CHATBOT_STATE, helpers, HistoricAlert, Location, SYSTEM } = require('brave-alert-lib')
 const db = require('./db/db')
 
 const incidentTypes = ['No One Inside', 'Person responded', 'Overdose', 'None of the above']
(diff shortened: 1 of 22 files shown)


trisapeace:
Task linked: [CU-hjwfx2 Frontend: Alert History List View: Implementation](https://app.clickup.com/t/hjwfx2)

db/db.js
```diff
@@ -238,27 +238,71 @@ async function updateSentAlerts(locationid, sentalerts, clientParam) {
...
 }
 
-// Saves the state and incident type into the sessions table
-async function updateSession(id, clientParam) {
+async function getHistoricAlertsByAlertApiKey(alertApiKey, maxHistoricAlerts, maxTimeAgoInMillis, clientParam) {
   try {
-    await runQuery('updateSession', 'UPDATE sessions SET updated_at = now() WHERE id = $1', [id], clientParam)
+    // Historic Alerts are those with status "Completed" or that were last updated longer ago than the SESSION_RESET_TIMEOUT
+    const results = await runQuery(
+      'getHistoricAlertsByAlertApiKey',
```
mariocimet, line 283: nice that Sensors inherits the improvements from getting this into Buttons :)

db/019-change-alert-reason-values.sql
```diff
@@ -0,0 +1,44 @@
...
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        -- ADD SCRIPT HERE
+        -- Note: To view enum types and their values in `psql`, use the command `\dT+`
+        CREATE TYPE alert_type_enum AS ENUM ('BUTTONS_NOT_URGENT', 'BUTTONS_URGENT', 'SENSOR_NO_MOTION', 'SENSOR_DURATION', 'SENSOR_UNKNOWN');
```
mariocimet, line 16: So I actually don't remember when we were coming up with the enum entries - do you remember why we opted for NO_MOTION over the extant STILLNESS?

README.md
```diff
@@ -153,7 +153,7 @@ You will need to populate the .env file with connection parameters for both redi
 To do this, cd into the BraveSensor-Server directory and run the following command
 
 ```
-sudo PG_PORT=<your db's port> PG_HOST=<your db's host> PG_PASSWORD=<your db's password> PG_USER=<your db's user> PG_DATABASE=<your db name> setup_postgresql_local.sh
+sudo PG_PORT=<your db's port> PG_HOST=<your db's host> PG_PASSWORD=<your db's password> PG_USER=<your db's user> PG_DATABASE=<your db name> ./setup_postgresql_local.sh
```
mariocimet, line 156: ha! finally

test/integration/dbTest/getHistoricAlertsByAlertApiKeyTest.js
```diff
@@ -0,0 +1,386 @@
...
+        'radarCoreId',
+        'radarType',
+        this.alertApiKey,
+        true,
+        false,
+      )
+
+      this.session1 = await db.createSession(locationid, 'phonenumber1', ALERT_TYPE.SENSOR_DURATION)
+      this.session2 = await db.createSession(locationid, 'phonenumber2', ALERT_TYPE.SENSOR_DURATION)
+      await helpers.sleep(2000) // Couldn't get around doing this because there is a trigger on updated_at to keep it correct
```
mariocimet, line 272: does the 2000 refer to SESSION_RESET_THRESHOLD_TEST in travis.yml? if so chould we use an env var instead to make this test less likely to flake?

index.js
```diff
@@ -630,9 +644,9 @@ app.post('/api/sensorEvent', Validator.body(['coreid', 'event']).exists(), async
       const coreId = request.body.coreid
       const sensorEvent = request.body.event
       if (sensorEvent === SENSOR_EVENT.DURATION) {
-        alertType = ALERT_REASON.DURATION
+        alertType = ALERT_TYPE.SENSOR_DURATION
       } else if (sensorEvent === SENSOR_EVENT.STILLNESS) {
```
mariocimet, line 648: Also on a schema note - should we update the firmware to send NO_MOTION (and consequently the Sensor Event enum? will it then become redundant?

db/db.js
```diff
@@ -238,27 +238,71 @@ async function updateSentAlerts(locationid, sentalerts, clientParam) {
...
+      [alertApiKey, ALERT_STATE.COMPLETED, maxTimeAgoInMillis, maxHistoricAlerts],
+      clientParam,
+    )
+
+    return results.rows
   } catch (err) {
     helpers.log(JSON.stringify(err))
   }
+
+  return null
```
mariocimet, line 305: So I noticed we have an expanding variety of behaviors when pg returns something other than the array of query results we expect. Throughout db.js we're doing:

`    if (results === undefined || results.rows.length === 0) {
      return null
    }
`

`    if (results === undefined) {
      helpers.log('Error: No location with associated API key exists')
      return null
    }
`

`    if (!results) {
      return null
    }
`

and now this - where null is default if the try block errors out (but not if it returns an empty array!). Seems worth it to decide on one? I think this one makes sense, and then the callers just need to be smart about handling empty arrays (which in most case works fine?)

mariocimet (approved):
approved with some notes about details

db/019-change-alert-reason-values.sql
trisapeace: I think just because it's what was already being used in the Alert Lib's enum. I can change it though, it would make more sense based on the terminology that we're already using.

index.js
trisapeace: We could do. But it would need to be the whole string `SENSOR_DURATION` and `SENSOR_NO_MOTION` (or `SENSOR_STILLNESS` after I change it in `brave-alert-lib`).

package.json
```diff
@@ -19,7 +19,7 @@
     "@sentry/tracing": "^6.2.5",
     "axios": "^0.21.1",
     "body-parser": "^1.19.0",
-    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v3.3.0",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#td-update-enums",
```
trisapeace, line 22: Note to self, this needs to be updated once `brave-alert-lib` has a new version.

db/019-change-alert-reason-values.sql
trisapeace: Will be fixed in this new version of `brave-alert-lib`: https://github.com/bravetechnologycoop/brave-alert-lib/pull/24

test/integration/dbTest/getHistoricAlertsByAlertApiKeyTest.js
trisapeace: No, it was just to make sure that session1 and session2 happen after the `1000` ms passed into `const rows = await db.getHistoricAlertsByAlertApiKey(this.alertApiKey, 5, 1000)` in the test. But I should actually be able to reduce it to 1000 to match that value and that will work too. I can use a variable to make that association more obvious. I haven't seen any flakiness with the 1000ms, but if our DB connection is slow for some reason, then I could see us needing to increase it eventually.

db/db.js
trisapeace: Just checked. In this case the caller is fine with an empty array. So we only want it to return null on errors.

BraveAlerterConfigurator.js
```diff
@@ -14,6 +14,8 @@ class BraveAlerterConfigurator {
       this.getAlertSessionByPhoneNumber.bind(this),
       this.alertSessionChangedCallback.bind(this),
       this.getLocationByAlertApiKey.bind(this),
+      this.getHistoricAlertsByAlertApiKey.bind(this),
+      () => {},
```
trisapeace, line 18: This is because the `brave-alert-lib` upgrade includes Dave's code for the notification count endpoint. This empty function will eventually be filled in by him for that.

mariocimet (approved):
Looks good! reapproved

package.json
trisapeace: Done.

PR #122, opened 2021-08-06
Milestone 1 Complete - CU-18r17vf

phelanm20:
Added call to cloud function for Brave Siren

- Added post request to particle API
- Added check for whether or not location has siren


mariocimet:
on the right track!

index.js
```diff
@@ -148,8 +149,22 @@ async function handleAlert(location, alertType) {
         fallbackMessage: `An alert to check on the bathroom at ${location.displayName} was not responded to. Please check on it`,
         fallbackToPhoneNumbers: location.fallbackNumbers,
         fallbackFromPhoneNumber: location.twilioNumber,
+        sirenParticleId: location.siren_particle_id,
```
trisapeace, line 152: The `alertInfo` object is only used as a parameter into `braveAlerter.startAlertSession`. So it doesn't need to have `sirenParticleId` in here and it can be moved into the same `else if` block as `braveAlerter.startAlertSession(alertInfo)`.

index.js
```diff
@@ -148,8 +149,22 @@ async function handleAlert(location, alertType) {
...
       }
-      braveAlerter.startAlertSession(alertInfo)
+
+      if (location.siren_particle_id != null) {
```
trisapeace, line 155: So Javascript is weird in that it has two different ways to check for equality. One which is the regular double-equals and the other which is the triple-equals. Generally speaking, what you normally want, is the triple-equals type, which means that checks like this would should actually be `!== null` and ones like on line 165 should be `=== null`, as funny as that looks.
Here's a good explanation, if you're interested in learning more: https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a

index.js
```diff
@@ -148,8 +149,22 @@ async function handleAlert(location, alertType) {
...
+        try{
+          await axios.post('https://api.particle.io/v1/devices/location.siren_particle_id/start-siren', {
```
trisapeace, line 157: Are you trying to get the location's `siren_particle_id` to be inserted in the middle of this URL? If so, JS has a cool syntax for that using backticks and `${}`. In this case, you probably want: 
```
`https://api.particle.io/v1/devices/${location.siren_particle_id}/start-siren`
```

More explanation if you'd like: https://www.w3schools.com/js/js_string_templates.asp

index.js
```diff
@@ -148,8 +149,22 @@ async function handleAlert(location, alertType) {
...
+            arg: 'start',
+            access_token: process.env.ACCESS_TOKEN
```
trisapeace, line 159: I suggest calling the token `PARTICLE_ACCESS_TOKEN` just because it's possible that we end up with a few third-party tools that also call their things access tokens, lol.

What you've done here is the correct way to get at an environment variable; however, we've come up with our own helper function to do it for us because it gives a different value when we run the tests than when we are in a real dev, staging, or production environment. So instead of `process.env.ENV_VAR_NAME` we use `helpers.getEnvVar('ENV_VAR_NAME')`.

trisapeace (commented):
I totally agree with Mario. This is a good start. You're on your way.

Here are a few housekeeping things that are just related to the way that we do things, but are in no way things that you would have been expected to know:
- Can you please add a line in the `CHANGELOG.md` file in the `[unreleased]` section that just gives a short description of the change and the ClickUp reference number? This helps us keep track of what changes in each deployment.
- In the name of your GitHub pull request, can you please put the ClickUp reference number (it'll look something like `CU-abiy82` where the stuff after `CU-` is in the URL of the ClickUp Task. That will help us keep the code the the Task associated because it will automatically make a comment on your PR with a link to the Task and will add a link to the PR from ClickUp, which can be helpful later on.
- When we add a new environment variable, we also add a dummy version of it to `.env.example`. The reason for this is because we never commit `.env` to GitHub because we have a public repository and we don't want just anyone on the Internet to know our passwords, API keys, access tokens, etc. But we do want the other devs on the team to know what environment variables are needed and where to find them. So the `.env.example` helps with that. The `_TEST` version of each one is used instead of the real one when we run the automated tests.
- One of the things that is automatically checked by GitHub when you create a Pull Request is the coding style. We use ESLint and Prettier to enforce certain rules like indentation and variable name style. To run this locally before you commit so that you can see any "problems" that it finds, you can do `npm run lint`

phelanm20:
> I totally agree with Mario. This is a good start. You're on your way.
> 
> Here are a few housekeeping things that are just related to the way that we do things, but are in no way things that you would have been expected to know:
> 
> * Can you please add a line in the `CHANGELOG.md` file in the `[unreleased]` section that just gives a short description of the change and the ClickUp reference number? This helps us keep track of what changes in each deployment.
> * In the name of your GitHub pull request, can you please put the ClickUp reference number (it'll look something like `CU-abiy82` where the stuff after `CU-` is in the URL of the ClickUp Task. That will help us keep the code the the Task associated because it will automatically make a comment on your PR with a link to the Task and will add a link to the PR from ClickUp, which can be helpful later on.
> * When we add a new environment variable, we also add a dummy version of it to `.env.example`. The reason for this is because we never commit `.env` to GitHub because we have a public repository and we don't want just anyone on the Internet to know our passwords, API keys, access tokens, etc. But we do want the other devs on the team to know what environment variables are needed and where to find them. So the `.env.example` helps with that. The `_TEST` version of each one is used instead of the real one when we run the automated tests.
> * One of the things that is automatically checked by GitHub when you create a Pull Request is the coding style. We use ESLint and Prettier to enforce certain rules like indentation and variable name style. To run this locally before you commit so that you can see any "problems" that it finds, you can do `npm run lint`

phelanm20:
When I run `npm run lint` I get 5472 errors all looking like variations of this:   `'23:50   error  Delete `␍`  prettier/prettier'`
I've installed both prettier and ESlint - not sure what is happening.

.env.example
```diff
@@ -62,4 +62,7 @@ SUBSEQUENT_ALERT_MESSAGE_THRESHOLD_TEST=1000
 
 # How large a window we use to calculate moving average of radar values
 RADAR_WINDOW_SIZE_SECONDS=15
-RADAR_WINDOW_SIZE_SECONDS_TEST=1
\ No newline at end of file
+RADAR_WINDOW_SIZE_SECONDS_TEST=1
+
+# Access Token for particle device
+PARTICLE_ACCESS_TOKEN=particle_access_token
```
trisapeace, line 68: Could you please add a `PARTICLE_ACCESS_TOKEN_TEST` dummy value under this as well? That will remind people to fill something in if they want to use it in their tests.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ the code was deployed.
 - Alert Type to the Dashboard.
 - Tracking of when sessions are first responded to (CU-hjwfx2).
 - `GET /alert/historicAlerts` endpoint (CU-hjwfx2).
+- Added cloud function call to siren in case of alert (CU-18r17vf)
```
trisapeace, line 19: Could you please add a period of the end of this to be consistent with the other lines?

db/021-add-siren-particle-id.sql
```diff
@@ -0,0 +1,20 @@
+DO $migration$
+    DECLARE migrationId INT;
+    DECLARE lastSuccessfulMigrationId INT;
+BEGIN
+    -- The migration ID of this file
+    migrationId := 20;
```
trisapeace, line 6: This ID should match the number in the file name.

db/021-add-siren-particle-id.sql
```diff
@@ -0,0 +1,20 @@
...
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        ALTER TABLE locations ADD COLUMN siren-particle-id TEXT;
```
trisapeace, line 14: Oh good. I'm glad this worked for you this time!

Can you please use underscores instead of hyphens in the column name. It will be more consistent with our other column names and also sometimes hyphens are interpreted as minus signs and we might get some weird attempts at subtraction when using hyphens.

index.js
```diff
@@ -137,19 +138,32 @@ async function handleAlert(location, alertType) {
...
-        reminderTimeoutMillis: location.reminderTimer,
-        fallbackTimeoutMillis: location.fallbackTimer,
-        reminderMessage: `This is a reminder to check on the bathroom`,
-        fallbackMessage: `An alert to check on the bathroom at ${location.displayName} was not responded to. Please check on it`,
-        fallbackToPhoneNumbers: location.fallbackNumbers,
-        fallbackFromPhoneNumber: location.twilioNumber,
+      
+      if (location.siren_particle_id !== null) {
+        try{
+          await axios.post('https://api.particle.io/v1/devices/${location.siren_particle_id}/start-siren', {
```
trisapeace, line 144: In order for the `${}` part to actually fill in the value, you'll need to use backticks instead of single quotes around this string. It's usually the same key as the one with the `~` on it.

index.js
```diff
@@ -137,19 +138,32 @@ async function handleAlert(location, alertType) {
...
+            arg: 'start',
+            access_token: helpers.getEnvVar('PARTICLE_ACCESS_TOKEN')
+          })
```
trisapeace, line 147: Hmmm, I know that I was the one to suggest using axios, but now I'm thinking that you'd be better off using their JS library: https://docs.particle.io/reference/SDKs/javascript/#callfunction 
It looks quite nice and then you don't need to worry about URL manipulation.

index.js
```diff
@@ -137,19 +138,32 @@ async function handleAlert(location, alertType) {
...
+        } catch (e) {
+          helpers.log(e)
+        }
+      } else if (location.siren_particle_id === null) {
```
trisapeace, line 151: You can just say `} else {` here. No need to recheck the value of `location.siren_particle_id`.

index.js
```diff
@@ -137,19 +138,32 @@ async function handleAlert(location, alertType) {
...
-        fallbackFromPhoneNumber: location.twilioNumber,
+      
+      if (location.siren_particle_id !== null) {
+        try{
+          await axios.post('https://api.particle.io/v1/devices/${location.siren_particle_id}/start-siren', {
+            arg: 'start',
+            access_token: helpers.getEnvVar('PARTICLE_ACCESS_TOKEN')
+          })
+        } catch (e) {
+          helpers.log(e)
```
trisapeace, line 149: Can you please use `helpers.logError` instead? That way it will let Sentry know when this happens as well.

index.js
trisapeace, line 144: So unfortunately `location` doesn't automatically get the `siren_particle_id` value just because it's been added to the DB. You'll need to update the `Location.js` class to add it. And then in `db.js`, there is a function called `createLocationFromRow` that will need to pass the `siren_particle_id` as a parameter to the contructor that you changed in `Location.js`. Then you'll be able to use it here.

trisapeace (commented):
Looking much better!

Still a few things to do that I've commented on. And I see that there are some conflicts with your base, probably because of merges that I've done into `devenv` after you branched off. We normally resolve these by rebasing, but it's a bit of a risky manoeuvre if you're not used to it. Let me know when you'd like to do this and we can do it together.

db/021-add-siren-particle-id.sql
phelanm20: Good catch!

phelanm20:
Updated pull request with changes. Should the updateLocation() function in db.js be updated with serial_particl_id as well?

trisapeace:
Right! That's a really good point about `updateLocation`. In general, yes, it should be. And that will take us down the rabbit hole because I think it's used by the Dashboard UI, which should also be updated to include this as an optional field. Maybe try to leave it as is for now and see how long we can get away without doing that.

.eslintrc.json
```diff
@@ -1,5 +1,7 @@
...
-    "extends": [
-        "eslint-config-brave"
-    ]
-}
+    "extends": ["airbnb", "prettier", "eslint-config-brave"],
+    "plugins": ["prettier"],
+    "rules": {
+        "prettier/prettier": ["error"]
+    }
+  }
```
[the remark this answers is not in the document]
phelanm20: Sadly I did not, this was in the guidelines of the link you sent

mariocimet:
This is looking really good. The bit in index.js where you call the particle api looks correct. To be sure, I'd like us to go through a test plan where we specify what behaviour we want to see when this is deployed. Do you remember how to deploy this version to the dev branch?

I'd also like us to have some test coverage for the new functionality - we can go through this together on a call.

trisapeace:
Task linked: [CU-18r17vf Integrate Sound/Light Firmware with Backend](https://app.clickup.com/t/18r17vf)

db/db.js
```diff
@@ -640,11 +640,11 @@ async function createLocationFromBrowserForm(locationid, displayName, doorCoreId
...
   try {
     await runQuery(
       'createLocation',
-      'INSERT INTO locations(locationid, responder_phone_number, movement_threshold, stillness_timer, duration_timer, reminder_timer, initial_timer, heartbeat_alert_recipients, twilio_number, fallback_phonenumbers, fallback_timer, display_name, door_particlecoreid, radar_particlecoreid, radar_type, alert_api_key, is_active, firmware_state_machine, client_id) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19)',
+      'INSERT INTO locations(locationid, responder_phone_number, movement_threshold, stillness_timer, duration_timer, reminder_timer, initial_timer, heartbeat_alert_recipients, twilio_number, fallback_phonenumbers, fallback_timer, display_name, door_particlecoreid, radar_particlecoreid, radar_type, alert_api_key, is_active, firmware_state_machine, sirenParticleId, client_id) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20)',
```
mariocimet, line 647: you're getting an error in the tests because here you're using the parameter name/"javascript-ish" name in the query string instead of the column name that you defined in your migration script. to fix, change sirenParticleId here to siren_particle_id

db/db.js
```diff
@@ -664,6 +664,7 @@ async function createLocation(locationid, phonenumber, movementThreshold, stilln
         alertApiKey,
         isActive,
         firmwareStateMachine,
+        sirenParticleId,
```
mariocimet, line 667: make sure this matches the order of parameters in the query string, sirenParticleId should go after clientId

test/integration/serverTest.js
```diff
@@ -154,6 +154,126 @@ describe('Brave Sensor server', () => {
...
+        'XeThru',
+        'alertApiKey',
+        true,
+        true,
+        'particleCoreIdTest',
+        client.id,
+      )
+      sandbox.stub(braveAlerter, 'startAlertSession')
+      sandbox.stub(braveAlerter, 'sendSingleAlert')
+      sandbox.spy(particle, 'callFunction')
```
mariocimet, line 188: in order to spy this function from the particle library, make sure to require it at the top of the file with this name, the same way you do in index.js:

`const particle = require('particle-api-js')`

test/integration/serverTest.js
```diff
@@ -154,6 +154,126 @@ describe('Brave Sensor server', () => {
...
+    
+    it('should return 200 to a request with no headers', async () => {
+      const response = await chai.request(server).post('/api/sensorEvent').send({})
+      expect(response).to.have.status(200)
+    })
+
+    it('should call particle.callFunction with the correct device ID if sirenParticleId is not null, and return 200', async () => {
+      const response = await firmwareAlert(radar_coreID, SENSOR_EVENT.STILLNESS)
+      expect(response).to.have.status(200)
+      expect(particle.callFunction).to.have.been.calledWith('particleCoreIdTest', helpers.getEnvVar('PARTICLE_ACCESS_TOKEN_TEST'))
```
mariocimet, line 207: I think actually here you want to use just helpers.getEnvVar('PARTICLE_ACCESS_TOKEN_TEST'), because helpers.getEnvVar adds '_TEST' when NODE_ENV=test, as it does when we run npm test

test/integration/serverTest.js
```diff
@@ -154,6 +154,126 @@ describe('Brave Sensor server', () => {
...
+        1000,
+        'locationName',
+        door_coreID,
+        radar_coreID,
+        'XeThru',
+        'alertApiKey',
+        true,
+        true,
+        'particleCoreIdTest',
+        client.id,
```
mariocimet, line 184: if you change the order of parameters above to match the query string, you should also change this to 
```
        client.id,
        'particleCoreIdTest',
```

mariocimet (changes requested):

db/db.js
phelanm20: Siren particle Id is second last in what I think is the query string - the reason why is I think that order was put in place somewhere when Theresa and I did the rebase (can't remember where though).

test/integration/serverTest.js
phelanm20: Do you mean replace line 207 with just helpers.getEnvVar('PARTICLE_ACCESS_TOkEN_TEST'),  or replace just the body of the calledWith() function?

test/integration/serverTest.js
phelanm20: Did not change the order of parameters - if the query string is different it is the what should be changed. I don't know where the query string is - could you point me to it?

db/db.js
mariocimet: oh! I think this was just my mistake oops

test/integration/serverTest.js
mariocimet: I meant replace the second parameter in the calledWith() function with just helpers.getEnvVar('PARTICLE_ACCESS_TOKEN)

test/integration/serverTest.js
mariocimet: my mistake! disregard

test/integration/serverTest.js
```diff
@@ -154,6 +155,138 @@ describe('Brave Sensor server', () => {
...
+      await db.clearSessions()
+      await db.clearLocations()
+      await db.clearClients()
+      sandbox.restore()
+      helpers.log('\n')
+    })
+
+    it('should return 200 to a request with no headers', async () => {
+      const response = await chai.request(server).post('/api/sensorEvent').send({})
+      expect(helpers.logError).to.have.been.called
```
trisapeace, line 202: In order to use this, you'll need `sandbox.spy(helpers, 'logError') in the `beforeEach`

phelanm20: done

package.json
```diff
@@ -23,6 +23,7 @@
     "chai-datetime": "^1.8.0",
     "cookie-parser": "^1.4.5",
     "cors": "^2.8.5",
+    "curl": "^0.1.4",
```
trisapeace, line 26: Please remove this and, in order to update the `package-lock.json` file as well, re-run `npm install`. The `dependencies` section is only for libraries used in the code. My guess is that you were using this on the command line for testing, right?

package.json
```diff
@@ -36,23 +37,24 @@
     "mustache": "^3.2.1",
     "particle-api-js": "^7.4.0",
     "pg": "^7.18.2",
+    "react": "^17.0.2",
```
trisapeace, line 40: Why did you need to add this?

test/integration/serverTest.js
```diff
@@ -154,6 +157,150 @@ describe('Brave Sensor server', () => {
...
+        true,
+        'particleCoreIdTest',
+        client.id,
+      )
+      sandbox.stub(braveAlerter, 'startAlertSession')
+      sandbox.stub(braveAlerter, 'sendSingleAlert')
+      sandbox.spy(particle, 'callFunction')
+      sandbox.spy(helpers, 'logError')
+    })
+    afterEach(async () => {
```
trisapeace, line 194: Can you please add a newline before the `afterEach` call?

test/integration/serverTest.js
```diff
@@ -154,6 +157,150 @@ describe('Brave Sensor server', () => {
...
+        deviceId: 'particleCoreIdTest',
+        name: 'start',
+        argument: 'start',
+        auth: helpers.getEnvVar('PARTICLE_ACCESS_TOKEN'),
+      })
+    })
+
+    it('should not call startAlertSession if particleSirenID is not null', async () => {
+      const sessions = await db.getAllSessionsFromLocation(testLocation1Id)
+      expect(sessions.length).to.equal(0)
```
trisapeace, line 230: The name of this test doesn't quite correspond with what you're actually testing. 
I'd either change the name of the test to `'should not create a new session if particleSirenID is not null`
or change the test to spy on the `braveAlerter.startAlertSession` function and use `expect(braveAlerter.startAlertSession).not.to.be.called` (my syntax might be off, but something like that).

test/integration/serverTest.js
```diff
@@ -154,6 +157,150 @@ describe('Brave Sensor server', () => {
...
+        true,
+        null,
+        client.id,
+      )
+      sandbox.stub(braveAlerter, 'startAlertSession')
+      sandbox.stub(braveAlerter, 'sendSingleAlert')
+      sandbox.spy(particle, 'callFunction')
+      sandbox.spy(helpers, 'logError')
+    })
+    afterEach(async () => {
```
trisapeace, line 268: Can you please add a newline above the `afterEach`?

test/integration/serverTest.js
```diff
@@ -154,6 +157,150 @@ describe('Brave Sensor server', () => {
...
+    it('should not call particle.callFunction if there sirenParticleId is null', async () => {
+      expect(particle.callFunction).to.have.not.been.called
+    })
+
+    it('should call startAlertSession if particleSirenId is null', async () => {
+      await firmwareAlert(radar_coreID, SENSOR_EVENT.STILLNESS)
+      const sessions = await db.getAllSessionsFromLocation(testLocation1Id)
+      const session = sessions[sessions.length - 1]
+      expect(session.alertType).to.equal(ALERT_TYPE.SENSOR_STILLNESS)
+    })
```
trisapeace, line 301: This one's name doesn't quite match up with what you're testing either. 
I'd either change the name to something like `should create a session with the right alert type if particleSirenId is null`
Or change the test to spy on `braveAlerter.startAlertSession` and test that it is called with the expected parameters by using an expects something like `expect(braveAlerter.startAlertSession).to.be.calledWithExactly({....})`

test/integration/serverTest.js
mariocimet: make sure to also add a call to firmwareAlert as the test stimulus for the API - right now its just running this db function and expect on an empty database that's had no input to it.

.env.example
```diff
@@ -62,4 +62,8 @@ SUBSEQUENT_ALERT_MESSAGE_THRESHOLD_TEST=1000
...
```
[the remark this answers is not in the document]
phelanm20: If it's section 3 of the README (environment variable changes) then I think both are updated. I will check tomorrow as James is testing now.  Is the production secret the same as the dev secret?

.eslintrc.json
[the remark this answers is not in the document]
phelanm20: Reverted

package.json
phelanm20: Yup, I'll remove this

package.json
phelanm20: I must have added this while debugging the node.js versioning issue following a stack overflow post. I'll remove it

test/integration/serverTest.js
phelanm20: Changed the name of the test and added a call to firmwareAlert

test/integration/serverTest.js
phelanm20: Updated the test name

test/integration/serverTest.js
```diff
@@ -154,6 +157,147 @@ describe('Brave Sensor server', () => {
...
+      expect(helpers.logError).to.have.been.called
+    })
+
+    it('should return 200 for a valid request', async () => {
+      const response = await firmwareAlert(radar_coreID, SENSOR_EVENT.STILLNESS)
+      expect(response).to.have.status(200)
+    })
+
+    it('should not call particle.callFunction if there sirenParticleId is null', async () => {
+      expect(particle.callFunction).to.have.not.been.called
```
trisapeace, line 290: This test needs a call to `firmwareAlert`

trisapeace:
Looks good. What sort of manual testing did you do to make sure that this works as expected?

PR #123, opened 2021-08-09
implement detailed heartbeats for FSM CU-1c1qd5u + Fix missing heartbeat alerts for FSM CU-4avrhp

otes10:
postgres db work will be the next step


db/redis.js
```diff
@@ -116,8 +116,17 @@ async function addVitals(locationid, signalStrength, cloudDisconnects) {
...
+// ignore comments included to allow arguments to be split across lines in pairs
+// prettier-ignore
+/* eslint-disable function-call-argument-newline */
+async function addEdgeDeviceHeartbeat(locationid, missedDoorMessages, doorLowBattery, doorHeartbeatReceived, stateTransitions) {
+  client.xadd(
+    `heartbeat:${locationid}`, 'MAXLEN', '~', '10000', '*',
+    'missedDoorMessages', missedDoorMessages,
+    'doorLowBattery', doorLowBattery,
+    'doorHeartbeatReceived', doorHeartbeatReceived,
+    'stateTransitions', stateTransitions
```
otes10, line 128: I left the state transitions array as one piece of a big redis entry rather than multiple redis entries

trisapeace: OK.

index.js
```diff
@@ -197,6 +197,21 @@ async function checkHeartbeat() {
...
+  const stateTransitionsObjects = []
+  function objToArray(value) {
+    const stateObject = {
+      state: value[0],
+      reason: reasonsTable[value[1]],
+      time: value[2],
+    }
+    stateTransitionsObjects.push(stateObject)
+  }
+  stateTransitionsArrays.forEach(objToArray)
```
trisapeace, line 211: This feels a bit awkward. It is not generally good practice for a function to modify a variable outside of it like this.
What would feel more natural to me is if you your `objToArray` function simply returned `stateObject`.

Then, instead of `stateTransitionsArrays.forEach(objToArray)`, you can use 
`const stateTransitionObjects = stateTransitionsArrays.map(objToArray)`
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map for more details about `map`.

You could also move the `objToArray` function (after giving it a more descriptive name) outside the `reformatHeartbeatStates` so that it doesn't need to be redefined every time `reformatHeartbeatStates` is called.

db/redis.js
trisapeace: What does this look like in Redis?
Is the `client.xadd` function smart enough to serialize the `stateTransitions` object into something pretty? 
Or should you use `JSON.stringify(stateTransitions)`?

index.js
```diff
@@ -197,6 +197,21 @@ async function checkHeartbeat() {
   }
 }
 
+function reformatHeartbeatStates(stateTransitionsArrays) {
```
trisapeace, line 200: This would be a great thing to write unit tests for.

index.js
```diff
@@ -442,11 +457,12 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
         response.status(200).json(errorMessage)
       } else {
         const message = JSON.parse(request.body.data)
-        const doorStatus = message.door_status
-        const doorTime = message.door_time
-        const insTime = message.ins_time
+        const missedDoorMessages = message.md
+        const doorLowBattery = message.lb
+        const doorHeartbeatReceived = message.dh
```
trisapeace, line 462: Just curious... is `doorHeartbeatReceived` going to be a timestamp? If so, which time zone will it be in?

trisapeace (commented):
This is looking really good.

A couple of housekeeping things:
- Can you please add a line in the CHANGELOG.md file in the [unreleased] section that just gives a short description of the change and the ClickUp reference number? This helps us keep track of what changes in each deployment.
- In the name of your GitHub pull request, can you please put the ClickUp reference number (it'll look something like CU-abiy82 where the stuff after CU- is in the URL of the ClickUp Task. That will help us keep the code the the Task associated because it will automatically make a comment on your PR with a link to the Task and will add a link to the PR from ClickUp, which can be helpful later on.

index.js
otes10: It will be a timestamp, in milliseconds relative to the last door heartbeat recorded

trisapeace: That's a good way to get around the timezone issue :smile:

db/redis.js
trisapeace, line 128: It would be great if you could htink about all these names a little more to see if you can make it clearer what types of values we would expect from them. For example, this one might be better as `millisSinceDoorHeatbeat`.

trisapeace:
Task linked: [CU-1c1qd5u Implement detailed heartbeats for FSM](https://app.clickup.com/t/1c1qd5u)

trisapeace:
Task linked: [CU-4avrhp Fix missing low battery alerts with Firmware State Machine](https://app.clickup.com/t/4avrhp)

db/redis.js
otes10: After some testing, it turns out redis won't accept arrays as an value. I will need to stringify

PR #125, opened 2021-08-20
Brave Siren Milestone2 (CU-1dffcm4)

phelanm20:
Added two app.post handlers for messages from the siren particle.

"addressed" handler:
Siren particle calls this when a responder pushes the physical button on the Brave Siren. This handler should stop the current alert session.

"escalate" handler:
Siren particle calls this when a responder does NOT push the physical button on the Brave Siren before the timer in the particle firmware runs out. This handler should start the normal chatbot.

### DIFF
--- a/index.js
+++ b/index.js
@@ -485,7 +485,7 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
   }
 })
 
-app.post('/api/sirenAddressed', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
+app.post('/api/sirenAddressed', Validator.body(['coreid']).exists(), async (request, response) => {
   try {
     const validationErrors = Validator.validationResult(request).formatWith(helpers.formatExpressValidationErrors)
 
@@ -514,7 +514,7 @@ app.post('/api/sirenAddressed', Validator.body(['coreid', 'event', 'data']).exis
             session.chatbotState = CHATBOT_STATE.COMPLETED
             await db.saveSession(session, client)
           } else {
-            helpers.logError(`Error stopping session and chatbot due to siren button press`)
+            helpers.logError(`Error stopping session and chatbot due to siren ${coreId} button press`)
           }
 
           await db.commitTransaction(client)
@@ -543,7 +543,7 @@ app.post('/api/sirenAddressed', Validator.body(['coreid', 'event', 'data']).exis
   }
 })
 
-app.post('/api/sirenEscalated', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
+app.post('/api/sirenEscalated', Validator.body(['coreid']).exists(), async (request, response) => {
   try {
     const validationErrors = Validator.validationResult(request).formatWith(helpers.formatExpressValidationErrors)
 
--- a/index.js
+++ b/index.js
@@ -482,7 +482,7 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
   }
 })
 
-app.post('/api/siren-addressed', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
+app.post('/api/sirenAddressed', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
   try {
     const validationErrors = Validator.validationResult(request).formatWith(helpers.formatExpressValidationErrors)
 
@@ -501,26 +501,27 @@ app.post('/api/siren-addressed', Validator.body(['coreid', 'event', 'data']).exi
         try {
           client = await db.beginTransaction()
           if (client === null) {
-            helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
+            helpers.logError(`sirenAddressedCallback: Error starting transaction`)
             return
           }
 
           const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
           if (session) {
             session.respondedAt = await db.getCurrentTime(client)
+            session.chatbotState = CHATBOT_STATE.COMPLETED
             await db.saveSession(session, client)
           } else {
-            helpers.logError(`error stopping session and chatbot due to siren button press`)
+            helpers.logError(`Error stopping session and chatbot due to siren button press`)
           }
 
           await db.commitTransaction(client)
         } catch (e) {
           try {
             await db.rollbackTransaction(client)
-            helpers.logError(`alertSessionChangedCallback: Rolled back transaction because of error: ${e}`)
+            helpers.logError(`sirenAddressedCallback: Rolled back transaction because of error: ${e}`)
           } catch (error) {
             // Do nothing
-            helpers.logError(`alertSessionChangedCallback: Error rolling back transaction: ${e}`)
+            helpers.logError(`sirenAddressedCallback: Error rolling back transaction: ${e}`)
           }
         }
         response.status(200).json('OK')
@@ -539,7 +540,7 @@ app.post('/api/siren-addressed', Validator.body(['coreid', 'event', 'data']).exi
   }
 })
 
-app.post('/api/siren-escalate', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
+app.post('/api/sirenEscalated', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
   try {
     const validationErrors = Validator.validationResult(request).formatWith(helpers.formatExpressValidationErrors)
 
@@ -558,7 +559,7 @@ app.post('/api/siren-escalate', Validator.body(['coreid', 'event', 'data']).exis
         try {
           client = await db.beginTransaction()
           if (client === null) {
-            helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
+            helpers.logError(`sirenEscalatedCallback: Error starting transaction`)
             return
           }
 
@@ -574,10 +575,10 @@ app.post('/api/siren-escalate', Validator.body(['coreid', 'event', 'data']).exis
         } catch (e) {
           try {
             await db.rollbackTransaction(client)
-            helpers.logError(`alertSessionChangedCallback: Rolled back transaction because of error: ${e}`)
+            helpers.logError(`sirenEscalatedCallback: Rolled back transaction because of error: ${e}`)
           } catch (error) {
             // Do nothing
-            helpers.logError(`alertSessionChangedCallback: Error rolling back transaction: ${e}`)
+            helpers.logError(`sirenEscalatedCallback: Error rolling back transaction: ${e}`)
           }
         }
         response.status(200).json('OK')
--- a/index.js
+++ b/index.js
@@ -569,6 +569,19 @@ app.post('/api/sirenEscalated', Validator.body(['coreid', 'event', 'data']).exis
           const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
           if (session) {
             session.chatbotState = CHATBOT_STATE.STARTED
+            const alertInfo = {
+              sessionId: session.id,
+              toPhoneNumber: location.responderPhoneNumber,
+              fromPhoneNumber: location.twilioNumber,
+              message: `This is a ${alertTypeDisplayName} alert. Please check on the bathroom at ${location.displayName}. Please respond with 'ok' once you have checked on it.`,
+              reminderTimeoutMillis: location.reminderTimer,
+              fallbackTimeoutMillis: location.fallbackTimer,
+              reminderMessage: `This is a reminder to check on the bathroom`,
+              fallbackMessage: `An alert to check on the bathroom at ${location.displayName} was not responded to. Please check on it`,
+              fallbackToPhoneNumbers: location.fallbackNumbers,
+              fallbackFromPhoneNumber: location.client.fromPhoneNumber,
+            }
+            braveAlerter.startAlertSession(alertInfo)
             await db.saveSession(session, client)
           } else {
             helpers.logError(`Siren not responded to - error starting normal chatbot`)
--- a/index.js
+++ b/index.js
@@ -568,7 +568,6 @@ app.post('/api/sirenEscalated', Validator.body(['coreid', 'event', 'data']).exis
 
           const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
           if (session) {
-            session.chatbotState = CHATBOT_STATE.STARTED
             const alertInfo = {
               sessionId: session.id,
               toPhoneNumber: location.responderPhoneNumber,
--- a/index.js
+++ b/index.js
@@ -509,7 +509,6 @@ app.post('/api/siren-addressed', Validator.body(['coreid', 'event', 'data']).exi
           const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
           if (session) {
             session.respondedAt = await db.getCurrentTime(client)
-            session.chatbotState = CHATBOT_STATE.COMPLETED    
             await db.saveSession(session, client)
           } else {
             helpers.logError(`error stopping session and chatbot due to siren button press`)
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
   }
 })
 
+app.post('/api/siren-addressed', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
+  try {
+    const validationErrors = Validator.validationResult(request).formatWith(helpers.formatExpressValidationErrors)
+
+    if (validationErrors.isEmpty()) {
+      const coreId = request.body.coreid
+      const location = await db.getLocationFromParticleCoreID(coreId)
```
trisapeace, line 491: I think that `getLocationFromParticleCoreID` only looks at the `door_particlecoreid` and `radar_particlecoreid` columns. You can modify it to also look at `siren_particle_id` so that this will work.

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
   }
 })
 
+app.post('/api/siren-addressed', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
```
trisapeace, line 485: I'd prefer camelCase for the route name, so `/api/siredAddressed` to match the style of `/api/sensorEvent`.

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+        helpers.logError(errorMessage)
+        // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
+        response.status(200).json(errorMessage)
+      } else {
+        let client
+
+        try {
+          client = await db.beginTransaction()
+          if (client === null) {
+            helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
```
trisapeace, line 504: I think `alertSessionChangedCallback` was copied here but should be changed to be more descriptive of where this error is coming from.

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+            return
+          }
+
+          const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
+          if (session) {
+            session.respondedAt = await db.getCurrentTime(client)
+            await db.saveSession(session, client)
+          } else {
+            helpers.logError(`error stopping session and chatbot due to siren button press`)
```
trisapeace, line 513: Can you please start your error message with a capital letter to match the others that start with non-function names?

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+          }
+
+          await db.commitTransaction(client)
+        } catch (e) {
+          try {
+            await db.rollbackTransaction(client)
+            helpers.logError(`alertSessionChangedCallback: Rolled back transaction because of error: ${e}`)
```
trisapeace, line 520: Same here about `alertSessionChangedCallback`

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+          } catch (error) {
+            // Do nothing
+            helpers.logError(`alertSessionChangedCallback: Error rolling back transaction: ${e}`)
```
trisapeace, line 523: Same here about `alertSessionChangedCallback`

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+          client = await db.beginTransaction()
+          if (client === null) {
+            helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
+            return
+          }
+
+          const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
+          if (session) {
+            session.respondedAt = await db.getCurrentTime(client)
+            await db.saveSession(session, client)
```
trisapeace, line 511: This looks good. But you'll also want to update the session's state to `COMPLETED` at the same time.

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+    }
+  } catch (err) {
+    const errorMessage = `Error calling ${request.path}: ${err.toString()}`
+    helpers.logError(errorMessage)
+    // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
+    response.status(200).json(errorMessage)
+  }
+})
+
+app.post('/api/siren-escalate', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
```
trisapeace, line 542: Same here about camelCase.

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+        helpers.logError(errorMessage)
+        // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
+        response.status(200).json(errorMessage)
+      } else {
+        let client
+
+        try {
+          client = await db.beginTransaction()
+          if (client === null) {
+            helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
```
trisapeace, line 561: Check for these wrong function names in error messages here and a couple times below too.

index.js
```diff
@@ -482,6 +482,120 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+            return
+          }
+
+          const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
+          if (session) {
+            session.chatbotState = CHATBOT_STATE.STARTED
+            await db.saveSession(session, client)
```
trisapeace, line 568: To start a regular session, it would need to call `braveAlerter.startAlertSession()` like in `handleAlert`, but I'm reconsidering this as a good fallback strategy because calling `startAlertSession` will create a new session in the `sessions` table, whereas you already have a session. Maybe you should instead call `braveAlerter.sendSingleAlert` to send a single text message to the fallback number and you can update the session state to `WAITING_FOR_REPLY`.

On the firmware side of things, you're planning to shut of the noise and lights when this function returns no matter what, right?

index.js
trisapeace, line 485: Will this webhook be called if someone pushes the "off" button when the siren isn't actually making noise?

trisapeace (commented):
Looking good. Let me know if you have any questions about my comments.

index.js
phelanm20: It was copied, changed to sirenAddressedCallback

index.js
phelanm20: Changed to sirenEscalatedCallback

index.js
phelanm20: I'll wait for @mariocimet 's feedback on this as I am still relatively unfamiliar with the code base. On the firmware side of things, yes, the siren shuts off after the timer runs out (modifiable to a variety of times).

index.js
phelanm20: No, the interrupt is detached after the siren ends

trisapeace: So if a person pushes the "off" button a whole bunch of times in quickly in a row, then the first push would stop the interrupt and only one webhook call would be made to this endpoint?

phelanm20: Yes - if I'm wrong it would be a hardware thing. I can test this specifically once end to end testing is set up. I've made a note of it.

index.js
```diff
@@ -485,6 +485,133 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
   }
 })
 
+app.post('/api/sirenAddressed', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
```
trisapeace, line 488: If you don't use `event` or `data`, then you should remove them from the `Validator.body()` call.

index.js
```diff
@@ -485,6 +485,133 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+            return
+          }
+
+          const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
+          if (session) {
+            session.respondedAt = await db.getCurrentTime(client)
+            session.chatbotState = CHATBOT_STATE.COMPLETED
+            await db.saveSession(session, client)
+          } else {
+            helpers.logError(`Error stopping session and chatbot due to siren button press`)
```
trisapeace, line 517: It might be helpful to include the `siren_particle_id` in this error log message to help with tracking down any issues that come up in the future.

index.js
```diff
@@ -485,6 +485,133 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+          if (client === null) {
+            helpers.logError(`sirenAddressedCallback: Error starting transaction`)
+            return
+          }
+
+          const session = await db.getUnrespondedSessionWithLocationId(location.locationid, client)
+          if (session) {
+            session.respondedAt = await db.getCurrentTime(client)
+            session.chatbotState = CHATBOT_STATE.COMPLETED
+            await db.saveSession(session, client)
```
trisapeace, line 515: Looks good.

index.js
```diff
@@ -485,6 +485,133 @@ app.post('/api/heartbeat', Validator.body(['coreid', 'event', 'data']).exists(),
...
+    }
+  } catch (err) {
+    const errorMessage = `Error calling ${request.path}: ${err.toString()}`
+    helpers.logError(errorMessage)
+    // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
+    response.status(200).json(errorMessage)
+  }
+})
+
+app.post('/api/sirenEscalated', Validator.body(['coreid', 'event', 'data']).exists(), async (request, response) => {
```
trisapeace, line 546: If you don't use `event` or `data` , then you should remove them from the `Validator.body()` call.

test/integration/serverTest.js
```diff
@@ -225,6 +225,26 @@ describe('Brave Sensor server', () => {
...
+      expect(response).to.have.status(200)
+    })
+
+    it('should start an alert session if a sirenEscalated call is receieved', async () => {
+      await firmwareAlert(radar_coreID, SENSOR_EVENT.STILLNESS)
+      await chai.request(server).post('/api/sirenEscalated').send({})
+      const sessions = await db.getAllSessionsFromLocation(testLocation1Id)
+      const session = sessions[sessions.length - 1]
+      expect(session.alertType).to.equal(ALERT_TYPE.SENSOR_STILLNESS)
+    })
```
trisapeace, line 246: All of these tests are going to get a validation error because you aren't sending `coreid` in the body.  I suggest making new helper functions for calling `POST /api/sirenAddressed` and `POST /api/sirenEscalated` similar to the `firmwareAlert` function.

It will also be good to have a test to see that the state was actually changed to `COMPLETED` when you call `POST /api/sirenAddressed`.

test/integration/serverTest.js
```diff
@@ -225,6 +225,26 @@ describe('Brave Sensor server', () => {
...
+      expect(response).to.have.status(200)
+    })
+
+    it('should send a 200 response if a sirenEscalated call is receieved', async () => {
+      await firmwareAlert(radar_coreID, SENSOR_EVENT.STILLNESS)
+      const response = await chai.request(server).post('/api/sirenEscalated').send({})
+      expect(response).to.have.status(200)
+    })
+
+    it('should start an alert session if a sirenEscalated call is receieved', async () => {
```
mariocimet, line 240: is this actually the case? As I understand it, we want to start an alert session when the Sensor raises a DURATION or STILLNESS alert (which happens in the sensorEvent endpoint). When the sirenEscalated endpoint is hit, my understanding is that we update an existing session, and escalate to the fallback number"? Does that make sense?

trisapeace: I had originally suggested to Maria that starting the regular chatbot session might be a good way to escalate the Siren, but I have changed my mind and now thing that just hitting the fallback would be better. We were waiting to hear your thoughts on it, but it sounds like you're also thinking that a fallback at this point would be the way to go, is that right?

test/integration/serverTest.js
phelanm20: I'm confused what function body I need to send coreid in. Radar_coreID is defined in the beforeEach section.

trisapeace: It needs to be set in the body that is being sent to `POST /api/sirenEscalated`. If you look in `firmwareAlert` you can see how Mario's sending a JSON object with `coreid` and other things to `POST /api/sensorEvent`.

phelanm20: What function body does he coreid need be included in?

trisapeace: Inside the curly braces of the `send({})` in  `await chai.request(server).post('/api/sirenEscalated').send({})`

test/integration/serverTest.js
phelanm20: Yes, just looked into it and you're right. A session is started as soon as the siren is called.

index.js
```diff
@@ -568,19 +568,8 @@ app.post('/api/sirenEscalated', Validator.body(['coreid']).exists(), async (requ
...
-              reminderTimeoutMillis: location.reminderTimer,
-              fallbackTimeoutMillis: location.fallbackTimer,
-              reminderMessage: `This is a reminder to check on the bathroom`,
-              fallbackMessage: `An alert to check on the bathroom at ${location.displayName} was not responded to. Please check on it`,
-              fallbackToPhoneNumbers: location.fallbackNumbers,
-              fallbackFromPhoneNumber: location.client.fromPhoneNumber,
-            }
-            braveAlerter.startAlertSession(alertInfo)
+            alertMessage= 'There is an unresponded siren. Please check on the bathroom at ${location.displayName}.'
+            braveAlerter.sendSingleAlert(location.responderPhoneNumber,location.twilioNumber,alertMessage)
```
trisapeace: Can you please change this to send it to the fallback phone numbers? It'll be an array of numbers, so you'll probably need to loop over them. The sender number should be `location.client.fromPhoneNumber` instead of `location.twilioNumber`.

There also needs to be a space after each comma in `(location.responderPhoneNumber,location.twilioNumber,alertMessage)` and a space before the equal sign in ` alertMessage= `.

You also need to use backquotes instead of single quotes in `'There is an unresponded siren. Please check on the bathroom at ${location.displayName}.'` in order for the `${}` part to work.

test/integration/serverTest.js
```diff
@@ -225,6 +259,30 @@ describe('Brave Sensor server', () => {
...
+      await firmwareAlert(radar_coreID, SENSOR_EVENT.STILLNESS)
+      await sirenAddressedAlert('particleCoreIdTest')
+      const session = await db.getMostRecentSession(testLocation1Id)
+      expect(session.chatbotState).to.equal(CHATBOT_STATE.COMPLETED)
+    })
+
+    it('should not call startAlertSession if particleSirenID is not null', async () => {
+      const sessions = await db.getAllSessionsFromLocation(testLocation1Id)
+      expect(sessions.length).to.equal(0)
+    })
```
trisapeace, line 285: We should also have tests to answer the questions:
- What happens with the given Siren Core ID isn't in the DB?
- What happens if not given a Siren Core ID
- Is the UpdatedAt value updated?
- Is the SendSingleAlert function called with the right parameters?

trisapeace (commented):
Getting close!

phelanm20:
Current functional state defined below:
When a sirenEscalated message is received on the backend and there are no existing sessions, nothing happens. Based on the code, it is expected that some kind of text message should be received. When a sirenEscalated message is received on the backend and there ARE existing sessions, an "additional session" text message is received on the responder phone. The second issue is that if there is an existing session, even if it was responded to, additional siren start messages do NOT start the siren.

trisapeace:
Task linked: [CU-1dffcm4 Brave Siren Backend Milestone 2](https://app.clickup.com/t/1dffcm4)

trisapeace:
Closing this PR in favour of this one: https://github.com/bravetechnologycoop/BraveSensor-Server/pull/127 because the rebase was going to be too difficult.

PR #131, opened 2021-10-05
No longer update Responded At when given an invalid incident category (CU-1ae1j60)

trisapeace:
- Also updated the tests for this function to better match the tests in
Buttons because I think the Buttons tests were written better for what
ended up being the implementation of this function. The tests that were
here in Sensors looked like they belonged with an older implementation
that we didn't end up using.

## Test plan
- :heavy_check_mark:  Deploy to `dev.sensors.brave.coop`
- :heavy_check_mark:  Trigger an alert on my XeThru+Argon device and see the respondedAt value is `null`
   - :heavy_check_mark: Respond 'OK' and see the respondedAt value change to the current time on the Dashboard
   - :heavy_check_mark: Respond to the incident category question with a valid incident category and see the respondedAt value does not change
   - :heavy_check_mark: Respond to the incident category question with an invalid incident category and see the respondedAt value does not change
   - :heavy_check_mark: Do not respond 'OK' to the button press and see that the respondedAt value remains `null` after the reminder message and after the fallback message
- :heavy_check_mark: Trigger an alert on my INS+Boron device and see the respondedAt value is `null`
   - :heavy_check_mark:  Respond 'OK' and see the respondedAt value change to the current time on the Dashboard
   - :heavy_check_mark: Respond to the incident category question with a valid incident category and see the respondedAt value does not change
   - :heavy_check_mark: Respond to the incident category question with an invalid incident category and see the respondedAt value does not change
   - :heavy_check_mark: Do not respond 'OK' to the button press and see that the respondedAt value remains `null` after the reminder message and after the fallback message

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,10 +10,17 @@
       "request": "launch",
       "name": "Mocha All within 'test' folder",
       "program": "${workspaceFolder}/server/node_modules/mocha/bin/_mocha",
-      "args": ["--timeout", "999999", "--require", "${workspaceFolder}/server/test/mochaFixtures.js", "--colors", "${workspaceFolder}/test/**/*"],
+      "args": [
+        "--timeout",
+        "999999",
+        "--require",
+        "${workspaceFolder}/server/test/mochaFixtures.js",
+        "--colors",
+        "${workspaceFolder}/server/test/**/*"
+      ],
       "console": "integratedTerminal",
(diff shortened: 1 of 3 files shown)


trisapeace:
Task linked: [CU-1ae1j60 Fix bug where `responded_at` is updated when the user gives an invalid incident category](https://app.clickup.com/t/1ae1j60)

PR #132, opened 2021-10-07
Transition to monorepo, add firmware unit tests for Particle console functions, add more input validation for Particle console functions (CU-vbep38) (CU-un6ybt)

mariocimet:
A monorepo lets us keep track of changes across firmware and server more easily. Note that only v4.0 of the firmware is to be included in the monorepo. Our practice in this repo for server code has been to only keep one, continuously updated version, and the intention is to adopt the same practice with firmware. We can keep the existing firmware repo around as a store for legacy/deprecated code until every device in the field has switched over.

I've started using Catch2 to write tests for the firmware code, starting with the Particle console functions. This required standing up a fair bit of mocks and more thorough testing will require more still. Along the way I noticed some gaps in input validation for the console functions so I've added more of that too.

One quirk is that now, to use the VSCode Particle Workbench, you need to open the firmware subfolder of the respository. One next direction I'd like us to take is to transition away from using that and simply use the particle api to compile/flash firmware, maybe through npm scripts.

Because files changed is less useful, here's a list of files/directories where I've made changes. In the firmware:
```
firmware/test (everything inside as well like mocks etc)
firmware/inc - particle custom String library
firmware/src/consoleFunctions.cpp - changes to input validation
README.md - transition from all versions in one repo to just having information for v4
```
I've also changed `.travis.yml` to run the firmware tests

Test Plan:
- Deploy Server to dev, and flash firmware to a test boron and perform basic verification tests:
  - alerts are properly started from Stillness and Duration conditions
  - heartbeats are properly sent and stored in redis
  - console functions reject badly formed inputs
    - empty string or incomplete IDs for doorID
    - negative numbers for timer/threshold functions


trisapeace:
Task linked: [CU-un6ybt Create Sensors monorepo with firmware](https://app.clickup.com/t/un6ybt)

README.md
```diff
@@ -1,321 +1,7 @@
-# BraveSensor-Server
+# BraveSensor
 
 [![Build Status](https://travis-ci.com/bravetechnologycoop/BraveSensor-Server.svg?branch=master)](https://travis-ci.com/bravetechnologycoop/BraveSensor-Server)
 
-# How to release a new version of BraveSensor-Server
+This is the BraveSensors monorepo. The Firmware and Server sections each have their own README.
```
trisapeace, line 5: Makes sense. Good idea.

firmware/README.md
```diff
@@ -0,0 +1,531 @@
+# BraveSensor-Firmware
+
+## Table of Contents
```
trisapeace, line 3: I like Heidi's table of contents (or whoever made this). We should eventually do this in the others too.

server/README.md
```diff
@@ -0,0 +1,321 @@
+# BraveSensor-Server
+
+[![Build Status](https://travis-ci.com/bravetechnologycoop/BraveSensor-Server.svg?branch=master)](https://travis-ci.com/bravetechnologycoop/BraveSensor-Server)
+
+# How to release a new version of BraveSensor-Server
+
+## 1. Update git to reflect the new release
+
+1. on your local machine, in the `BraveSensor-Server` repository:
```
trisapeace, line 9: Are we going to rename the repo as part of this work?

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+   1. pull the latest code ready for release: `git checkout devenv && git pull origin devenv`
+
+   1. decide on an appropriate version number for the new version
```
trisapeace, line 13: This is going to be interesting since what will be a breaking change in firmware won't be in server, and vice versa. Maybe we need to start using two tags? Like a `server-v5.0.0` and `firmware-v1.3.0`?

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+   1. update `CHANGELOG.md` by moving everything in `Unreleased` to a section for the new version
```
trisapeace, line 15: If we do separate tags, then I guess this `CHANGELOG.md` can move into `/server` and a new one would be needed in `/firmware`.

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+   1. `cd` into the `~/BraveSensor-Server/sensor-helm-chart` directory
```
trisapeace, line 17: Needs `server` in the path.

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+      1. Update `Chart.yaml` with a new `version` and new `appVersion`, if applicable
+
+      1. Run `git log -1 --format=%h` to get the container image tag
+
+      1. Update `values.yaml` by putting the tag of your new container image in the `tag` field of `image`
+
+   1. make a new commit directly on `devenv` which updates the changelog and helm chart
+
+   1. tag the new commit - for example, if the version number is v1.0.0, use `git tag v1.0.0`
```
trisapeace, line 27: Here we could change the name of the tag to specify that it's the server?

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+## 4. Deploy changes to the Kubernetes Cluster
+
+We use [helm](https://helm.sh/docs/) to manage our deployments. Helm allows us to deploy different instances of the application corresponding to dev, staging, and production environments as different 'releases'. Before updating production, we run smoke tests to verify that there's nothing obviously wrong with a release candidate.
+
+### Smoke testing
+
+1. Run `ssh brave@sensors-admin.brave.coop`
+
+1. cd into the ~/BraveSensor-Server directory and run `git pull origin master` to get the latest version of the helm chart
```
trisapeace, line 75: Need to add `server` to the path.

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Updating production
+
+1. Run the command `helm upgrade production sensor-helm-chart`
+
+# Deploying to the staging or development environment
+
+1. Run `ssh brave@sensors-admin.brave.coop`
+
+1. cd into the ~/BraveSensor-Server directory
```
trisapeace, line 91: Need to add `server` to the path.

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Environment Variables
+
+For local development, we use .env files to set environment variables. To set one of these up, copy the `.env.example` and rename it to `.env`, and fill in the values as appropriate.
+
+## Database Setup
+
+You will need to populate the `.env` file with connection parameters for both redis and Postgresql. If you are using a local database you will also need to setup the database schema.
+
+To do this, cd into the BraveSensor-Server directory and run the following command
```
trisapeace, line 153: Need to add `server` to this directory name.

server/README.md
```diff
@@ -0,0 +1,321 @@
...
+| ----------- | --------- | ------------------- |
+| Production  | doadmin   | backend-replacement |
+| Staging     | bravetest | staging             |
+| Development | bravetest | bravetest           |
+
+To access a database shell
+
+1. Log in to Digital Ocean
+
+1. Navigate to Databases --> odetect-db
```
trisapeace, line 239: Oh, the DB name can be updated to `odetect-db-jul-22-backup` while you're in here.

firmware/src/BraveSensorProductionFirmware.ino
```diff
@@ -0,0 +1,69 @@
+/*
```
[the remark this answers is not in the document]
trisapeace: I see the answer in the PR description now. I'll take a look though those files!

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ the code was deployed.
 - Low battery checks in heartbeat api for firmware state machine (CU-4avrhp).
 - Tracking of reset reason to reconnection alerts (CU-z9e1md).
 - Heartbeat messages enriched, including state transition data (CU-1c1qd5u).
+- `/firmware` folder with production code for Boron devices (CU-un6ybt).
```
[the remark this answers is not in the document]
mariocimet: my initial instinct was for only the existing production version to be included in there. Siren is a bit weird because it's not legacy code, nor is it slated for production - but given that it's merged into the server code... I think you're right and i should put it in there.

For Argons, my understanding is that the code is deprecated/legacy until and if we run into the case where Boron's don't work. If that happens we go to the old repo and fish it out to integrate it

firmware/.gitignore
```diff
@@ -0,0 +1,15 @@
+#ignore files that may contain sensitive client information
+SetupFiles/
+setupFirmware.h
+
+#ignore vscode settings
+.vscode/
+
+#ignore Argon binary files
```
trisapeace, line 8: Can we also ignore the binaries for the tests? Like `firmware/test/ConsoleTests`?

firmware/test/base.h
```diff
@@ -0,0 +1,24 @@
...
+#include <iostream>
+
+// Mocks
+#include "mocks/eeprom.h"
+#include "mocks/Particle.h"
+#include "../inc/spark_wiring_string.h"
+#include "../inc/spark_wiring_string.cpp"
+#include "../src/im21door.h"
+
+
```
trisapeace, line 14: Any particular reason for the extra whitespace?

firmware/test/consoleFnTests.cpp
```diff
@@ -0,0 +1,338 @@
...
+            }
+        }
+
+        WHEN( "the function is called with 0" ) {
+            toggle_debugging_publishes("0");
+
+            THEN( "debug flag should remain false" ) {
+                REQUIRE( stateMachineDebugFlag == false );
+            }
+        }
```
trisapeace, line 26: What should the return value be in these two cases?

firmware/test/consoleFnTests.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+        WHEN( "the function is called with 0" ) {
+            toggle_debugging_publishes("0");
+
+            THEN( "debug flag should be false" ) {
+                REQUIRE( stateMachineDebugFlag == false );
+            }
+        }
```
trisapeace, line 58: What should the return value be in these two cases?

firmware/src/consoleFunctions.cpp
```diff
@@ -0,0 +1,271 @@
...
+  }
+  //else parse new threshold
+  else {
+    int timeout = input.toInt();
+    //increase timeout value to from seconds to ms
+    timeout = timeout*1000;
+
+    if(timeout == 0){
+        //string.toInt() returns 0 if input not an int
+        //and a threshold value of 0 makes no sense, so return -1
```
trisapeace, line 100: This is unfortunate because I can see a use case for setting initial_timer to 0. In fact, I've tried it before in my testing because I don't think we actually need the inital_timer state at all with the simplified state machine. But anyways, I can see why it needs to be here, so no need to change anything.

firmware/test/consoleFnTests.cpp
```diff
@@ -0,0 +1,338 @@
+#define CATCH_CONFIG_MAIN
```
trisapeace, line 1: I suggest naming this function with `consoleFunctionsTests.cpp` to match the exact name of the file that you're testing.

firmware/test/consoleFnTests.cpp
```diff
@@ -0,0 +1,338 @@
...
+                REQUIRE( stateMachineDebugFlag == false );
+            }
+
+            THEN( "the return value should be -1" ) {
+                REQUIRE( returnVal == -1 );
+            }
+
+        }
+    }
+        GIVEN( "A true debug flag" ) {
```
trisapeace, line 41: All these tests look great. But could you please run through this file the check that the indentation is consistent and that there aren't any extra newlines? Thanks!

firmware/test/mocks/Particle.h
```diff
@@ -0,0 +1,48 @@
...
+
+public:
+    bool connected() const
+    {
+        return true;
+    }
+
+    bool publish(char const* const szEventName,
+                 char const* const szData,
+                 int /*PublishFlag*/ const flags)
```
trisapeace, line 36: What is this `/*PublishFlag*/` comment about?

firmware/test/mocks/Particle.h
```diff
@@ -0,0 +1,48 @@
...
+
+enum PublishFlag
+{
+    PUBLIC,
+    PRIVATE,
+    NO_ACK,
+    WITH_ACK
+};
+
+struct os_queue_t{
```
trisapeace, line 17: To be consistent, I think this open brace should be on the next line.

firmware/test/mocks/Particle.h
mariocimet: it's one of the custom types that Particle defines that I ended up not mocking anywhere, it doesn't need to be there.

server/README.md
mariocimet: is this true? I think breaking change still has a system-wide definition, can you think of an example of one of these changes that "breaks" in firmware and not in server?

trisapeace (approved):
Approved assuming README changes  are done as discussed in the code review meeting today and that `brave-alert-lib` dependency is updated to `v6.0.0` in the Server code.

PR #139, opened 2021-11-22
Added cluster setup script

mariocimet:
This PR adds a script to be run in conjunction with instructions on the Living Doc for creating a new Kubernetes cluster

Test plan: 

:heavy_check_mark:  Created a cluster using this script and these manifests with @trisapeace and verified a working dashboard


server/cluster-setup/clusterSetup.sh
```diff
@@ -0,0 +1,18 @@
...
+helm repo add jetstack https://charts.jetstack.io # cert manager
+helm repo add bitnami https://charts.bitnami.com/bitnami # redis
+helm repo update
+
+# install packages onto Kubernetes
+helm install sensor ingress-nginx/ingress-nginx --set controller.publishService.enabled=true # install nginx ingress controler
+helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --version v1.6.0 --set installCRDs=true # install cert-manager
+kubectl create namespace redis
+kubectl apply -f redis-storage.yaml
+helm install sensor-redis bitnami/redis --namespace redis --set persistence.storageClass=redis-storage --set auth.enabled=false --set architecture=standalone
```
trisapeace, line 17: What do you think about specifying the version of these helm charts to avoid problems of flags changing? It looks like it can be done with the `--version` flag like the `cert-manager` on line 14.

trisapeace (approved):

server/cluster-setup/clusterSetup.sh
mariocimet: let me look into this, because that flag I think refers to the cert-manager version, not the chart version - and you can have more than one chart (and accompanying flags) for the same app version. Will update after checking

trisapeace: Thanks :smile:

trisapeace: We didn't end up doing this.

PR #146, opened 2021-12-23
Trim all values when adding or editing Clients and Locations through the API (CU-1pdmvee)

trisapeace:
## Test Plan

- :heavy_check_mark: Add a new client with leading, trailing, and internal whitespace for all values where this is possible through the UI. See that
   - :heavy_check_mark: client is added in DB without the leading and trailing whitespace
- :heavy_check_mark: Try to add a new client with only whitespace for each value individually for all values where this is possible through the UI. See that
   - :heavy_check_mark: client cannot be added if the value was required
   - :heavy_check_mark: uses 'null' when the value was not required
- :heavy_check_mark: Add a new client with an existing `displayName` but with leading and trailing whitespace. See that
   - :heavy_check_mark: client cannot be added
- :heavy_check_mark:  Edit an existing client with leading, trailing, and internal whitespace for all values where this is possible through the UI. See that
   - :heavy_check_mark: client is updated in DB without the leading and trailing whitespace
- :heavy_check_mark: Try to edit an existing client with only whitespace for each value individually for all values where this is possible through the UI. See that
   - :heavy_check_mark: client cannot be updated if the value was required
   - :heavy_check_mark: uses 'null' when the value was not required
- :heavy_check_mark: Edit an existing client with an existing `displayName` but with leading and trailing whitespace. See that
   - :heavy_check_mark: client cannot be updated

- :heavy_check_mark: Add a new location with leading, trailing, and internal whitespace for all values where this is possible through the UI. See that
   - :heavy_check_mark: location is added in DB without the leading and trailing whitespace
- :heavy_check_mark: Try to add a new location
   - :heavy_check_mark: location cannot be added if the value was required
- :heavy_check_mark: Edit an existing location with leading, trailing, and internal whitespace for all values where this is possible through the UI. See that
   - :heavy_check_mark: location is updated in DB without the leading and trailing whitespace
- :heavy_check_mark: Try to edit an existing location with only whitespace for each value individually for all values where this is possible through the UI. See that
   - :heavy_check_mark: location cannot be updated if the value was required
- :heavy_check_mark: `created_at` and `updated_at` behave as expected when locations are added or updated
   
Note that I'm not bothering to test the values that do not pass the UI validation (for example `responderPhoneNumber`). The trimming should be enforced at the API level, but I don't anticipate us trying to hack our own system by using the API instead of the UI (or the DB directly).

### DIFF
--- a/server/test/unit/vitalsTest/checkHeartbeatTest.js
+++ b/server/test/unit/vitalsTest/checkHeartbeatTest.js
@@ -54,9 +54,11 @@ function locationFactory(overrides = {}) {
     overrides.initialTimer !== undefined ? overrides.initialTimer : 1,
     overrides.isActive !== undefined ? overrides.isActive : true,
     overrides.firmwareStateMachine !== undefined ? overrides.firmwareStateMachine : true,
-    overrides.client !== undefined ? overrides.client : clientFactory(),
     overrides.sirenParticleId !== undefined ? overrides.sirenParticleId : 'fakeSirenParticleId',
     overrides.sentLowBatteryAlertAt !== undefined ? overrides.sentLowBatteryAlertAt : '2021-03-09T19:37:28.176Z',
+    overrides.createdAt !== undefined ? overrides.createdAt : '2021-05-05T19:37:28.176Z',
+    overrides.updatedAt !== undefined ? overrides.updatedAt : '2021-06-07T03:19:30.832Z',
+    overrides.client !== undefined ? overrides.client : clientFactory(),
   )
 }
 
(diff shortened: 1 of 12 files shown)


trisapeace:
Task linked: [CU-1pdmvee Trim values from Dashboard Edit and Create forms before saving in the DB](https://app.clickup.com/t/1pdmvee)

server/test/unit/vitalsTest/checkHeartbeatTest.js
```diff
@@ -54,9 +54,11 @@ function locationFactory(overrides = {}) {
     overrides.initialTimer !== undefined ? overrides.initialTimer : 1,
     overrides.isActive !== undefined ? overrides.isActive : true,
     overrides.firmwareStateMachine !== undefined ? overrides.firmwareStateMachine : true,
-    overrides.client !== undefined ? overrides.client : clientFactory(),
```
trisapeace, line 57: I had this in the wrong spot before. Should be fixed now.

PR #147, opened 2022-01-10
Remove moment.js and upgrade dependencies

trisapeace:
- Wasn't using moment anywhere anymore
- Upgraded based on npm audit
- No longer need to ignore a security vulerability in Travis. It must
  now have a patch in whatever library that was

## Test plan
- :heavy_check_mark: Automated tests pass (since moment was already not being used and the other two upgraded dependencies are only used for tests, this should be sufficient)
- :heavy_check_mark: Run smoke tests on Sensors Dev


server/package.json
```diff
@@ -18,7 +18,7 @@
     "@sentry/node": "^6.2.5",
     "@sentry/tracing": "^6.2.5",
     "axios": "^0.21.2",
-    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v6.2.0",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#td-update-dependencies",
```
trisapeace, line 21: Reminder to self to change this once https://github.com/bravetechnologycoop/brave-alert-lib/pull/33 is merged.

trisapeace: Done.

PR #162, opened 2022-05-12
Attempt #2 to fix the deadlock issue (CU-2atyr1z)

trisapeace:
This is 3-pronged attack:
1. Remove the auto-refresh from the Vitals pages to reduce the number of times the `getRecentVitals*` DB queries are called, thus reducing the likelihood of a deadlock for these functions
2. Add a `sensors_vitals_cache` table in the DB that will keep only the most recent vitals data for each location. This will dramatically speed up the queries used by the Vitals pages, thus reducing the likelihood of a deadlock for these functions.
3. Add the ability to turn on and off DB debug logs that I can use if the deadlocks continue to happen. This will help me identify which functions are running at the time of the deadlock and hopefully track down why they are happening.

With the new, faster query for getting the most recent vitals, it made it possible to join on the `locations` table and display even those locations which had never sent a heartbeat message. So I added that.

Then I couldn't help myself from adding the links to the client and locations pages from the Vitals and Client Vitals pages, because it's I've been wishing that I did this the first time around because it's been a pain to navigate and see the locationid for all these places.

## Test Plan
- :heavy_check_mark: Deploy to Sensors Dev
- :heavy_check_mark: Run smoke tests
   - :heavy_check_mark: With `IS_DB_LOGGING=true`
   - :heavy_check_mark: With `IS_DB_LOGGING=false`
   - :heavy_check_mark: With IS_DB_LOGGING absent from the .env (should behave as though IS_DB_LOGGING=false)
- :heavy_check_mark: Vitals page
   -  :heavy_check_mark: Does not auto-refresh
   - :heavy_check_mark: Shows the most-recent sensors vitals data for each location
   - :heavy_check_mark: Locations which have never sent a heartbeat appear with "Never" for the last seen values (remember to check the hover text) and "unknown" for the door battery low flag
   - :heavy_check_mark: Client name now links to the client summary page
   - :heavy_check_mark: Location name now links to the location sessions page
- :heavy_check_mark: Client Vitals page 
   - :heavy_check_mark: Does not auto-refresh
   - :heavy_check_mark: Shows the most-recent sensors vitals data for each of the given Client's locations
   - :heavy_check_mark: Locations which have never sent a heartbeat appear with "Never" for the last seen values (remember to check the hover text) and "unknown" for the door battery low flag
   - :heavy_check_mark: Location name now links to the location sessions page
- :heavy_check_mark: When a location sends its first heartbeat message (i.e. doesn't have any entries in `sensors_vitals` or `sensors_vitals_cache`
   - :heavy_check_mark: Inserted into the `sensors_vitals` table
   - :heavy_check_mark: Inserted into the `sensors_vitals_cache` table with the same values and the current time as `updated_at`
- :heavy_check_mark: When a location sends a subsequent heartbeat message (i.e. already has entries in `sensors_vitals` and `sensors_vitals_cache`
   - :heavy_check_mark: Inserted into the `sensors_vitals` table
   - :heavy_check_mark: Updated the location's rows in `sensors_vitals_cache` table with the new values and the current time as `updated_at`
- :heavy_check_mark: Make a copy of the production DB and run the DB migration script on it
   - :heavy_check_mark: Puts the most recent vitals for each location from `sensors_vitals` into `sensors_vitals_cache`


trisapeace:
Task linked: [CU-2atyr1z Investigate and fix Sensors deadlocks](https://app.clickup.com/t/2atyr1z)

server/package.json
```diff
@@ -18,7 +18,7 @@
     "@sentry/node": "^6.2.5",
     "@sentry/tracing": "^6.2.5",
     "axios": "^0.21.2",
-    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v6.7.0",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#td-add-db-logging",
```
trisapeace, line 21: Reminder to update this once brave-alert-lib PR is merged

trisapeace: Done

PR #163, opened 2022-05-19
Cleanup to hopefully help stop deadlocking (CU-2atyr1z, CU-289wcj9)

trisapeace:
- Removed Brave Siren to reduce the code base and because it had a call to `beginTransaction` that I wanted to get rid of
- Removed the transaction from `handleHeartbeat` and `/api/door` and instead moved it into just the `sendLowBatteryAlert` function because that is the only part of these that needed to do a read-and-update-the-same-table. By putting this closer together, hopefully that will reduce the likelihood of a conflict.
- Updated the email address to `clientsupport@brave.coop` instead of `contact@brave.coop` while I was in here
- Fixed a typo in the `README`

## Test Plan
- :heavy_check_mark: Deploy to Dev
- :heavy_check_mark: Run smoketests
- :heavy_check_mark: Send Boron heartbeats with and without low battery
- :heavy_check_mark: Send Argon door messages with and without low battery
- :heavy_check_mark: See disconnection message has new email address


trisapeace:
Task linked: [CU-2atyr1z Investigate and fix Sensors deadlocks](https://app.clickup.com/t/2atyr1z)

PR #166, opened 2022-05-31
Added disableReset and enableReset to delay ota when in session (CU-1tdbg9r)

PeterScholtens:
Reference the old pull request: https://github.com/bravetechnologycoop/BraveSensor/pull/164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ the code was deployed.
 
 ### Added
 - Changed door sensor scanning to use filters (CU-2bffw4q). 
+- Delay for the OTA reset, so to not reset while in session (CU-1tdbg9r).
 
 ## [5.3.0] - 2022-05-17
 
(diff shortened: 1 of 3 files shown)


trisapeace (approved):

PeterScholtens:
Testing: Originally was waiting 10mins for door sensor to send a heartbeat, but i found out on friday for some reason my im21 door sensor was not sending one. So upon Theresa's advice downloaded a ble scanner on my phone to check. Then i talked to Sampath and he recommended just using the door open broadcast as a method to pretend there was a heartbeat

PeterScholtens:
Test2: Flashed my firmware onto device and without opening the door, waited 60 secs (9 mins in the real code), which enables the reset. Unmarked my device as development device. The boron then immediately downloaded the update and reset, which is the required behaviour. This test was to see, if the boron would reset if the heartbeat timer was not ever changed.

PeterScholtens:
Test4: Flashed my firmware onto device and then waited 60 secs (9 mins in the real code), so that reset was enabled. Made movement so state machine went to state1, then I unmarked my device as development device and it downloaded the update, but did not reset until I stopped making movement and upon returning to state0, it reseted which is the required behaviour. This test was to see, if the boron would reset while in session, which it did not.

PR #168, opened 2022-06-02
Improved Door Heartbeat Debug Publishes (CU-2chv6je)

BraveKerry:
When debugging publishes are turned on, publish the raw advertising data from the door sensor whenever it sends a heartbeat. 
Testing that is done: 
- When door sensor sends a heartbeat, the particle console receives 1-4 publishes of the door sensor raw data, and that data is of expected values. 
- When publishes are off, door sensor does not publish heartbeats.

### DIFF
--- a/firmware/boron-ins-fsm/src/debugFlags.h
+++ b/firmware/boron-ins-fsm/src/debugFlags.h
@@ -1,0 +1,6 @@
+#ifndef DEBUG_FLAGS_H
+#define DEBUG_FLAGS_H
+
+extern bool stateMachineDebugFlag; 
+
+#endif
--- a/firmware/boron-ins-fsm/src/im21door.cpp
+++ b/firmware/boron-ins-fsm/src/im21door.cpp
@@ -6,6 +6,7 @@
 #include "Particle.h"
 #include "im21door.h"
 #include "flashAddresses.h"
+#include "debugFlags.h"
 
 //define global variables so they are allocated in memory
 //initialize door ID to agreed upon intial value
@@ -69,7 +70,7 @@ doorData checkIM21(){
     // Checks if the 2nd bit (counting from 0) of doorStatus is 1
     // read as: doorLowBatteryFlag is true if doorStatus AND 0b0100 is not 0b0000
     doorLowBatteryFlag = (currentDoorData.doorStatus & (1 << 2)) != 0;
-
+    
     // Checks if the 3rd bit of doorStatus is 1
     if ((currentDoorData.doorStatus & (1 << 3)) != 0){
       doorHeartbeatReceived = millis();
@@ -162,13 +163,32 @@ void threadBLEScanner(void *param) {
     
     //loop over all devices found in the BLE scan
     for (BleScanResult scanResult : scanResults) {
-
       //place advertising data in doorAdvertisingData buffer array
+      // Door heartbeat message: 
+      // dooradvertisingdata[0] (1 byte): Firmware Version
+      // dooradvertisingdata[1-3] (3 bytes): Last 3 bytes of door sensor address
+      // dooradvertisingdata[4] (1 byte): Type ID (door sensor, thermometer, smoke detector, etc.)
+      // dooradvertisingdata[5] (1 byte): Event Data
+        // bit[3]: heartbeat?
+        // bit[2]: low battery?
+        // bit[1]: door open?
+        // bit[0]: tamper alarm?
+      // dooradvertisingdata[6] (1 byte): Control Data
+      // More info: https://drive.google.com/file/d/1ZbnHi7uA_xMWVIiMbQjZlbT3OOoykbr4/view?usp=sharing        
       scanResult.advertisingData().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, doorAdvertisingData, BLE_MAX_ADV_DATA_LEN);
       
       //extract door status and dump it in the queue
       scanThreadDoorData.doorStatus = doorAdvertisingData[5];
       scanThreadDoorData.controlByte = doorAdvertisingData[6];
+
+      if ((scanThreadDoorData.doorStatus & (1 << 3)) != 0 && stateMachineDebugFlag) { //
+        //from particle docs, max length of publish is 622 chars, I am assuming this includes null char
+        char debugMessage[622] = "";
+        for (int i = 0; i < BLE_MAX_ADV_DATA_LEN; i++) {
+          snprintf(debugMessage + strlen(debugMessage), sizeof(debugMessage), "%02X ", doorAdvertisingData[i]); 
+        }
+        Particle.publish("Door Heartbeat Received", debugMessage, PRIVATE); 
+      }     
       os_queue_put(bleQueue, (void *)&scanThreadDoorData, 0, 0); 
     }//endfor
 
--- a/firmware/boron-ins-fsm/src/stateMachine.cpp
+++ b/firmware/boron-ins-fsm/src/stateMachine.cpp
@@ -8,6 +8,7 @@
 #include "im21door.h"
 #include "ins3331.h"
 #include "flashAddresses.h"
+#include "debugFlags.h"
 #include <queue>
 
 //define and initialize state machine pointer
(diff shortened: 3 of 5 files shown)


trisapeace:
Task linked: [CU-2chv6je Optimize debug logging of door sensor data](https://app.clickup.com/t/2chv6je)

firmware/boron-ins-fsm/src/im21door.cpp
```diff
@@ -174,6 +174,25 @@ void threadBLEScanner(void *param) {
...
+        // dooradvertisingdata[0] (1 byte): Firmware Version
+        // dooradvertisingdata[1-3] (3 bytes): Last 3 bytes of door sensor address
+        // dooradvertisingdata[4] (1 byte): Type ID (door sensor, thermometer, smoke detector, etc.)
+        // dooradvertisingdata[5] (1 byte): Event Data
+          // bit[3]: heartbeat?
+          // bit[2]: low battery?
+          // bit[1]: door open?
+          // bit[0]: tamper alarm?
+        // dooradvertisingdata[6] (1 byte): Control Data
+        // More info: https://drive.google.com/file/d/1ZbnHi7uA_xMWVIiMbQjZlbT3OOoykbr4/view?usp=sharing
```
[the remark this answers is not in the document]
trisapeace: What do you think about moving this description up to where the incoming data is first stored in `dooradvertisingdata`?

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -8,6 +8,7 @@
 #include "im21door.h"
 #include "ins3331.h"
 #include "flashAddresses.h"
+#include "debugFlags.h"
```
trisapeace, line 11: Since you've included this here, should `bool stateMachineDebugFlag = false;` be removed from this file and maybe put it its own `.cpp` file? Or is that unnecessary?

firmware/boron-ins-fsm/src/im21door.cpp
```diff
@@ -174,6 +174,25 @@ void threadBLEScanner(void *param) {
...
+        // dooradvertisingdata[4] (1 byte): Type ID (door sensor, thermometer, smoke detector, etc.)
+        // dooradvertisingdata[5] (1 byte): Event Data
+          // bit[3]: heartbeat?
+          // bit[2]: low battery?
+          // bit[1]: door open?
+          // bit[0]: tamper alarm?
+        // dooradvertisingdata[6] (1 byte): Control Data
+        // More info: https://drive.google.com/file/d/1ZbnHi7uA_xMWVIiMbQjZlbT3OOoykbr4/view?usp=sharing        
+
+        if (stateMachineDebugFlag) { //(doorAdvertisingData[5] & (1 << 3)) != 0 &&
```
trisapeace, line 189: Should this be uncommented so that the debug message is only sent after a heartbeat message? `//(doorAdvertisingData[5] & (1 << 3)) != 0 &&` 

Is this going to send three times if the message arrives on all three channels?

firmware/boron-ins-fsm/src/im21door.cpp
```diff
@@ -174,6 +174,25 @@ void threadBLEScanner(void *param) {
...
+          char debugMessage[622] = "";
```
trisapeace, line 190: Where does the 622 come from?

firmware/boron-ins-fsm/src/debugFlags.h
```diff
@@ -0,0 +1,6 @@
+#ifndef DEBUG_FLAGS_H
```
trisapeace, line 1: Can you please add a line in the CHANGELOG?
And in the description or a comment in this PR, describe the testing that you've done to feel confident that this works.

trisapeace (commented):
Looks good. Just a couple of questions that I've put in the comments.

firmware/boron-ins-fsm/src/im21door.cpp
```diff
@@ -69,7 +70,6 @@ doorData checkIM21(){
     // Checks if the 2nd bit (counting from 0) of doorStatus is 1
     // read as: doorLowBatteryFlag is true if doorStatus AND 0b0100 is not 0b0000
     doorLowBatteryFlag = (currentDoorData.doorStatus & (1 << 2)) != 0;
-
```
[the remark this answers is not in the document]
BraveKerry: I like the idea of having all functions fit on one screen. So I 'automatically' remove newlines if I feel like a pre-existing comment(s) separate the two sections just fine. I can add it back if you'd like.

firmware/boron-ins-fsm/src/im21door.cpp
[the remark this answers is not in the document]
BraveKerry: There's also another comment on the very top of [im21door.h](https://github.com/bravetechnologycoop/BraveSensor/blob/main/firmware/boron-ins-fsm/src/im21door.h) that explains what `doorAdvertisingData[5]`/`doorStatus` means. Should that also be relocated?

firmware/boron-ins-fsm/src/stateMachine.cpp
BraveKerry: I feel like given the naming of the variable and the structure of our includes, `stateMachine.cpp` feels like the parent of the sensors. Therefore, initializing it there should be fine (unless we move it to the `.ino`, but it doesn't get used there).

firmware/boron-ins-fsm/src/im21door.cpp
BraveKerry: Yes I should make the change to that it only sends for heartbeat messages. 
The publishes will happen 1-4 for each of the channels, I can avoid that by moving it so that it is processed in `checkIM21`, which bypasses the os_queue. However at that point, it's pretty much the same data as the door data seen in the normal debug messages.

firmware/boron-ins-fsm/src/im21door.cpp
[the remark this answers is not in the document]
trisapeace: So funny. Different people's version of readability. I find the extra lines help me not become bogged down in the text, but I can also understand the advantage of fitting everything on one screen.

In general, I'd say try to stick with the conventions of the codebase that you're in. So if there are newlines, then they should stay. And if there aren't newlines, then they shouldn't be added. It also makes reviewing the diff easier because then less stuff changes. Thanks!

firmware/boron-ins-fsm/src/im21door.cpp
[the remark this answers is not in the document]
trisapeace: Hmmm, maybe it's not even necessary to keep the one in im21door.h after the comment that you've added. Put yourself in the shoes of someone new to the codebase and trying to understand. Where would you look for it? Where would it be most helpful?

firmware/boron-ins-fsm/src/stateMachine.cpp
trisapeace: Sounds good to me. Thanks!

firmware/boron-ins-fsm/src/im21door.cpp
trisapeace: No, you're right, nicer to have it here. And since we aren't going to always be in debug mode, maybe it makes sense to have it published literally every time. I think I'm OK with it as it is.

firmware/boron-ins-fsm/src/im21door.cpp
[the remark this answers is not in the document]
BraveKerry: Gotcha, I'll remove this change.

firmware/boron-ins-fsm/src/im21door.cpp
BraveKerry: By "every time", do you mean every door heartbeat or every door broadcast?

firmware/boron-ins-fsm/src/im21door.cpp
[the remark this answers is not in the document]
BraveKerry: I noticed that the comment is also very similar to the constant declarations, therefore I agree that it is not necessary to keep that one.

PeterScholtens (approved):
lgtm! 🔥🔥

firmware/boron-ins-fsm/src/im21door.cpp
trisapeace: Haha, ambiguous, I'm not even sure which of those two is the one that I want.

What I mean is to not do anything fancy to deduplicate messages that we get more than once on different channels.

BraveKerry: I see. I'll go with the former then.

trisapeace (approved):
Fantastic. Can't wait to see if this reveals anything interesting.

PR #171, opened 2022-06-16
Door Sensor Address Filter Miss (CU-2f1x56y)

BraveKerry:
Updated to include CASEY_H_1's door sensor address in the filter. 
Testing: 
- Firmware still works with the previous hard-coded door sensor address door sensors (i.e. the one I have at home).


trisapeace:
Task linked: [CU-2f1x56y Door Sensor Address Filter Miss](https://app.clickup.com/t/2f1x56y)

BraveKerry:
Unfortunately there is no wildcard feature in the filter. However, depending on whether or not we run into more door sensors with unusual addresses, I can reimplement the door-sensor-last-three-bytes check. 
Yes, we would have to do that during provisioning. 
I have updated the changelog.

trisapeace (approved):
Awesome. Thanks! Go ahead and merge.

Can you please document the step about checking the door sensor values in whatever documentation there is for provisioning Sensors?

PR #172, opened 2022-06-23
Console function to force Reset (CU-2e5adgd)

PeterScholtens:


trisapeace:
Task linked: [CU-2e5adgd Console function to override disable Reset](https://app.clickup.com/t/2e5adgd)

PeterScholtens:
Test1:  Flashed the boron with new firmware, and tried the console function with invalid input such as 0 or any number except for 1, nothing happened except -1 was returned. Entered '1' and received a message before the boron reset. 
Result: Test passed

PeterScholtens:
Test2: Flashed the boron with new firmware and the using the particle console flashed to an old firmware. Like i thought the firmware downloaded but did not reset as it was not pass the DEVICE_RESET_THRESHOLD. Passed a '1' into the force_reset() console function and it reset to the old firmware.
Result: Test passed

PeterScholtens:
Test3: I wrote in the consoleFunctionsTests.cpp
Result: Test passed

firmware/README.md
```diff
@@ -389,6 +390,21 @@ The IM21 door sensors each have a sticker on them with their door IDs. On the bo
 - 1 - if door ID echoed to the cloud
 - -1 - if bad input was received and door ID was neither parsed or echoed to the cloud
 
+### **force_reset(String)**
```
trisapeace, line 393: Thanks for noticing this and updating the docs!

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -21,13 +21,14 @@
 void setupConsoleFunctions(){
 
   //particle console function declarations, belongs in setup() as per docs
+  Particle.function("Force_Reset", force_reset); 
+  Particle.function("Turn_Debugging_Publishes_On_Off", toggle_debugging_publishes);   
   Particle.function("Change_Initial_Timer", initial_timer_set); 
   Particle.function("Change_Duration_Timer", duration_timer_set);     
   Particle.function("Change_Stillness_Timer", stillness_timer_set);  
   Particle.function("Change_INS_Threshold", ins_threshold_set);
-  Particle.function("Turn_Debugging_Publishes_On_Off", toggle_debugging_publishes);
```
trisapeace, line 28: Any particular reason why this line moved?

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -21,13 +21,14 @@
...
+  Particle.function("Force_Reset", force_reset); 
+  Particle.function("Turn_Debugging_Publishes_On_Off", toggle_debugging_publishes);   
   Particle.function("Change_Initial_Timer", initial_timer_set); 
   Particle.function("Change_Duration_Timer", duration_timer_set);     
   Particle.function("Change_Stillness_Timer", stillness_timer_set);  
   Particle.function("Change_INS_Threshold", ins_threshold_set);
-  Particle.function("Turn_Debugging_Publishes_On_Off", toggle_debugging_publishes);   
   Particle.function("Change_IM21_Door_ID", im21_door_id_set); 
-
+
```
trisapeace, line 31: Oops, looks like you accidentally added some whitespace here. Could you please remove it?

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -268,4 +269,32 @@ int im21_door_id_set(String command) { // command is a long string with all the
...
+  //entering 0 vs entering bad input. So convert to char and use ascii table
+  const char* holder = command.c_str();
+
+  if(*(holder+1) != 0){
+    //any string longer than 1 char is invalid input, so
+    returnFlag = -1;
+  }
+  else if(*holder == '1'){
+    returnFlag = 1;
+    bool msg_sent = Particle.publish("YOU SHALL NOT PANIC!!", "Reset has begun so ignore the future particle message about failure to call force_reset()", PRIVATE | WITH_ACK);
```
trisapeace, line 289: Love it.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -268,4 +269,32 @@ int im21_door_id_set(String command) { // command is a long string with all the
...
+    if (msg_sent)
+      System.reset();
```
trisapeace, line 291: I am personally of the opinion that in C-derived languages, one should put curly braces around the bodies of conditionals and loops. I have two main reasons:
1. If you want to add something to the body of the loop/conditional, then you might forget to add the braces and then your code will be wrong even if the indentation makes it look correct. So difficult to debug
2. If you want to add something to the body of the loop/conditional and you do remember to add the curly braces, then in my preferred style that would add the opening brace on the same line as the if/while/for/etc which would make it show up in the git diff when really nothing there changed.

That being said, I don't change the firmware enough to know what the typical style here is. So I would defer to what is normal in the rest of the code.

firmware/boron-ins-fsm/test/mocks/System.h
```diff
@@ -0,0 +1,17 @@
...
+public:
+    MockSystem()
+    {
+    }
+
+public:
+    bool reset() const
+    {
+        printf("System Reset Works");
+        return true;
```
trisapeace, line 13: It would be really nice if there was some way that we could tell that this function was being called or not. 

Is it possible to add an extern variable in this file for `resetWasCalled`? 
Then
1. Before each test, you can set it to false
2. In this mocked function you will set it to true
3. Then in the test you can check that it is now true, thus showing that the function was, in fact, called

I don't think we're doing this anywhere else yet, but it feels like it should work. Could you please try it out? Let me know if my description isn't clear enough!

trisapeace: If this works, it would be amazing if you could extend the `mocks/Particle.h` file's `publish` function to do something similar, but instead to put the whole string into the `extern` variable so it can be checked no only that it was called but that it was called with the expected parameters.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ the code was deployed.
 
 ### Added
 - Boron Sensor heartbeat optimizations, to improve debugging and door connectivity (CU-24qak6k).
+- Console function to enable developers to force reset and bypass DEVICE_RESET_THRESHOLD (CU-2e5adgd).
```
trisapeace, line 16: Thanks for remembering.

Before this is merged, could you please squash your commits and rebase off `main`? Thanks!

trisapeace (commented):
Looks great. Just some testing improvements that it would be great if you could do. Thanks!

firmware/README.md
PeterScholtens: No worries

firmware/boron-ins-fsm/src/consoleFunctions.cpp
PeterScholtens: For sure

CHANGELOG.md
PeterScholtens: Im not sure how do to that. Usually i rebase and merge.....?

trisapeace: The way I normally squash commits together is using the interactive rebase feature in git `git rebase -i HEAD~10` where 10 is the number of commits back in time that you now is more than you're going to need to modify. 

It will display all those 10 (or however many you specify) commits in the text editor with the word `pick` beside each one and a big set of instructions of what words you can change `pick` to in order to do other actions with it. 

Your most recent commit will be at the bottom. If you want to squash it with the commit directly above it in the list, you replace `pick` with `s` (or `squash`).

So in your case, because this whole PR should be a single commit, you can squash all the way back to your first commit (which will still say `pick`). 

Then you save the text file with all the `squash` and `pick` values and it will show you a combined version of the commit message to edit. Edit it down to what you'd like the commit message to be for this one new amazing commit and save it. 

Then if you look at your history using `git log`, you'll only see the one commit with the new message.

This will give us a nice, clean git history when it's committed into `main`. It will also make during the rebase from `main` easier because you will only have to deal with any conflicts once instead of a whole chain of them, once for each commit on your branch.

Like any rebase operation, this changes your git history, so when you next push to GitHub, you'll need to force push (`git push -f`). But don't do this until you're sure that you've got it the way you want it.

If you mess up at any point, try `rebase --abort` or just delete your branch locally and do a fresh `git pull` from GitHub of your branch and try again.

If you'd like to walk through it with me, we can do that too!

firmware/boron-ins-fsm/src/consoleFunctions.cpp
PeterScholtens: Well I figured that since "Turn_Debugging_Publishes_On_Off" is used the most on the particle console, it would be nicer to just have it at the top of the list. I also inserted the force reset at the top, so it is just under the force update. I can move it back if you do not like it.
Screen Shot 2022-06-27 at 10 34 46 AM

firmware/boron-ins-fsm/test/mocks/System.h
PeterScholtens: It worked for the reset, so ill do it for the publish as well

CHANGELOG.md
PeterScholtens: Oh ok, cool. Yes if you could walk me through it that would be great! Thank you!

firmware/boron-ins-fsm/src/consoleFunctions.cpp
trisapeace: Haha, no, that's great. I didn't realize that the order in the code changed the order in Particle. This makes sense to me. Thanks!

trisapeace (approved):
Looks great! Go ahead and merge once the tests pass.

PeterScholtens:
All tests are passing

PR #176, opened 2022-07-14
Add Country, Country Subdivision, and Building Type to CSV Export (CU-2c6crcn)

trisapeace:
- Used an extension table because this data is only every used by
  the CSV export and I didn't want to have to alter the Client
  object to add fields that aren't used anywhere else in the
  program. This also gives the flexibility to easily add and remove
  columns in this table for this need as more reporting requirements
  come up or change
- For now, it is an MVP and is not included
  on the New Client or Update Client pages. If it is missing when
  someone does the CSV export and they need it filled in, they
  can ask someone to add it to the DB
- Just used the `text` datatype to keep this super flexible for now. If we find that it's a problem later, we can introduce an `enum` type.

## Test Plan
- :heavy_check_mark: Deploy to Dev
- :heavy_check_mark: Run smoke tests
- :heavy_check_mark: Export CSV
   - :heavy_check_mark: Old columns still have the correct data
   - :heavy_check_mark: Country column exists and has the correct values
   - :heavy_check_mark: Country Subdivision column exists and has the correct values
   - :heavy_check_mark: Building Type column exists and has the correct values
- :heavy_check_mark: Add country, country subdivision, and/or building type to some and then all of clients to see that it is allowed to be blank and that all the values appear in the exported CSV
- :heavy_check_mark: Create a new Client
   - :heavy_check_mark: Empty row in `clients_extension` is added
- :heavy_check_mark: Update an existing Client that has an entry in the `client_extension` table
- :heavy_check_mark: Update an existing Client that does not have an entry in the `client_extension` table

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ the code was deployed.
 
 - Console function to enable developers to force reset and bypass DEVICE_RESET_THRESHOLD (CU-2e5adgd).
 - The ability to use different language chatbot messages (CU-2dtutrx).
+- Country, Country Subdivision, and Building Type columns to the CSV Export (CU-2c6crcn).
 
 ## [6.1.0] - 2022-06-27
 
--- a/server/dashboard.js
+++ b/server/dashboard.js
@@ -183,6 +183,9 @@ async function downloadCsv(req, res) {
     'Session State',
     'Alert Type',
     'Session Responded By',
+    'Country',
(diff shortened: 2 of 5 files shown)


trisapeace:
Task linked: [CU-2c6crcn Add client region (country and state/province) and building type to Buttons and Sensor DB and CSV export](https://app.clickup.com/t/2c6crcn)

PR #183, opened 2022-09-12
Added string_convert namespace for string_convert.cpp functions

kerryzhang-brave-coop:
Fixed an issue with string_convert function ultoa conflicting with stdlib function

Deleted commented out function to clean up code

Testing: Code compiles and unit tests pass

### DIFF
--- a/firmware/boron-ins-fsm/inc/spark_wiring_string.cpp
+++ b/firmware/boron-ins-fsm/inc/spark_wiring_string.cpp
@@ -59,11 +59,11 @@ void dtoa (double val, unsigned char prec, char *sout) {
     unsigned long first = (unsigned long)(fixed / scale);
     unsigned long second = (unsigned long)(fixed % scale);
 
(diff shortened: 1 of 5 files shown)


firmware/boron-ins-fsm/inc/spark_wiring_string.cpp
```diff
@@ -116,39 +116,39 @@ String::String(unsigned char value, unsigned char base)
 {
 	init();
 	char buf[9];
-	utoa(value, buf, base);
+	string_convert::utoa(value, buf, base);
```
trisapeace, line 119: Oh, do namespaces usually use underscores or camelcase?

kerryzhang-brave-coop: Discussed in call and agreed naming scheme should be fine

trisapeace (approved):

PR #186, opened 2022-10-24
Change the "Change_IM21_Door_ID" Particle cloud function echo functionality to return the door ID (CU-3455bc0)

kerryzhang-brave-coop:
These changes change the im21_door_id_set() function in consoleFunctions.cpp to always return the current door ID as a decimal integer by pulling the door ID from EEPROM memory and then converting it to a decimal integer.

The mock EEPROM located in eeprom.h has been modified to always return 0x123456 when the address parameter is set to get the value of the door ID.

Modified Change_IM21_Door_ID unit tests to work correctly with the new return value. 

The changes were tested on a Boron by changing the door ID using the Particle console function to an actual door sensor ID, and observing that the return value was the door ID converted to a decimal number. It was confirmed that the door sensor still communicated with the Boron as expected. It was confirmed that sending "e" to the change_im21_door_id command correctly returned the current door ID converted to a decimal number. Then, the Boron was restarted and all steps repeated, confirming that the results were the same. 

It was also tested that the changes do not affect existing door IDs stored on Borons which were set using older versions of the console command. A Boron was flashed with firmware v60401 and the door ID was set using the Particle function to an actual door ID. It was confirmed that the door sensor still communicated with the Boron as expected. Then, the same Boron was flashed with firmware built from these changes, restarted, and it was confirmed that the door sensor still communicated with the Boron as expected.


trisapeace:
Task linked: [CU-3455bc0 Change the "Change_IM21_Door_ID" Particle cloud function echo functionality to return the door ID](https://app.clickup.com/t/3455bc0)

firmware/boron-ins-fsm/test/mocks/eeprom.h
```diff
@@ -1,5 +1,17 @@
+/*
```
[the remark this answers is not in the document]
trisapeace: I'm interested in your thoughts on this as well.

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -91,24 +91,27 @@ SCENARIO( "Change_IM21_Door_ID", "[change door id]" ) {
...
+        globalDoorID.byte3 = (uint8_t)strtol("AA",NULL,16);
 
-        WHEN( "the function is called with e" ) {
+        WHEN( "The function is called with e and a valid door ID was previously set" ) {
+            im21_door_id_set("00,10,00");
             int returnVal = im21_door_id_set("e");
 
-            THEN( "should return 1" ) {
-                REQUIRE( returnVal == 1);
+            THEN( "The function should return the door ID that was previously set" ) {
```
trisapeace, line 101: Looks like this test is failing on Travis.

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -91,24 +91,27 @@ SCENARIO( "Change_IM21_Door_ID", "[change door id]" ) {
...
+                REQUIRE( returnVal == 4096 );
             }
         }
 
         WHEN( "the function is called with a valid door ID" ) {
             int returnVal = im21_door_id_set("EF,CD,AB");
 
-            THEN( "should return 1" ) {
-                REQUIRE( returnVal == 1);
+            THEN( "The function should return the door ID converted to a decimal number" ) {
```
trisapeace, line 112: Looks like this test is failing on Travis.

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -295,7 +301,8 @@ SCENARIO( "Set Stillness Timer", "[stillness timer]" ) {
 SCENARIO( "Set INS Threshold", "[ins threshold]" ) {
 
     GIVEN( "A starting initial threshold of 10" ) {
-            ins_threshold = 10;
+        ins_threshold = 10;
+        ins_threshold_set("10");
 
         WHEN( "the function is called with 'e'" ) {
```
trisapeace, line 307: Looks like this test is failing on Travis.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -260,14 +260,23 @@ int im21_door_id_set(String command) { // command is a long string with all the
     globalDoorID.byte1 = (uint8_t)strtol(byteholder3,NULL,16);
 
     //write new global door ID to flash
-    EEPROM.put(ADDR_IM21_DOORID,globalDoorID.byte1);  
-    EEPROM.put((ADDR_IM21_DOORID+1),globalDoorID.byte2);  
-    EEPROM.put((ADDR_IM21_DOORID+2),globalDoorID.byte3);  
+    EEPROM.put(ADDR_IM21_DOORID, globalDoorID.byte1);  
+    EEPROM.put((ADDR_IM21_DOORID+1), globalDoorID.byte2);  
+    EEPROM.put((ADDR_IM21_DOORID+2), globalDoorID.byte3);
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: No problem

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -260,14 +260,23 @@ int im21_door_id_set(String command) { // command is a long string with all the
...
+    EEPROM.put((ADDR_IM21_DOORID+2), globalDoorID.byte3);  
   
   } //end if-else
 
-  return 1;
+  // convert door ID to decimal for return value
+  // get the door ID
+  EEPROM.get(ADDR_IM21_DOORID, doorIDHolder.byte1);
+  EEPROM.get((ADDR_IM21_DOORID + 1), doorIDHolder.byte2);
+  EEPROM.get((ADDR_IM21_DOORID + 2), doorIDHolder.byte3);
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: Good point, I modified the code to get the return value from local variables instead of calling EEPROM and additional time

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -91,24 +91,27 @@ SCENARIO( "Change_IM21_Door_ID", "[change door id]" ) {
 
     GIVEN( "A default global door id" ) {
         globalDoorID.byte1 = (uint8_t)strtol("AA",NULL,16);
-        globalDoorID.byte1 = (uint8_t)strtol("AA",NULL,16);
-        globalDoorID.byte1 = (uint8_t)strtol("AA",NULL,16);
+        globalDoorID.byte2 = (uint8_t)strtol("AA",NULL,16);
+        globalDoorID.byte3 = (uint8_t)strtol("AA",NULL,16);
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: Thanks!

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -91,24 +91,27 @@ SCENARIO( "Change_IM21_Door_ID", "[change door id]" ) {
...
 
-        WHEN( "the function is called with e" ) {
+        WHEN( "The function is called with e and a valid door ID was previously set" ) {
+            im21_door_id_set("00,10,00");
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: Good idea, I'll choose a different door ID

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
kerryzhang-brave-coop: This is strange, because the tests all pass on my machine, I'll look into what may be causing this. Same for the test cases below

firmware/boron-ins-fsm/test/mocks/eeprom.h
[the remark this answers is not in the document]
kerryzhang-brave-coop: Hmm, I read the page you linked as well as a few other results on google. What do you think are the main downsides of using a fake for this? The pages I looked at mentioned that both mocks and fakes are used for unit tests, and several also used fake memory as an example of a case where you might use a fake. 

I agree that making a mock that sets return values for EEPROM functions would look similar. Do you think having the fake for other existing and future uses would be less reliable than a mock? i.e. it will catch less errors, etc.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -221,56 +221,63 @@ int ins_threshold_set(String input){
...
-    EEPROM.put((ADDR_IM_DOORID+2),globalDoorID.byte3);  
-  
-  } //end if-else
+    EEPROM.put(ADDR_IM_DOORID, globalDoorID.byte1);
+    EEPROM.put((ADDR_IM_DOORID + 1), globalDoorID.byte2);
+    EEPROM.put((ADDR_IM_DOORID + 2), globalDoorID.byte3);
 
-  return 1;
+    // put door ID in buffer for return value
+    snprintf(buffer, sizeof(buffer), "%02X%02X%02X", globalDoorID.byte1, globalDoorID.byte2, globalDoorID.byte3);
```
trisapeace, line 275: Here and for the `e` echo return value, I was actually expecting this to return with the bytes in the opposite order:
`snprintf(buffer, sizeof(buffer), "%02X%02X%02X", globalDoorID.byte3, globalDoorID.byte2, globalDoorID.byte1);`. This way it will match the order on the door sensor sticker itself, right?

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -109,25 +108,25 @@ SCENARIO( "Turn_Debugging_Publishes_On_Off", "[toggle debug flag]" ) {
...
+            im21_door_id_set("56,34,12");
             int returnVal = im21_door_id_set("e");
 
-            THEN( "should return 1" ) {
-                REQUIRE( returnVal == 1);
+            THEN( "The function should return the door ID that was previously set" ) {
+                // There exists a bug where the last two bytes of returnVal are set
+                // to 00 inside this THEN macro, even if the value was previously
+                // correct. 
+                REQUIRE( returnVal == 1193046 );
```
trisapeace, line 120: I was expecting this to return `5649426` which is 0x563412.

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -151,6 +150,7 @@ SCENARIO( "Change_IM21_Door_ID", "[change door id]" ) {
 SCENARIO( "Set Initial Timer", "[initial timer]" ) {
     GIVEN( "A starting initial timer of 10 milliseconds" ) {
         state1_max_time = 10000;
+        initial_timer_set("10");
```
trisapeace, line 153: Can you please explain to me what is happening here?

I'm not sure I understand the relationship between `state1_max_time` and `initial_timer_set`. Do both of these need to be here?

firmware/boron-ins-fsm/test/mocks/mock_eeprom.h
```diff
@@ -1,21 +1,45 @@
...
 
     template <typename T>
-    void put(int const _address, T const& _data)
-    {
+    void put(int const _address, T const& _data) {
+        if (_address >= BORON_EEPROM_SIZE) {
+            return;
+        }
+
+        memcpy(mockMemory + (_address * BYTES_PER_ELEMENT), &_data, sizeof(_data));
```
trisapeace, line 42: I still think that this is too complicated for what we want to do here. Just the fact that you ran into issues with the size of memory shows that it is fragile and that the test can fail for a reason outside the code that it's supposed to be testing.

I would prefer that the `get` be implemented to always return fixed values. It can be different for each address (if that would make the test better) by putting a if/else inside that will return different values depending on what `_address` is.

And the PUT doesn't have to do anything. I think you'll still be able to do your two tests even with this simplified mock because `e` will return a value that you can use and the positive int is processed in variables to form the return value, it isn't re-retrieved from EEPROM. What do you think?

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ the code was deployed.
 
 - Production deployment instructions in the README.
 - Created string_convert namespace for string_convert.cpp functions.
+- Modified im21_door_id_set function to return the current door ID, converted to a decimal number (CU-3455bc0).
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: I moved this in my latest commit, thanks for the reminder

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -109,25 +108,25 @@ SCENARIO( "Turn_Debugging_Publishes_On_Off", "[toggle debug flag]" ) {
...
-        WHEN( "the function is called with e" ) {
+        WHEN( "The function is called with e and a valid door ID was previously set" ) {
+            im21_door_id_set("56,34,12");
             int returnVal = im21_door_id_set("e");
 
-            THEN( "should return 1" ) {
-                REQUIRE( returnVal == 1);
+            THEN( "The function should return the door ID that was previously set" ) {
+                // There exists a bug where the last two bytes of returnVal are set
+                // to 00 inside this THEN macro, even if the value was previously
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: I removed the comment in my latest commit

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -109,25 +108,25 @@ SCENARIO( "Turn_Debugging_Publishes_On_Off", "[toggle debug flag]" ) {
 SCENARIO( "Change_IM21_Door_ID", "[change door id]" ) {
 
     GIVEN( "A default global door id" ) {
-        globalDoorID.byte1 = (uint8_t)strtol("AA",NULL,16);
-        globalDoorID.byte1 = (uint8_t)strtol("AA",NULL,16);
-        globalDoorID.byte1 = (uint8_t)strtol("AA",NULL,16);
 
-        WHEN( "the function is called with e" ) {
+        WHEN( "The function is called with e and a valid door ID was previously set" ) {
+            im21_door_id_set("56,34,12");
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: This is a good point. I changed the GIVEN to instead be "a valid global door ID"

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -221,56 +221,63 @@ int ins_threshold_set(String input){
 
 }
 
-
-
-//particle console function to get/set door sensor ID
```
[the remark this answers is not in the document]
kerryzhang-brave-coop: Good catch, I didn't know about this section, I've changed it now to include the changes

firmware/boron-ins-fsm/src/consoleFunctions.cpp
kerryzhang-brave-coop: This is completely true, and I've changed it in my latest commit. I also renamed doorIDHolder to use the same byte order as globalDoorID as I think that is what caused the confusion

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
kerryzhang-brave-coop: After the fix in the previous comment, I also reversed the door ID that was being set to 12,34,56, making this the correct value to check for

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
kerryzhang-brave-coop: This is true, after the original EEPROM changes I made, only the initial_timer_set needed to be included. Since I am no longer using the mock EEPROM to store and get actual values, I removed the initial_timer_set for this test, as well as all of the following tests.

firmware/boron-ins-fsm/test/mocks/mock_eeprom.h
kerryzhang-brave-coop: This works for me. I've made the changes in my latest commit.

firmware/boron-ins-fsm/test/mocks/mock_eeprom.h
```diff
@@ -1,21 +1,39 @@
...
+ * Mock EEPROM memory for Particle Boron
+ * Reference: https://docs.particle.io/reference/device-os/api/eeprom/eeprom/
+ * 
+ * NOTE: Due to an issue where sizeof() returns different values on different 
+ * operating systems for certain data types, this mock EEPROM implementation
+ * differs from the actual Boron EEPROM by always allocating BYTES_PER_ELEMENT
+ * bytes for each element being stored. In reality, the EEPROM only has 
+ * BORON_EEPROM_SIZE bytes of memory. When data is read from or written to the
+ * address pointed to by _address, sizeof(data) is used to read or write a
+ * number of bytes equal to the size of the data type.
```
trisapeace, line 11: I think this can be removed now with the new implementation, right?

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -221,56 +221,63 @@ int ins_threshold_set(String input){
...
 
     //write new global door ID to flash
-    EEPROM.put(ADDR_IM_DOORID,globalDoorID.byte1);  
-    EEPROM.put((ADDR_IM_DOORID+1),globalDoorID.byte2);  
-    EEPROM.put((ADDR_IM_DOORID+2),globalDoorID.byte3);  
-  
-  } //end if-else
+    EEPROM.put(ADDR_IM_DOORID, globalDoorID.byte3);
+    EEPROM.put((ADDR_IM_DOORID + 1), globalDoorID.byte2);
+    EEPROM.put((ADDR_IM_DOORID + 2), globalDoorID.byte1);
```
trisapeace, line 272: I'm concerned that this part is now backwards.

Given a door ID string "AB,CD,EF":
`split1`, `byteholder3`, and `globalDoorID.byte3` look like they will be `AB`.
`split2`, `byteholder2`, and `globalDoorID.byte2` look like they will be `CD`.
and
`split3`, `byteholder1`, and `globalDoorID.byte1` look like they will be `EF`.

which means I would have expected
ADDR_IM_DOORID = `EF` = `globalDoorID.byte1`
ADDR_IM_DOORID = `CD` = `globalDoorID.byte2`
ADDR_IM_DOORID = `AB` = `globalDoorID.byte3`

Am I reading it wrong? Have you tried this on a real Boron?

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -221,56 +221,63 @@ int ins_threshold_set(String input){
...
-    EEPROM.get((ADDR_IM_DOORID+1),holder.byte2);  
-    EEPROM.get((ADDR_IM_DOORID+2),holder.byte3);  
+  // if echo, publish current door ID
+  if(*checkForEcho == 'e'){  
+    EEPROM.get(ADDR_IM_DOORID, doorIDHolder.byte3);  
+    EEPROM.get((ADDR_IM_DOORID + 1), doorIDHolder.byte2);  
+    EEPROM.get((ADDR_IM_DOORID + 2), doorIDHolder.byte1);  
 
-    char buffer[64];
     snprintf(buffer, sizeof(buffer), "{\"byte1\":\"%02X\", \"byte2\":\"%02X\", \"byte3\":\"%02X\"}",
```
trisapeace, line 243: I've always found this output confusing. What do you think about changing the returned JSON to `{"doorId": "AB,CD,EF"}` where the string is in the same order as when it's given the the `im21_door_id_set` command?

firmware/boron-ins-fsm/test/mocks/mock_eeprom.h
kerryzhang-brave-coop: Thanks for reminding me, I removed it and replaced it with an updated description

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -217,6 +217,63 @@ int ins_threshold_set(String input) {
 }
 
 // particle console function to get/set door sensor ID
+<<<<<<< HEAD
```
trisapeace, line 220: Oops, looks like some merging stuff is happening here.

PR #187, opened 2022-10-25
Update Sensor firmware to work with IM24 door sensors (CU-3644bwf)

kerryzhang-brave-coop:
We want to start using IM24 door sensors because IM21s are no longer available. Add an additional address to the BLE Scan Filter with the first 6 characters of IM24 door IDs hard coded in, allowing for the last 6 characters to be inputted using the Particle Console function.

When testing with debugging publishes on, and using the console function to set the door ID, the debugging publishes contain the correct door data. These changes allow both the IM21 and IM24 door sensors to be paired to the boron. Note that the particle door data publishes are different due to the anti-tamper flag on the IM24. It could potentially also be beneficial to test with an actual sensor including the radar unit to make sure that the behaviour is still the same. 

All references to the IM21 in the code are changed to instead refer to a generic IM door sensor. Functions, comments and file names have been updated. 

A new function, isDoorStatusUnknown(), was added to ensure that the state machine remains in state 0 while the door sensor status is initialized to 0x99. 

Theresa and Kerry tested these changes on two actual sensors, with radar attached and Boron enclosed. One sensor was paired to an IM21, and the other sensor was paired to an IM24. After turning on debugging publishes and setting initial and duration timers to a short (5 seconds) period, the correct state transitions were observed when opening and closing the door sensors, and a duration alert was correctly sent out after the duration timer expired. Heartbeat messages were also received as expected. No regressions were observed with the IM21 as expected, and the IM24 functioned identically to the IM24 as expected. Additionally, with both sensors, it was observed that the state machine remained in state 0 as expected, while the door status was still its initial value 0x99.


trisapeace:
Task linked: [CU-3644bwf Update Sensor firmware to work with IM24 door sensors](https://app.clickup.com/t/3644bwf)

firmware/boron-ins-fsm/src/im21door.cpp
```diff
@@ -164,15 +164,17 @@ void threadBLEScanner(void *param) {
   while(true){
     //filter has to be remade each time, as globalDoorId (from eeprom) is usually read while this thread is already in while loop
     BleScanFilter filter; 
-    char address[18]; 
+    char address[18];
```
trisapeace, line 167: Thanks for removing the unnecessary spaces!

firmware/boron-ins-fsm/src/im21door.cpp
```diff
@@ -164,15 +164,17 @@ void threadBLEScanner(void *param) {
...
     sprintf(address, "B8:7C:6F:%02X:%02X:%02X", globalDoorID.byte3, globalDoorID.byte2, globalDoorID.byte1); 
-    filter.deviceName("iSensor ").address(address); 
+    filter.deviceName("iSensor ").address(address);
     sprintf(address, "8C:9A:22:%02X:%02X:%02X", globalDoorID.byte3, globalDoorID.byte2, globalDoorID.byte1); 
-    filter.address(address); 
+    filter.address(address);
+    sprintf(address, "AC:9A:22:%02X:%02X:%02X", globalDoorID.byte3, globalDoorID.byte2, globalDoorID.byte1); 
+    filter.address(address);
+
+    // scanning for IM24 door sensors of correct device ID
```
trisapeace, line 175: This scans for both IM21 and IM24 door sensors, right? If so, could you please update this comment?

firmware/boron-ins-fsm/src/im21door.cpp
```diff
@@ -164,15 +164,17 @@ void threadBLEScanner(void *param) {
   while(true){
```
trisapeace, line 164: Can you please add a line in the CHANGELOG about this addition?

trisapeace (commented):
Looks great. Glad it was such a quick change.

Thanks for explaining your testing in the description. I agree with you that this should be tested on a real sensor (ceiling and door unit together) to make sure that it completely works. Could you please do that and update the PR with a description of your testing? Thanks!

kerryzhang-brave-coop:
I'll be sure to test these changes with an actual sensor the next time I'm in the office and able to grab one

firmware/boron-ins-fsm/src/im21door.cpp
kerryzhang-brave-coop: No problem

firmware/boron-ins-fsm/src/im21door.cpp
kerryzhang-brave-coop: I updated this comment in my latest commit

firmware/boron-ins-fsm/src/im21door.cpp
kerryzhang-brave-coop: I added a line in the changelog for these changes in my latest commit

trisapeace (approved):

PR #189, opened 2022-11-25
Remove server-side state machine (CU-m0we0t)

trisapeace:
Remove everything related to the server-side state machine:
- removed `firmware_state_machine` flag from the `locations` DB table because from now on, all locations will be firmware state machine true
- removed `door_particlecoreid` from the `locations` DB table because all our devices use a single Boron now, so we will just need the `radar_particlecoreid` value
- removed Redis from the server because this was just used to store the XeThru, door, and state data from deprecated devices, but left it in the helm chart to leave our current instance where it is until I can extract the data somehow

Note that I've already added `[SSM]` to the name of all the locations that were server-side state machine devices in our database in all environments. This way if there is anything weird in their history, hopefully we'll realize that it is because of this.

## Test plan
- ✔️ Passes automated tests
- :heavy_check_mark: Deploy to dev
- :heavy_check_mark: Trigger a Duration alert on a real device
- :heavy_check_mark: Trigger a Stillness alert on a real device
- :heavy_check_mark: Get heartbeat from a real device
- :heavy_check_mark: Go through the chatbot flow in all ways
- :heavy_check_mark: Hit the PA endpoints using PA
   - :heavy_check_mark: Buy Twilio Number
   - :heavy_check_mark: Create location
   - :heavy_check_mark: Get Sensor clients 
- :heavy_check_mark: Hit the Dashboard endpoints
   - ✔️ Login/logout/unable to access pages without logging in
   - ✔️ Look at all the pages
   - ✔️ New Client
   - ✔️ Edit Client
   - ✔️ New Location (door ID and firmware state machine fields are gone)
   - ✔️ Edit Location (door ID and firmware state machine fields are gone)
   - :heavy_check_mark: Export CSV
- :heavy_check_mark: Plug in real XeThru device and confirm no new entries are going into Redis
- :heavy_check_mark: No errors in the logs


trisapeace:
Task linked: [CU-m0we0t Remove Sensor server state machine code](https://app.clickup.com/t/m0we0t)

PR #191, opened 2022-12-12
Do not transition to idle state upon receiving stillness or duration alert (CU-39w5av5) (CU-3c6my58)

kerryzhang-brave-coop:
This change modifies the sensor state machine to prevent it from transitioning to idle state after sending a stillness alert. Instead, it resets the stillness timer and transition back to stillness state. Additionally, this change also prevents the sensor state machine from transitioning to idle state after sending a duration alert. Instead, it transitions back to the duration state. The duration timer is not reset. 

These changes were tested on a Boron with the updated firmware uploaded to it. The following state machine paths were tested:

1. Set the stillness timer to a high value (such that a stillness alert is unlikely to be sent) and set the duration timer to a relatively low value (such as around 30 seconds). Transition to duration state and let the total duration reach the duration timer. Confirm that a duration alert is sent and that the state goes back to duration state. After the duration alert is sent and the state goes back to duration state, let the total duration reach the duration timer again. Confirm that no further duration alerts are sent. Open the door to reset to idle state and repeat this process. Confirm a duration alert is sent as expected after resetting the system.

2. Set the stillness timer to a high value (such that a stillness alert is unlikely to be sent) and set the duration timer to a relatively low value (such as around 30 seconds). Transition to stillness state and let the total duration reach the duration timer. Confirm that a duration alert is sent and that the state goes back to duration state. After the duration alert is sent and the state goes back to duration state, let the total duration reach the duration timer again. Confirm that no further duration alerts are sent. Open the door to reset to idle state and repeat this process. Confirm a duration alert is sent as expected after resetting the system.

3. Set the duration timer to a higher value than the stillness timer (such as around 75 seconds for the duration timer and 30 seconds for the stillness timer). Transition to stillness state and let the stillness duration reach the stillness timer. Confirm that a stillness alert is sent and that the state goes back to stillness state. Let the stillness duration reach the stillness timer again. Confirm that a stillness alert is sent again, and that the state goes back to stillness state. While in stillness state, wait for the total duration to reach the duration timer. Confirm that a duration alert is sent when the total duration reaches the duration timer.


trisapeace:
Task linked: [CU-39w5av5 Brainstorm ways to improve the Sensor system to not miss Stillness after a "Person Responded" alert](https://app.clickup.com/t/39w5av5)

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ the code was deployed.
 ### Changed
 
 - Travis CI to run new unit tests as a part of the build process. 
+- Sensor state machine to reset stillness timer and transition back to stillness state after receiving a stillness alert (CU-39w5av5).
+- Sensor state machine to transition back to duration state after receiving a duration alert (CU-39w5av5).
```
trisapeace, line 18: Can you update this one about Duration to include the part about Duration alerts only happening once and use the ClickUp Task https://app.clickup.com/t/3c6my58 for it?

Can you also reference that clickup ID in the name of this PR?

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -208,17 +212,19 @@ void state2_duration(){
     Log.warn("Door opened, session over, going to idle from state2_duration");
     publishStateTransition(2, 0, checkDoor.doorStatus, checkINS.iAverage);
     saveStateChange(2, 2);
+    hasDurationAlertBeenSent = 0; // session is over, so reset flag
     stateHandler = state0_idle;
-
   }
-  else if(millis() - state2_duration_timer >= state2_max_duration){
+  else if((millis() - state2_duration_timer >= state2_max_duration)
+          && !hasDurationAlertBeenSent){
```
trisapeace, line 219: Was the line too long if this was all on one line?

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -260,26 +266,29 @@ void state3_stillness(){
     Log.warn("door opened, session over, going from state3_stillness to idle");
     publishStateTransition(3, 0, checkDoor.doorStatus, checkINS.iAverage);
     saveStateChange(3, 2);
+    hasDurationAlertBeenSent = 0; // session is over, so reset flag
```
trisapeace, line 269: What do you think about resetting `hasDurationAlertBeenSent` in `state0_idle` instead of in both state 2 and state 3?

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -260,26 +266,29 @@ void state3_stillness(){
...
-    stateHandler = state0_idle;
+    hasDurationAlertBeenSent = 1; // a duration alert has been sent, so do not send another one
+    stateHandler = state2_duration;
   }
   else if(millis() - state3_stillness_timer >= state3_max_stillness_time){
 
-    Log.warn("stillness alert, going from state3 to idle after publish");
-    publishStateTransition(3, 0, checkDoor.doorStatus, checkINS.iAverage);
+    Log.warn("stillness alert, going from state3 to state2 after publish");
+    publishStateTransition(3, 2, checkDoor.doorStatus, checkINS.iAverage);
```
trisapeace, line 286: It looks like the way you've implemented this, you actually stay in state 3, so this needs to be updated or removed.

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -260,26 +266,29 @@ void state3_stillness(){
...
+    hasDurationAlertBeenSent = 0; // session is over, so reset flag
     stateHandler = state0_idle;
-
   }
-  else if(millis() - state2_duration_timer >= state2_max_duration){
+  else if((millis() - state2_duration_timer >= state2_max_duration)
+          && !hasDurationAlertBeenSent){
 
     Log.warn("See duration alert, going from state3 to idle after alert publish");
     publishStateTransition(3, 0, checkDoor.doorStatus, checkINS.iAverage);
```
trisapeace, line 276: This is no longer accurate because you're doing to State 2 instead of idle

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -208,17 +212,19 @@ void state2_duration(){
...
+    hasDurationAlertBeenSent = 0; // session is over, so reset flag
     stateHandler = state0_idle;
-
   }
-  else if(millis() - state2_duration_timer >= state2_max_duration){
+  else if((millis() - state2_duration_timer >= state2_max_duration)
+          && !hasDurationAlertBeenSent){
 
     Log.warn("See duration alert, going from state2_duration to idle after alert publish");
     publishStateTransition(2, 0, checkDoor.doorStatus, checkINS.iAverage);
```
trisapeace, line 222: This is no longer accurate because you're going to state 2 instead of idle.

trisapeace (commented):
Looks good. Just a couple things that I commented on.

CHANGELOG.md
kerryzhang-brave-coop: I made these changes in my latest commit. Thanks

firmware/boron-ins-fsm/src/stateMachine.cpp
kerryzhang-brave-coop: I put this on a new line out of habit, but you are right, it is short enough that it can go on the same line

firmware/boron-ins-fsm/src/stateMachine.cpp
kerryzhang-brave-coop: I like this more, thanks. I made the change in my latest commit

firmware/boron-ins-fsm/src/stateMachine.cpp
kerryzhang-brave-coop: Good catch. I updated the log and removed the state transition publish

firmware/boron-ins-fsm/src/stateMachine.cpp
kerryzhang-brave-coop: I updated this in my latest commit

firmware/boron-ins-fsm/src/stateMachine.cpp
kerryzhang-brave-coop: I updated this in my latest commit by removing the publish state transition. However, do you think that we should keep the saveStateChange() here? We publish an alert, but do not change states - please let me know what you think. I personally think we should remove or modify it.

firmware/boron-ins-fsm/src/stateMachine.cpp
kerryzhang-brave-coop: Do you think that we should keep the saveStateChange() here? We publish an alert, but do not change states - please let me know what you think. I personally think we should remove or modify it.

trisapeace: I actually think that we should keep the `saveStateChange` call. It's unfortunately named now that we are no longer changing state, but its purpose is to add the state change and its reason to the queue of changes that are sent to the server in the next heartbeat message. It's useful to have all the Stillness and Duration alerts appear as reasons in the heartbeat, so even though the state doesn't change.

firmware/boron-ins-fsm/src/stateMachine.cpp
trisapeace: Good question. See earlier response about `saveStateChange`.

firmware/boron-ins-fsm/src/stateMachine.cpp
trisapeace: I'd be open to changing the name of the `saveStateChange` function to something more accurate now. However, the JSON key `state` that is sent as part of the heartbeat will be a pain to change because the server is using it to know what to store in the DB. Up to you.

trisapeace (approved):
Looks great. See my comments about `saveStateChange` and make modifications if you want to. Either way, I'm good with this. Feel free to Squash and Merge when you're ready. Thanks!

PR #192, opened 2022-12-23
Do not send the current time as the last received door heartbeat when a Sensor is reset (CU-34atvnc)

kerryzhang-brave-coop:
To implement this change, we can store the time that the last received heartbeat was received on the Boron EEPROM memory. Then, upon restarting the Boron, get the value stored in memory rather than getting the current time. 

IMPORTANT NOTE: millis() gets the time elapsed after the Boron started, so some additional logic to calculate the actual time elapsed between the Boron startup and the value stored in memory will be required to make these changes work.

However, it is important to note that the changes must be able to handle the case where no value is stored in the EEPROM memory yet. One possibility would be to check if the memory at the location is all 1's, since the EEPROM memory initializes by default to all 1s.

### DIFF
--- a/firmware/boron-ins-fsm/src/flashAddresses.h
+++ b/firmware/boron-ins-fsm/src/flashAddresses.h
@@ -30,6 +30,6 @@
 //struct with three unsigned chars (uint8_t)
 #define ADDR_IM_DOORID 20	
 
-//next available address is 20 + 3 = 23
+#define LAST_HEARTBEAT_RECEIVED 24 
 
 #endif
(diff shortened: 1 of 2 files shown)


trisapeace:
Task linked: [CU-34atvnc Do not send the current time as the last received door heartbeat when a Sensor is reset](https://app.clickup.com/t/34atvnc)

trisapeace:
Closed in favour of https://github.com/bravetechnologycoop/BraveSensor/pull/198

PR #198, opened 2023-01-23
Change the door last seen behaviour (CU-34atvnc)

trisapeace:
- Because the Boron doesn't have a clock that keeps actual time, the millis() function returns the number of milliseconds since the device started/restarted. We were using this to calculate the last time that we saw a message from the door sensor. This works fine as long as we have seen at least one door message since the device last started up. In the case where we haven't seen any door messages yet, then it will return simply the number of milliseconds since the start up. This is almost certainly going to be less than the number of milliseconds since the actual last door message, and there is no way for the Boron to know how long its been since the last time because it's starting with no information. However, the DB has been tracking this information. So by sending '-1' when the Boron doesn't know how long it's been since the previous time, then the backend can assume that no new messages have come in since the last one that it knows about. This will prevent the bug in CU-34atvnc by making sure that the last seen door value doesn't automatically fall below the threshold on Boron restart.
- Renamed the doorLastHeartbeat to doorLastMessage to be more accurate. I think it was causing some confusion as to its actual implmentation on the other side
- Also updated the README to be more generic with IM21 and IM24 door sensors where appropriate

## Test Plan
- [x] Deploy to dev and to the two Sensors in my bathroom, one with an IM21 and the other with an IM24
- [x] Run smoketests
- [x] Restart Sensors by unplugging and replugging
   - [x] The first Boron heartbeat sends `"doorLastMessage": "-1"`
   - [x] The Sensor Vital is stored in the `sensors_vitals` table with `lastSeenDoor` value the same as the previous row
   - [x] The Sensor Vital is cached in the `sensors_vitals_cache` table with `lastSeenDoor` value the same as the previous row
- [x] Wait for a subsequent Boron heartbeat which should contain a `doorLastMessage` that is not equal to -1 (it will probably be 1000)
   - [x] The Sensor Vital is stored in the `sensors_vitals` table with `lastSeenDoor` value a new value that is the current time minus the number of ms given in `doorLastMessage`
   - [x] The Sensor Vital is cached in the `sensors_vitals_cache` table with `lastSeenDoor` value a new value that is the current time minus the number of ms given in `doorLastMessage`
- [x] Open and close the door a few times, and then wait for another Boron heartbeat which should contain a `doorLastMessage` that is not equal to -1
   - [x] The Sensor Vital is stored in the `sensors_vitals` table with `lastSeenDoor` value a new value that is the current time minus the number of ms given in `doorLastMessage`
   - [x] The Sensor Vital is cached in the `sensors_vitals_cache` table with `lastSeenDoor` value a new value that is the current time minus the number of ms given in `doorLastMessage`
- [x] While the Boron is turned off, erase the row in `sensors_vitals_cache` for one of the Sensors so that the next one will seem like the very first Boron heartbeat for the given Sensor
   - [x] The Sensor Vital is stored in the `sensors_vitals` table with `lastSeenDoor` value a the current time
   - [x] The Sensor Vital is cached in the `sensors_vitals_cache` table with `lastSeenDoor` value that is the current time 
- [x] Trigger a real Stillness Alert and run through the chatbot

### DIFF
--- a/server/vitals.js
+++ b/server/vitals.js
@@ -186,13 +186,9 @@ async function handleHeartbeat(req, res) {
         // Must send 200 so as not to be throttled by Particle (ref: https://docs.particle.io/reference/device-cloud/webhooks/#limits)
         res.status(200).json(errorMessage)
       } else {
-        const currentDbTime = await db.getCurrentTime()
-
         const message = JSON.parse(req.body.data)
         const doorMissedMessagesCount = message.doorMissedMsg
         const doorLowBatteryFlag = message.doorLowBatt
-        const doorTimeSinceLastHeartbeat = new Date(currentDbTime)
-        doorTimeSinceLastHeartbeat.setMilliseconds(currentDbTime.getMilliseconds() - message.doorLastHeartbeat)
         const resetReason = message.resetReason
         const stateTransitionsArray = message.states.map(convertStateArrayToObject)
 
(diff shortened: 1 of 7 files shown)


trisapeace:
Task linked: [CU-34atvnc Do not send the current time as the last received door heartbeat when a Sensor is reset](https://app.clickup.com/t/34atvnc)

server/vitals.js
```diff
@@ -183,28 +183,27 @@ async function handleHeartbeat(req, res) {
...
-          location.locationid,
-          doorMissedMessagesCount,
-          doorLowBatteryFlag,
-          doorTimeSinceLastHeartbeat,
-          resetReason,
-          stateTransitionsArray,
-        )
+        let doorLastSeenAt
+        if (message.doorLastMessage.toString() === '-1') {
+          // TODO transaction?
```
trisapeace, line 198: Put it in a transaction?

trisapeace: Decided not to in order to not introduce new places for deadlocks. I think the chance to this function being called twice for the same location in a short enough period of time to cause a race condition is very low.

server/vitals.js
```diff
@@ -183,28 +183,27 @@ async function handleHeartbeat(req, res) {
...
+          const mostRecentSensorVitals = db.getMostRecentSensorsVitalWithLocationid(location.locationid)
+          doorLastSeenAt = mostRecentSensorVitals.doorLastSeenAt
+        } else {
+          const currentDbTime = await db.getCurrentTime()
+          doorLastSeenAt = DateTime.fromJSDate(currentDbTime).minus(message.doorLastMessage)
+        }
+
+        await db.logSensorsVital(location.locationid, doorMissedMessagesCount, doorLowBatteryFlag, doorLastSeenAt, resetReason, stateTransitionsArray)
```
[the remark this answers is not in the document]
trisapeace: Done.

PR #202, opened 2023-02-13
add and check for api key (CU-yfdndf)

susanchung-brave:
for security purposes and to prevent random yoo-hoos from pretending to be our stuff!

testplan:
- [x] make a POST request to heartbeat with an incorrect API key to confirm that the error message gets displayed
    - [x] check that no other changes are permitted as a result (receive a text that states that the sensor is still disconnected)
    - [x] confirm that it also doesn't send a low battery alert text message when there's an incorrect API key
- [x] make a POST request to heartbeat with the correct API key to confirm that it works
    - [x] check that access is allowed by receiving a text message that sensor has reconnected
    - [x] confirm that it also works for low battery alerts by making a modified POST request
- [x] use actual sensor with incorrect API key to confirm that it doesn't work
    - [x] Stillness alert: check the logs for error message + shouldn't receive Stillness alert message
    - [x] Duration alert: check the logs for error message + shouldn't receive Duration alert message
- [x] use actual sensor with correct API key to confirm that it works
    - [x] Stillness alert: should receive Stillness alert message
    - [x] Duration alert: should receive Duration alert message


trisapeace:
Task linked: [CU-yfdndf Add an API key to Particle Webhooks](https://app.clickup.com/t/yfdndf)

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -56,6 +61,54 @@ async function lowBatteryHeartbeat(coreId) {
...
+      const badRequest = {
+        coreid: radar_coreID,
+        data: `{"doorMissedMsg": 0, "doorLowBatt": true, "doorLastMessage": 1000, "resetReason": "NONE", "states":[]}`,
+        api_key: badpassword,
+      }
+      const response = await chai.request(server).post('/api/heartbeat').send(badRequest)
+      expect(response).to.have.status(200)
+      it('should log the error because the api key does not match', () => {
+        expect(helpers.log).to.have.been.calledWith(`Access not allowed to: incorrect key ${badpassword}`)
+      })
```
trisapeace, line 96: This is a separate test, so would be better at the same level as the other `it` functions.

Whenever an `expect` fails, then it will return out of the `it` function that it's in. Therefore, if the `expect(response).to.have.status(200)` on line 93 fails, then this test will never run. Whereas, we would want this code to run and know whether the error was logged regardless of whether the right status code was being returned or not.

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -56,6 +61,54 @@ async function lowBatteryHeartbeat(coreId) {
...
+      sandbox.restore()
+    })
+
+    it('should return 200 for the valid request', async () => {
+      const badRequest = {
+        coreid: radar_coreID,
+        data: `{"doorMissedMsg": 0, "doorLowBatt": true, "doorLastMessage": 1000, "resetReason": "NONE", "states":[]}`,
+        api_key: badpassword,
+      }
+      const response = await chai.request(server).post('/api/heartbeat').send(badRequest)
```
trisapeace, line 92: I suggest that you factor this part out and put it in a descriptively-named function near the top of the file like what was done with `normalHeartbeat`.

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -56,6 +61,54 @@ async function lowBatteryHeartbeat(coreId) {
...
+      sandbox.stub(braveAlerter, 'startAlertSession')
+      sandbox.stub(braveAlerter, 'sendSingleAlert')
+      sandbox.spy(helpers, 'logSentry')
+      sandbox.stub(db, 'logSensorsVital')
+    })
+
+    afterEach(async () => {
+      await db.clearTables()
+      sandbox.restore()
+    })
```
trisapeace, line 84: Do you have an understanding of how the `beforeEach` and `afterEach` functions work?

server/vitals.js
```diff
@@ -180,7 +182,14 @@ async function handleHeartbeat(req, res) {
     if (validationErrors.isEmpty()) {
       const coreId = req.body.coreid
       const location = await db.getLocationFromParticleCoreID(coreId)
```
trisapeace, line 184: There is a speed optimization possible related to this line here. In general, calls to the database are slow. What do you think you could do here to make the `handleHeartbeat` function as lean as it can be.

server/test/integration/vitalsTest/handleHeartbeatTest.js
susanchung-brave: ohhhh okay that makes sense. thanks! i'll move it out of this level!

server/test/integration/vitalsTest/handleHeartbeatTest.js
susanchung-brave: sounds good! thank you : )

server/test/integration/vitalsTest/handleHeartbeatTest.js
susanchung-brave: i've had experience with them! my user of the `beforeEach` and `afterEach` here were very much based on the other tests  😅

server/test/integration/serverTest.js
```diff
@@ -60,13 +63,33 @@ describe('Brave Sensor server', () => {
...
+      it('should return 200', () => {
+        expect(this.response).to.have.status(200)
+      })
+
+      it('should log the error', () => {
+        expect(helpers.logError).to.be.calledWithExactly(`Access not allowed: incorrect key ${badpassword}`)
+      })
+
+      it('should not start the alert session since access should not have been given', () => {
+        expect(braveAlerter.startAlertSession).to.not.be.calledOnce
```
trisapeace, line 86: What do you think of using `to.not.be.called` instead of `to.not.be.calledOnce`. Having the "Once" in there makes me wonder whether it would pass if it were called more than once.

server/test/integration/serverTest.js
```diff
@@ -81,9 +104,27 @@ describe('Brave Sensor server', () => {
...
```
trisapeace, line 121: Same here about the "Once"

server/test/integration/serverTest.js
```diff
@@ -81,9 +104,27 @@ describe('Brave Sensor server', () => {
...
+
+      it('should return 200', () => {
+        expect(this.response).to.have.status(200)
+      })
+
+      it('should log the error', () => {
+        expect(helpers.logError).to.be.calledWithExactly(`Access not allowed: incorrect key ${badpassword}`)
+      })
+
+      it('should not start the alert session since access should not have been given', () => {
```
trisapeace, line 120: This is great. I'm glad that you thought to check this.

server/index.js
```diff
@@ -113,33 +115,42 @@ async function handleAlert(location, alertType) {
...
+          response.status(200).json('OK')
+        }
       } else {
-        const errorMessage = `Bad request to ${request.path}: Invalid event type`
-        helpers.logError(errorMessage)
-      }
-      const location = await db.getLocationFromParticleCoreID(coreId)
-      if (!location) {
-        const errorMessage = `Bad request to ${request.path}: no location matches the coreID ${coreId}`
+        const errorMessage = `Access not allowed: incorrect key ${apiKey}`
```
trisapeace, line 150: For this real error log, I'd actually leave out the incorrect key value that was sent. 

In general, we want to log enough data that we can debug problems, but not so much that any hacker would be able to learn more about our system. So it's possible that someone provides an API key that is close, like they just did a typo. In which case, that typo would now be logged and a potential hacker might guess what the problem was... maybe extra whitespace or capitalization and then be able to know how API key.

server/vitals.js
```diff
@@ -171,49 +173,65 @@ async function sendLowBatteryAlert(locationid) {
...
+            doorMissedMessagesCount,
+            doorLowBatteryFlag,
+            doorLastSeenAt,
+            resetReason,
+            stateTransitionsArray,
+          )
+          res.status(200).json('OK')
+        }
+      } else {
+        const errorMessage = `Access not allowed: incorrect key ${apiKey}`
```
trisapeace, line 231: Same thing about logging the incorrect key here.

server/vitals.js
```diff
@@ -171,49 +173,65 @@ async function sendLowBatteryAlert(locationid) {
...
-        let doorLastSeenAt
-        if (message.doorLastMessage.toString() === '-1') {
-          const mostRecentSensorVitals = await db.getMostRecentSensorsVitalWithLocationid(location.locationid)
-          if (mostRecentSensorVitals === null) {
-            // If it doesn't have any previous heartbeats, just use the current time
-            const currentDbTime = await db.getCurrentTime()
-            doorLastSeenAt = currentDbTime
+      if (webhookAPIKey === apiKey) {
+        const coreId = req.body.coreid
+        const location = await db.getLocationFromParticleCoreID(coreId)
```
trisapeace, line 187: Nice! You got it!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ the code was deployed.
 ### Security
 
 - Upgrade dependencies (CU-860phzbq5).
+- Add API Key to webhooks (CU-yfdndf).
```
trisapeace, line 17: I'd specify in here that this is the Particle webhooks because it's possible that other third party systems that we interact with also use webhooks.

server/test/integration/vitalsTest/handleHeartbeatTest.js
trisapeace: Awesome! Just checking that you understood what was going on here :smile:

server/index.js
```diff
@@ -113,33 +115,42 @@ async function handleAlert(location, alertType) {
   }
 }
 
-app.post('/api/sensorEvent', Validator.body(['coreid', 'event']).exists(), async (request, response) => {
+app.post('/api/sensorEvent', Validator.body(['coreid', 'event', 'api_key']).exists(), async (request, response) => {
```
trisapeace, line 118: Have you tried to see if this works on your real Sensor? If so, could you please update your test plan to include what tests you ran?

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -56,6 +87,69 @@ async function lowBatteryHeartbeat(coreId) {
...
+      sandbox.spy(helpers, 'log')
+      sandbox.stub(db, 'logSensorsVital')
+    })
+
+    afterEach(async () => {
+      await db.clearTables()
+      sandbox.restore()
+    })
+
+    it('should return 200 for the valid request', async () => {
```
trisapeace, line 113: What do you mean by "for the valid request" in the name of this test?

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -56,6 +87,69 @@ async function lowBatteryHeartbeat(coreId) {
...
+      const request = {
+        coreid: radar_coreID,
+        data: `{"doorMissedMsg": 0, "doorLowBatt": true, "doorLastMessage": 1000, "resetReason": "NONE", "states":[]}`,
+        api_key: badpassword,
+      }
+      const response = await chai.request(server).post('/api/heartbeat').send(request)
```
trisapeace, line 119: Is there a particular reason to define this request here? Is this different than `normalHeartbeatWithIncorrectAPIKey` in an important way?

server/test/integration/serverTest.js
susanchung-brave: oh interesting, thank you for pointing that out! i'll change it and test it again!

server/test/integration/serverTest.js
susanchung-brave: cool thanks!

server/index.js
susanchung-brave: ohhhh that's a great point, thank you! i'll update the error log accordingly.

server/test/integration/vitalsTest/handleHeartbeatTest.js
susanchung-brave: i think i meant to say that it's a valid request because all the fields have information filled in!

server/test/integration/vitalsTest/handleHeartbeatTest.js
susanchung-brave: i think i got thrown off on how to get status 200 from it and then forgot to come back to update it to use the helpers!

server/vitals.js
susanchung-brave: thank you for your guidance with that!!

server/vitals.js
susanchung-brave: got it!

server/test/integration/serverTest.js
susanchung-brave: thank you :)

server/.env.example
```diff
@@ -136,6 +136,10 @@ PA_PASSWORD_SECONDARY_TEST=ABpassword
 DOMAIN=domain.brave.coop
 DOMAIN_TEST=domain.brave.coop
 
+# API Key for Particle Webhook to make API calls to Heartbeat
```
trisapeace, line 139: This comment specifies heartbeat, but it's also used in the other calls. Could you please change it either so that it includes those explicitly or make it more general by removing the Heartbeat part?

trisapeace (commented):
Looks good. Just one change to one comment.

server/.env.example
susanchung-brave: thanks for catching that! really appreciate it : )

PR #209, opened 2023-04-04
 Split Client and Location is_active into is_displayed, is_sending_alerts, is_sending_vitals (CU-860ptt5rp)

trisapeace:
- This gives us more fine-grained control over the messages and the display properties of the clients
- Only displays Clients and Locations on the dashboard if their is_displayed value is true
- Only sends vitals messages about Locations whose is_sending_vitals value is true and their Client's is_sending_vitals is true
- Only sends Stillness/Duration alerts from Locations whose is_sending_alerts value is true and their Client's is_sending_alerts is true

## Test Plan
- [x] Dashboard
   - [x] Overview
      - [x] Can see all the columns
      - [x] Clients that have `is_displayed=false` do not appear
      - [x] Locations that have `is_displayed=false` do not appear
      - [x] Clients have correct values for `is_sending_alerts` and `is_sending_vitals`
      - [x] Locations have correct values for `is_sending_alerts` and `is_sending_vitals` based on their DB value and their client's DB value
      - [x] Clients dropdown menu contains all Clients with `is_displayed = true` and only them
   - [x] Client Page
      - [x] Can view client page of a client with `is_displayed=false` by using its URL
      - [x] Can see all the columns
      - [x] Locations that have `is_displayed=false` do not appear
      - [x] Locations have correct values for `is_sending_alerts` and `is_sending_vitals` based on their DB value and their client's DB value
      - [x] Clients dropdown menu contains all Clients with `is_displayed = true` and only them
   - [x] Location Page
      - [x] No change
      - [x] Can view Location page of a Location with `is_displayed=false` by using its URL
      - [x] Clients dropdown menu contains all Clients with `is_displayed = true` and only them
   - [x] Edit Client Page
      - [x] Can view Edit Client page of a Client with `is_displayed=false` by using its URL
      - [x] No longer have "Is this Client active?" question
      - [x] `is_displayed` question only accepts true or false
      - [x] `is_sending_alerts` question only accepts true or false
      - [x] `is_sending_vitals` question only accepts true or false
      - [x] Change all values and submit, see changes in the DB
      - [x] Clients dropdown menu contains all Clients with `is_displayed = true` and only them
   - [x] New Client Page
      - [x] Reminder note above the submit button says the correct values of `is_displayed`, `is_sending_alerts`, and `is_sending_vitals`
      - [x] Submit and see new client in DB with `is_displayed=true`, `is_sending_alerts=false`, and `is_sending_vitals=false`
[...]

PR #210, opened 2023-04-18
Don't display locations on Vitals page if client.isDisplayed is false (CU-860ptt5rp)

trisapeace:
- This was missed in #209 

## Test Plan
- [x] Vitals Page
   - [x] Locations whose Client has is_displayed=false do not appear
   - [x] Locations that have is_displayed=false do not appear
   - [x] Clients dropdown menu contains all Clients with is_displayed = true and only them

### DIFF
--- a/server/dashboard.js
+++ b/server/dashboard.js
@@ -105,7 +105,7 @@ async function renderVitalsPage(req, res) {
 
     const sensorsVitals = await db.getRecentSensorsVitals()
     for (const sensorsVital of sensorsVitals) {
-      if (sensorsVital.location.isDisplayed) {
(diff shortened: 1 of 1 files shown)


trisapeace:
Task linked: [CU-860ptt5rp Split the Sensors isActive status into three parts: isDisplayed, isSendingAlerts, and isSendingVitals](https://app.clickup.com/t/860ptt5rp)

trisapeace:
I also noticed that editing a location whose client has `isDisplayed=false` gives the weird result of it changing client. This is because the Client dropdown menu only shows displayed clients. 

I'm not going to bother doing anything about this right now since these Dashboards are due to be replaced by PA and I don't think we'll run into the situation often where someone will try to update a location that they would need to get to through its URL without a link to it anywhere else on the Dashboard.

PR #211, opened 2023-04-24
Allow heartbeatPhoneNumbers to be empty from the Dashboard (CU-2uad09a)

trisapeace:
- It is a valid use case to have this field be an empty array. In this case, only the Responder Phone Numbers will receive vitals text messages. Before this code change, we would need to make this happen directly in the DB, which carries the risk of human error. This will be better, and it wasn't a complicated change, so it was worth doing even though the Dashboard is supposed to move to PA

## Test plan
- [x] Deploy to dev
- [x] Add a new client in the Dashboard
   - [x] With nothing in the "Heartbeat Phone Numbers" field - should create client with an empty array of heartbeatPhoneNumbers in the DB
   - [x] With only whitespace in the "Heartbeat Phone Numbers" field - should show show validation error
   - [x] With a valid phone number in the "Heartbeat Phone Numbers" field - should create client
   - [x] With a invalid phone number in the "Heartbeat Phone Numbers" field - should show validation error
- [x] Edit a client in the Dashboard
   - [x] With nothing in the "Heartbeat Phone Numbers" field - should modify client to have an empty array of heartbeatPhoneNumbers in the DB
   - [x] With only whitespace in the "Heartbeat Phone Numbers" field - should show show validation error
   - [x] With a valid phone number in the "Heartbeat Phone Numbers" field - should modify client
   - [x] With a invalid phone number in the "Heartbeat Phone Numbers" field - should show validation error

### DIFF
--- a/server/dashboard.js
+++ b/server/dashboard.js
@@ -385,7 +385,7 @@ async function renderClientDetailsPage(req, res) {
 }
 
 const validateNewClient = [
-  Validator.body(['displayName', 'fallbackPhoneNumbers', 'fromPhoneNumber', 'heartbeatPhoneNumbers', 'incidentCategories']).trim().notEmpty(),
+  Validator.body(['displayName', 'fallbackPhoneNumbers', 'fromPhoneNumber', 'incidentCategories']).trim().notEmpty(),
   Validator.body(['reminderTimeout', 'fallbackTimeout']).trim().isInt({ min: 0 }),
   Validator.oneOf([
     Validator.body(['responderPhoneNumbers']).trim().notEmpty(),
@@ -421,6 +421,10 @@ async function submitNewClient(req, res) {
           : null
       const newResponderPushId = data.responderPushId && data.responderPushId.trim() !== '' ? data.responderPushId : null
       const newAlertApiKey = data.alertApiKey && data.alertApiKey.trim() !== '' ? data.alertApiKey : null
+      const newHeartbeatPhoneNumbers =
+        data.heartbeatPhoneNumbers !== undefined && data.heartbeatPhoneNumbers.trim() !== ''
+          ? data.heartbeatPhoneNumbers.split(',').map(phone => phone.trim())
+          : []
 
       const newClient = await db.createClient(
         data.displayName,
@@ -431,7 +435,7 @@ async function submitNewClient(req, res) {
         data.fallbackPhoneNumbers.split(',').map(phone => phone.trim()),
         data.fromPhoneNumber,
         data.fallbackTimeout,
-        data.heartbeatPhoneNumbers.split(',').map(phone => phone.trim()),
+        newHeartbeatPhoneNumbers,
         data.incidentCategories.split(',').map(category => category.trim()),
         true,
         false,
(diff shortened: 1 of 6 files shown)


trisapeace:
Task linked: [CU-2uad09a Sensors Dashboard Allow Heartbeat Phone Number Empty](https://app.clickup.com/t/2uad09a)

server/dashboard.js
```diff
@@ -499,6 +502,10 @@ async function submitEditClient(req, res) {
           : null
       const newResponderPushId = data.responderPushId && data.responderPushId.trim() !== '' ? data.responderPushId : null
       const newAlertApiKey = data.alertApiKey && data.alertApiKey.trim() !== '' ? data.alertApiKey : null
+      const newHeartbeatPhoneNumbers =
+        data.heartbeatPhoneNumbers !== undefined && data.heartbeatPhoneNumbers.trim() !== ''
+          ? data.heartbeatPhoneNumbers.split(',').map(phone => phone.trim())
+          : []
```
susanchung-brave, line 508: Is the difference between line 508 for newHeartbeatPhoneNumbers `[]` and line 502 for newResponderPhoneNumbers `null` that newResponderPhoneNumbers isn't allowed to be empty? Furthermore, why do we check that `data.heartbeatPhoneNumbers !== undefined` but don't have to for `data.responderPhoneNumbers`?

Sorry for the dumb question, my brain is a little muddy at the moment and I got curious :)

susanchung-brave (approved):

server/dashboard.js
trisapeace: Honestly, I don't remember exactly, but I think it has to do with when I was trying to write the Brave Alert App, which was going to be a replacement for SMS on the Responder Phones. At that time, the Responder Phone could be designated with the AppPushID and AlertApiKey together OR with a responderPhoneNumber. So the responderPhoneNumber field either contained a single phone number or it was `null`. So when the field is blank, that means that it should be `null` in the DB and that AppPushID and AlertApiKey should not be blank/null.

Whereas now, what I'm actually wanting is an empty array so that I can just do a for-loop over the array and send the right number (in this case 0) messages. If I used `null` instead, then I'd have to do an extra check for null or the loop over the array would (I think) not work the way that I'm expecting.

I did notice this discrepancy while I was coding it too, and thought about changing the one for `responderPhoneNumbers` since now that we allow multiple Responder Phone Numbers, an empty array is probably better than `null` for the same for-loop reason. But in the end, I didn't want to change it and maybe end up with unanticipated consequences for what I hoped was a simple change.

As for the `!== undefined` check. I do think that it should have it in the `responderPhoneNumbers` case, but when I tried it out as is without the check, it seemed to work fine. And then I tried `heartbeatPhoneNumbers` without the check but that didn't work. So again, it fell under the category of stuff that I didn't want to touch in case it caused more issues than it solved. Normally, I'd consider that a bit of a lazy answer, but the Dashboard is something that will hopefully be redesigned when it's moved to PA and is only used by us internally, so it wasn't worth the time for me to figure it out.

susanchung-brave: Oooh that makes sense! Thank you for taking the time to explain this to me : )

PR #213, opened 2023-04-27
Add a second (longer) stillness threshold (CU-860q2an4r)

trisapeace:
Increases the stillness threshold after the first Stillness Alert
    - Adds a Console function to set and change the Long Stillness Timer
    - After the first Stillness Alert, uses the Long Stillness Timer instead
      of the normal Stillness Timer to determine how long to wait before
      triggering subsequent Stillness Alerts
Removes SUBSEQUENT_ALERT_MESSAGE_THRESHOLD
    - I can't find our reasoning for adding this in the first place. I think
      that it was to not annoy our Responders with too many alerts in a row.
      I think that as long as our clients have reasonable Duration and
      Stillness thresholds, this shouldn't be an issue. There will only
      ever be one Duration Alert during a session and each Stillness Alert
      should be spaced out by the Stillness Threshold. This may be an
      artifact of something that was used by the Server-side State Machine
      and isn't needed for the Firmware State Machine
    - I'm removing it now because it confuses the issue of when
      subsequent Stillness Alerts will actually be sent to the Responder
      Phone
db.clearLocation also clears relevant rows in `sensors_vitals*`
    - So that the smoke tests will still run even if the SmokeTestLocation
      somehow gets some heartbeats in the DB

## Test Plan
- [x] Deploy v9.3.4 to my test Sensor, which has a version where the value in of `ADDR_INITIALIZE_STATE3_MAX_LONG_STILLNESS_TIME_FLAG` in EEPROM is `0xFFFF`. This should be the value for Borons who are brand new or for ones with firmware before this version
    - [x] Long Stillness Alert Threshold is copied from Stillness Alert Threshold
- [x] Deploy v9.3.5 to my test Sensor, which has a version where the value in of `ADDR_INITIALIZE_STATE3_MAX_LONG_STILLNESS_TIME_FLAG` in EEPROM is `0x8888`. This should be the value for Borons that have already been at this version at this version once before and had their Long Stillness Alert Threshold value initialized
   - [x] Retrieves the previously initialized version of Long Stillness Alert Threshold from EEPROM
- [x] Deploy v9.3.6 to my test Sensor, which is the real version that I want to test and deploy
   - [x] Use console function to update all the values, including the Long Stillness Threshold
   - [x] Sending invalid input to the Long Stillness Threshold console function returns -1
   - [x] Restart the Sensor and echo all the console functions to see that they retrieved their correct values
   - [x] First Stillness alert is triggered after the Stillness Alert Threshold
   - [x] Second, third, fourth, etc Stillness alerts are triggered after the Long Stillness Alert Threshold
   - [x] If a Duration alert occurs before the first Stillness Alert, the first Stillness Alert is still after the Stillness Alert Threshold
   - [x] If a Duration alert occurs after the first Stillness Alert, then all subsequent Stillness Alerts are still after the Long Stillness Alert Threshold
   - [x] The first Stillness Alert of all subsequent sessions is still the Stillness Alert Threshold
   - [x] Go back and forth between States 2 and 3 (Movement and Stillness, respectively) and the first Stillness Alert is still after the Stillness Alert Threshold and subsequent Stillness Alerts are after the Long Stillness Alert Threshold
- [x] Deploy server to dev
   - [x] Stillness/Duration Alerts in quick succession always start their own session if there isn't one already
   - [x] Stillness/Duration Alerts in quick succession always send the "additional alert" message if there is already a session in progress
   - [x] Run the smoke tests in such a way that the SmokeTestClient and SmokeTestLocation are still in the DB. Then add a row in `sensors_vitals_cache`, `sensors_vitals`, and `clients_extension` referring to that location/client. Then run the smoke tests again and see that they pass


firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -155,7 +156,7 @@ int stillness_timer_set(String input) {
 
     // if e, echo the current threshold
     if (*holder == 'e') {
-        EEPROM.get(ADDR_STATE3_MAX_STILLNES_TIME, state3_max_stillness_time);
+        EEPROM.get(ADDR_STATE3_MAX_STILLNESS_TIME, state3_max_stillness_time);
```
trisapeace, line 159: Fixing the typo in this constant name to avoid future headaches.

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -241,21 +267,23 @@ void state3_stillness() {
...
+        Particle.publish("Stillness Alert", "stillness alert!!!", PRIVATE);
+        state3_stillness_timer = millis();                     // reset the stillness timer
+        max_stillness_time = &state3_max_long_stillness_time;  // set to compare against the longer stillness threshold for the rest of this session
+        stateHandler = state3_stillness;
+    }
     else if ((millis() - state2_duration_timer >= state2_max_duration) && !hasDurationAlertBeenSent) {
-        Log.warn("See duration alert, going from state3 to state2 after alert publish");
-        publishStateTransition(3, 2, checkDoor.doorStatus, checkINS.iAverage);
+        Log.warn("duration alert, remaining in state3 after publish");
+        publishStateTransition(3, 3, checkDoor.doorStatus, checkINS.iAverage);
```
trisapeace, line 282: I changed this because there isn't really any reason to go back into the Duration state, we didn't see any movement. So it would have just transitioned from Duration state back to Stillness. Might as well just stay in Stillness.

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -241,21 +267,23 @@ void state3_stillness() {
         saveStateChangeOrAlert(3, 2);
         stateHandler = state0_idle;
     }
+    else if (millis() - state3_stillness_timer >= *max_stillness_time) {
```
trisapeace, line 270: Moved this to before the Duration Alert test because if we have reached the criteria for both a Stillness and a Duration Alert,  I want the Stillness one to be sent first because it's more important.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
susanchung-brave: Nice catch!!

firmware/boron-ins-fsm/src/stateMachine.cpp
susanchung-brave: Yeah that's great! The transitions between duration <-> stillness were confusing me when I was testing my sensor earlier so I'm glad you made this change!

PR #216, opened 2023-06-02
Fix inititalization of Long Stillness Timer (CU-860r11fng)

trisapeace:
- On the very first initialization of v9.4.0, the Long Stillness Timer value in EEPROM was begin set correctly to the current value of the previously-only Stillness Timer. But then that value wasn't being copied into the variable that was used in the State Machien to determine whether the threshold had been hit *doh*. So added that in
- And add more information to the debug mode messages

## Test Plan
- [x] Reproduce the problem on my local Sensor: without making any other changes, comment out the if-else statement in `stateMachine.cpp` so that the state3_max_long_stillness initialization part always happens
   - [x] Set my Short Stillness Timer = 15 and Long Stillness Timer = 30, then restart the Sensor. Turn on debug mode. See that the stillness_timer value is 15000 for the first Stillness Alert and then 120000 for the subsequent Stillness alerts
   - [x] Set the Long Stillness Timer = 30. See that the stillness_timer value is 15000 for the  first Stillness Alert and then 30000 for the subsequent Stillness Alerts
- [x] Fixed initialization on my local Sensor: make the changes in this PR but also comment out the if-else statement in `stateMachine.cpp` so that the state3_max_long_stillness initialization part always happens
   - [x] Set my Short Stillness Timer = 15 and Long Stillness Timer = 30, then restart the Sensor. Turn on debug mode. See that the stillness_timer value is 15000 for the first Stillness Alert and then 15000 for the subsequent Stillness alerts (to simulate when there was only the Short Stillness Timer) and that echoing the Long Stillness Timer returns 15
   - [x] Set the Long Stillness Timer = 30. See that the stillness_timer value is 15000 for the  first Stillness Alert and then 30000 for the subsequent Stillness Alerts
- [x] Double-check non-initialization case: make changes in this PR without anything commented out. On my Sensor, this means that I'll end up in the `else` case where it will read the Short and Long Stillness Timers from EEPROM.
   - [x] Set my Short Stillness Timer = 15 and Long Stillness Timer = 30, then restart the Sensor. Turn on debug mode. See that the stillness_timer value is 15000 for the first Stillness Alert and then 30000 for the subsequent Stillness alerts and that echoing the Long Stillness Timer returns 30

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ the code was deployed.
 
 ## [Unreleased]
 
+### Fixed
+
+- Initialization of Long Stillness Threshold (CU-860r11fng).
+
 ## [9.4.0] - 2023-05-23
 
 ### Changed
--- a/firmware/README.md
+++ b/firmware/README.md
@@ -517,9 +517,13 @@ Debug Message
 **Event Data**
 
 1. state: the current state the machine is in
-2. door_status: the current door status byte
-3. INS_val: the current filtered inPhase value from the INS radar
-4. timer_status: if the current state uses a timer, this contains the time in milliseconds that the timer has counted up to thus far. If the current state does not contain a timer, this is set to 0.
+1. door_status: the current door status byte
+1. INS_val: the current filtered inPhase value from the INS radar
+1. INS_threshold: the current threshold that we compare the INS_val against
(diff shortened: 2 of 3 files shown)


trisapeace:
Task linked: [CU-860r11fng Debug and fix long stillness alert initialization](https://app.clickup.com/t/860r11fng)

susanchung-brave (approved):

PR #217, opened 2023-06-27
Add auto-off for debug messages (CU-860r0z9zb)

trisapeace:
- Debug publishes will automatically turn off after 8 hours so that we don't end up accidentally leaving it on for days at a time, using up all our Particle data operations
- Removed the `stateMachineDebugFlag` from stateMachine.h and just included `debugFlags.h` instead. That will reduce the repetition
- Moved `lastDebugPublish` into `debugFlags.h` to keep all the debug-related things together and to clarify that it is a global variable
- Moved the `DEBUG_PUBLISH_INTERVAL` constant into `debugFlags.h` also to keep all the debug-related things together

## Test Plan
- [x] Change values of `DEBUG_PUBLISH_INTERVAL` and `DEBUG_AUTO_OFF_THRESHOLD` to something easy to test in real time and deploy to real Sensor (see version 9.5.6)
- [x] When Sensor first turns on
   - [x] No debug messages
- [x] Turn on the debug messages
   - [x] Before the `DEBUG_AUTO_OFF_THRESHOLD`
      - [x] See a debug message every `DEBUG_PUBLISH_INTERVAL` ms
      - [x] See state transition messages
      - [x] When you echo the debug publish, get back `1`
   - [x] After the `DEBUG_AUTO_OFF_THRESHOLD`
      - [x] No debug messages
      - [x] No state transition messages
      - [x] When you echo the debug publish, get back `0`
- [x] If you turn debug mode on again while it's already on, then it doesn't turn off until `DEBUG_AUTO_OFF_THRESHOLD` after the most recent time that you turned it on

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ the code was deployed.
 
 ## [Unreleased]
 
+### Added
+
+- Auto-shut off for debug publishes after 8 hours (CU-860r0z9zb).
+
 ## [9.5.0] - 2023-06-02
 
 ### Fixed
--- a/firmware/README.md
+++ b/firmware/README.md
@@ -206,11 +206,19 @@ The length of time defaults to 2 minutes. It is defined in milliseconds, so the 
 
 ### DEBUG_PUBLISH_INTERVAL
 
-This is defined via a macro in the stateMachine.h header file.
+This is defined via a macro in the debugFlags.h header file.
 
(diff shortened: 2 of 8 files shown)


trisapeace:
Task linked: [CU-860r0z9zb Shut off debug mode automatically after a certain period of time](https://app.clickup.com/t/860r0z9zb)

PR #219, opened 2023-06-30
Ps pir

trisapeace:


trisapeace:
This appears to be work done by Peter to test the PIRS sensor as an alternative to the INS. We are still using the INS, but will keep this branch just in case this is useful in the future.

PR #220, opened 2023-06-30
Ps hp testing

trisapeace:


trisapeace:
Seems to be code that Peter wrote to test the HP sensor as a potential replacement for the INS. We are still using the INS. So just keeping this PR in case it's useful later.

PR #221, opened 2023-06-30
Td mass notifications (CU-w9bcb5)

trisapeace:


trisapeace:
Task linked: [CU-w9bcb5 Add a way to send a custom message to all active responder phones in Buttons and Sensors](https://app.clickup.com/t/w9bcb5)

trisapeace:
Old branch to make the generic email/push notifications . This is now no longer useful because we are moving the Dashboard into PA and we aren't doing the Alert App, so no need for push notifications.

PR #229, opened 2023-09-08
Feature/GitHub actions

phenom5886:
- Build and deploy to Development pipeline
- Update release environment variable with commit hash as part of process

Tested using trigger on push, but now switched to be a manually executed job (can then be executed from any branch to facilitate deploying different feature branches to development). 

Manually executable jobs only work once the code is in main, and branches will need to of course be then created from main to include this code in order to use the jobs.


server/setup_postgresql.sh
```diff
@@ -1,10 +1,35 @@
 #!/bin/bash
 
-pg_password=$1
-pg_user=$2
-pg_host=$3
-pg_dbname=$4
-pg_port=$5
+
```
trisapeace, line 3: Just curious... why did this script need to change?

.github/workflows/ecs-deploy.yml
```diff
@@ -0,0 +1,80 @@
...
+          NEW_ECS_TASK_DEFINITION_API_MIGRATIONS_ARN=$(aws ecs register-task-definition --region "${{ secrets.aws_region }}" --family "brave-devices-${{ inputs.sdlc }}-api-db-migrations" --cli-input-json "${NEW_ECS_TASK_DEFINITION_API_MIGRATIONS}" | jq -r .taskDefinition.taskDefinitionArn)
+          echo "db_migrations_task_arn=${NEW_ECS_TASK_DEFINITION_API_MIGRATIONS_ARN}" >> $GITHUB_OUTPUT
+
+      - name: Run DB Migrations
+        run: |
+          echo "Running DB Migrations"
+          API_MIGRATIONS_TASK_ARN=$(aws ecs run-task --region "${{ secrets.aws_region }}" --task-definition ${{ steps.registerTaskDefinitions.outputs.db_migrations_task_arn }} --cluster "brave-devices-${{ inputs.sdlc }}-cluster" --network-configuration "{ \"awsvpcConfiguration\": { \"assignPublicIp\":\"DISABLED\", \"securityGroups\": [\"${{ secrets.db_migrations_security_group }}\"], \"subnets\": [\"${{ secrets.db_migrations_subnet }}\"]}}" --query 'tasks[].taskArn' --output text | rev | cut -d'/' -f1 | rev)
```
trisapeace, line 68: Is it possible for this step to show the output from the `setup_postgresql.sh` script? Or is the output of this script viewable somewhere else?

Right now, I just see the `echo`ed text like this:
image

trisapeace (commented):
Thanks @phenom5886 , this looks really cool and you did it up so quickly. I have just a few questions that I've added.

.github/workflows/deploy-development.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Git clone
+        uses: actions/checkout@v3
+      - name: SetupEnv
+        id: SetupEnv
+        run: |
+          echo "sdlc=${{ env.SDLC }}" >> $GITHUB_OUTPUT
+          echo "commit_sha=${GITHUB_SHA::10}" >> $GITHUB_OUTPUT
+
+  BuildContainerImage:
+    uses: bravetechnologycoop/BraveSensor/.github/workflows/ecr-build-push.yml@main
```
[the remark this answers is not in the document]
phenom5886: I think we can actually remove the @[branch] part these days, previously it was a requirement. We can test this and in theory it should just then use the version in the current branch. 

If this doesn't work then I don't think it's a big deal, these particular files shouldn't change on a regular basis and the version in main should be fine across any branch being deployed typically.

.github/workflows/ecr-build-push.yml
```diff
@@ -0,0 +1,50 @@
...
+      contents: read    # This is required for actions/checkout
+
+jobs:
+  DockerBuildPushECR:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git clone
+        uses: actions/checkout@v3
+      - name: configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v1.7.0
```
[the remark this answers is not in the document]
phenom5886: No, no reason that we can't update this to use the latest.

.github/workflows/deploy-development.yml
```diff
@@ -0,0 +1,56 @@
+name: Deploy AWS Development
```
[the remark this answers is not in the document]
phenom5886: Correct, there would be no reason for us to use CodePipeline with these GitHub Actions jobs in place. I believe these will give you greater flexibility and a better experience on the whole.

server/setup_postgresql.sh
phenom5886: To run the migrations I am now actually firing up a container on the cluster and executing this script from within it. All the DB connection information is injected as environment variables so this modification was the allow the script to read them from the environment if they are there. If they are not there, it should then expect them as input arguments as before so should continue to function in the old manner as well if required.

.github/workflows/ecs-deploy.yml
phenom5886: I can look into this - we would need to stream the logs from the container running the migrations, which is definitely possible. 

The logs would currently be visible in CloudWatch inside the target account, i.e. Development account in this instance.

trisapeace: Need to check that if this script fails, the whole pipeline should fail.

PR #231, opened 2023-09-14
Reset Stillness Timer (CU-860rbtg4k)

trisapeace:
- Create a new Particle cloud function that when called will reset the current Stillness Timer value, thus changing the state machine to wait at least another Stillness Threshold seconds before the next Stillness Alert
- Call this new cloud function whenever a Responder first responds to an alert

## Test plan
- [x] Deploy to Dev (use new AWS Dev environment)
- [x] Run smoketests
- [x] Hit the cloud function using Particle Console
   - [x] When in State 0 (idle)
      - [x] Nothing happens
      - [x] Returns -1 
   - [x] When in State 1 (initial)
      - [x] Nothing happens
      - [x] Returns -1
   - [x] When in State 2 (duration)
      - [x] Nothing happens
      - [x] Returns -1
   - [x] When in State 3 (stillness)
      - [x] When no alerts have happened yet
         - [x] Nothing happens
         - [x] Returns -1
      - [x] When a Stillness Alert has just occurred
         - [x] Can see the stillness timer value in the debug messages go to 0
         - [x] Returns the value of the stillness timer 
      - [x] When a Duration Alert has just occurred
         - [x] Can see the stillness timer value in the debug messages go to 0
         - [x] Returns the value of the stillness timer 
      - [x] When waiting for a Short Stillness Threshold
         - [x] Can see the stillness timer value in the debug messages go to 0
         - [x] Returns the value of the stillness timer 
      - [x] When waiting for the Long Stillness Threshold
- [x] Using my real device
   - [x] When in State 3 (stillness)
     - [x] Respond to a Duration Alert
        - [x] See the stillness timer value in the debug messages go to 0
        - [x] Stillness Alert occurs the right amount of time after that 
     - [x] Respond to a Short Stillness Alert 
        - [x] See the stillness timer value in the debug messages go to 0
        - [x] Stillness Alert occurs the right amount of time after that 
     - [x] Respond to a Long Stillness Alert
        - [x] See the stillness timer value in the debug messages go to 0
        - [x] Stillness Alert occurs the right amount of time after that 
  - [x] Logs the old and new stillness timer values
  - [x] Trigger a Stillness Alert, open door, second session is in state 3 (Stillness) but hasn't yet triggered a Stillness Alert, then Responder responds
     - [x] See that the stillness timer value in the debug message does NOT go to 0
  - [x] Trigger a Stillness Alert, open door, trigger a Stillness Alert in second session, be in state 3 (Stillness), then Responder responds
      - [x] See the stillness timer value in the debug messages go to 0
  - [x] Trigger a Duration Alert, open door, trigger a Stillness Alert in second session, be in state 3 (Stillness), then Responder responds
      - [x] See the stillness timer value in the debug messages go to 0
   - [x] What happens if the firmware is an older version when the Responder responds? - It logs `HTTP error 404 from https://api.particle.io/v1/products/<productId>/devices/e00fce680dcdd859b8089f31/Reset_Stillness_Timer_For_Alerting_Session : for device <deviceId>`
   - [x] Does not send more than one Duration Alert per session
- [x] When a Device that is registered on Particle Console is disconnected/offline
   - [x] Should log the error
- [x] When a device doesn't isn't registered on Particle Console (i.e. when running the smoketests)
   - [x] Should log the error

### DIFF
--- a/firmware/README.md
+++ b/firmware/README.md
@@ -433,6 +433,20 @@ Use this console function to force the boron to reset. Useful for development an
 - Nothing if 1 is entered, but does send a message to the particle console to warn about particle's future failed to call message
 - -1 - when bad data is entered
 
+### **reset_stillness_timer_for_alerting_session(String)**
+
+**Description:**
+
+Use this console function to reset the current stillness timer to the current time if and only if the current session has sent at least one alert. This acts to extend the period of time before subsequent Stillness Alerts.
+
+**Argument(s):**
+
+1. Enter any String
+
+**Return(s):**
+
+- The length of the stillness timer before it was reset, converted from a ` long` to an `int`.
+
 ## State Machine Published Messages
 
 ### **Stillness Alert**
--- a/firmware/boron-ins-fsm/src/consoleFunctions.cpp
+++ b/firmware/boron-ins-fsm/src/consoleFunctions.cpp
@@ -29,6 +29,7 @@ void setupConsoleFunctions() {
     Particle.function("Change_Long_Stillness_Timer", long_stillness_timer_set);
     Particle.function("Change_INS_Threshold", ins_threshold_set);
     Particle.function("Change_IM21_Door_ID", im21_door_id_set);
+    Particle.function("Reset_Stillness_Timer_For_Alerting_Session", reset_stillness_timer_for_alerting_session);
 }
 
 bool isValidIM21Id(String input) {
@@ -183,6 +184,18 @@ int stillness_timer_set(String input) {
     }
 
     return returnFlag;
+}
+
+// returns the previous length of the stillness timer
+int reset_stillness_timer_for_alerting_session(String input) {
+    int returnValue = -1;
+
+    if (hasDurationAlertBeenSent || hasStillnessAlertBeenSent) {
+        returnValue = (int)(millis() - state3_stillness_timer);
+        state3_stillness_timer = millis();
+    }
+
+    return returnValue;
 }
 
 // returns long stillness timer length if valid input is given, otherwise returns -1
--- a/firmware/boron-ins-fsm/src/stateMachine.cpp
+++ b/firmware/boron-ins-fsm/src/stateMachine.cpp
@@ -26,8 +26,9 @@ unsigned long state3_max_stillness_time = STATE3_MAX_STILLNESS_TIME;
 unsigned long state3_max_long_stillness_time =
     STATE3_MAX_STILLNESS_TIME;  // By default, we use the same value for max_stillness_time and max_long_stillness_time
 int resetReason = System.resetReason();
-// record whether a duration alert has been sent within the same session
+// record whether an alert has been sent within the same session
 bool hasDurationAlertBeenSent;
+bool hasStillnessAlertBeenSent;
 // which max stillness time are we currently comparing against
 // using pointers so that the value pointed to be max_stillness_time will be the correct values of state3_max_stillness_time or
 // state3_max_long_stillness_time even if they change due to a console function call during the session
@@ -49,7 +50,10 @@ void setupStateMachine() {
     stateMachineDebugFlag = 0;
 
     // default to no duration alert sent
-    hasDurationAlertBeenSent = 0;
+    hasDurationAlertBeenSent = false;
+
+    // default to no stillness alert sent
+    hasStillnessAlertBeenSent = false;
 }
 
 void initializeStateMachineConsts() {
@@ -113,8 +117,9 @@ void state0_idle() {
     checkDoor = checkIM();
     // this returns 0.0 if the INS has no new data to transmit
     checkINS = checkINS3331();
-    // session is over in idle state, so reset the whether duration alert has been sent flag
-    hasDurationAlertBeenSent = 0;
+    // session is over in idle state, so reset the whether alert has been sent flags
+    hasDurationAlertBeenSent = false;
+    hasStillnessAlertBeenSent = false;
     // set to compare against the shorter stillness threshold
     max_stillness_time = &state3_max_stillness_time;
 
@@ -226,7 +231,7 @@ void state2_duration() {
         saveStateChangeOrAlert(2, 4);
         Log.error("Duration Alert!!");
         Particle.publish("Duration Alert", "duration alert", PRIVATE);
-        hasDurationAlertBeenSent = 1;  // a duration alert has been sent, so do not send another one
+        hasDurationAlertBeenSent = true;
         stateHandler = state2_duration;
     }
     else {
@@ -272,6 +277,7 @@ void state3_stillness() {
         saveStateChangeOrAlert(3, 5);
         Log.error("Stillness Alert!!");
         Particle.publish("Stillness Alert", "stillness alert!!!", PRIVATE);
+        hasStillnessAlertBeenSent = true;
         state3_stillness_timer = millis();                     // reset the stillness timer
         max_stillness_time = &state3_max_long_stillness_time;  // set to compare against the longer stillness threshold for the rest of this session
         stateHandler = state3_stillness;
@@ -282,7 +288,7 @@ void state3_stillness() {
         saveStateChangeOrAlert(3, 4);
         Log.error("Duration Alert!!");
         Particle.publish("Duration Alert", "duration alert", PRIVATE);
-        hasDurationAlertBeenSent = 1;  // a duration alert has been sent, so do not send another one
+        hasDurationAlertBeenSent = true;
         stateHandler = state3_stillness;
     }
     else {
--- a/firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
+++ b/firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
@@ -3,8 +3,12 @@
 #include "../src/consoleFunctions.cpp"
 #include "../src/consoleFunctions.h"
 #include "../src/flashAddresses.h"
+#include "../src/stateMachine.h"
 
 IMDoorID globalDoorID = {0xAA, 0xAA, 0xAA};
+unsigned long state3_stillness_timer;
+bool hasDurationAlertBeenSent;
+bool hasStillnessAlertBeenSent;
 
 SCENARIO("Turn_Debugging_Publishes_On_Off", "[toggle debug flag]") {
     GIVEN("A false debug flag") {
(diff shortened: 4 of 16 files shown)


trisapeace:
Task linked: [CU-860rbtg4k Reset the Stillness Timer when a Responder responds](https://app.clickup.com/t/860rbtg4k)

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -437,4 +441,71 @@ SCENARIO("Force_Reset", "[force reset]") {
...
+                REQUIRE((returnVal >= expectedValue - 1000 && returnVal <= expectedValue));
+            }
+
+            THEN("The new stillness timer value is the current value of millis()") {
+                // Testing a one-second range to prevent test flakiness
+                int currentTimeInMillis = (int)millis();
+                REQUIRE((state3_stillness_timer >= currentTimeInMillis - 1000 && state3_stillness_timer <= currentTimeInMillis));
+            }
+        }
+    }
```
tmackinnon, line 491: All three tests look good - out of curiosity is it possible for both the hasDurationAlertBeenSent & hasStillnessAlertBeenSent to be true?

server/test/unit/BraveAlerterConfiguratorTest/alertSessionChangedCallbackTest.js
```diff
@@ -64,18 +65,67 @@ describe('BraveAlerterConfigurator.js unit tests: alertSessionChangedCallback',
...
+
+    const expectedSession = sessionFactory({
+      id: sessionId,
+      chatbotState: CHATBOT_STATE.WAITING_FOR_CATEGORY,
+      respondedAt: this.testCurrentTime,
+      respondedByPhoneNumber: newRespondedByPhoneNumber,
+    })
+
+    expect(particleFunctions.resetStillnessTimer).to.be.calledWith(expectedSession.location.radarCoreId, helpers.getEnvVar('PARTICLE_PRODUCT_GROUP'))
+  })
```
tmackinnon, line 96: I'm not as clear on what this is testing. Is it whether the function is being called with the appropriate arguments?

tmackinnon (commented):
Looks great Theresa! I just had one question for my own reference, to make sure I'm understanding everything properly. :)

server/test/unit/BraveAlerterConfiguratorTest/alertSessionChangedCallbackTest.js
trisapeace: This is testing the condition that we actually call the Particle Cloud function to reset the Stillness Timer.

The `alertSessionChangedCallback` is a function that is called by Brave Alert Lib whenever there is a chatbot interaction. As a parameter, it sends the and `AlertSession` object containing information about what the new state of the session is. Then the `alertSessionChangedCallback` function looks at the given `AlertSession` object to figure out what it wants to do.

What we really want to do is call `particleFunctions.resetStillnessTimer` when the responder texts us "OK". But the way we have to do that is by looking at the properties of the `AlertSession` object that is passed to `alertSessionChangedCallback`.

When a Responder texts "OK" we are either moving from `CHATBOT_STATE.STARTED` or `CHATBOT_STATE.WAITING_FOR_REPLY` to `CHATBOT_STATE.WAITING_FOR_CATEGORY`. And the `AlertSession` object always contains the new properties, not the old ones. So our trigger will be when it sends us `CHATBOT_STATE_WAITING_FOR_CATEGORY`. And we only want to reset the stillness timer the *first* time that the session is in `WAITING_FOR_CATEGORY`. 

However, it is possible to transition to `WAITING_FOR_CATEGORY` more than once in the same session by responding with an invalid number when asked to categorize the alert. So we look at the Session object returned from the DB to see if it already had a `respondedAt` or `respondedByPhoneNumber` field. If these are `null` or `undefined` then we know that this is the first time that this session is in the `WAITING_FOR_CATEGORY` state.

This is kind of what I was saying about how using Brave Alert Lib ended up making things more complicated than they necessarily needed to be. Please let me know if this still doesn't make sense!

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -437,4 +441,71 @@ SCENARIO("Force_Reset", "[force reset]") {
...
+        hasStillnessAlertBeenSent = false;
+
+        WHEN("the function is called with any string") {
+            int returnVal = reset_stillness_timer_for_alerting_session("any string");
+
+            THEN("The function should return -1") {
+                REQUIRE(returnVal == -1);
+            }
+
+            THEN("The stillness timer value did not change") {
```
ahmaddaadaa, line 506: I'd suggest opting for AND_THEN over AND in this situation because it conveys that we're dealing with two assertions within the same test structure.

server/Session.js
```diff
@@ -1,14 +1,14 @@
...
     this.chatbotState = chatbotState
     this.alertType = alertType
     this.createdAt = createdAt
     this.updatedAt = updatedAt
     this.incidentCategory = incidentCategory
     this.respondedAt = respondedAt
     this.respondedByPhoneNumber = respondedByPhoneNumber
+    this.location = location
   }
 }
```
ahmaddaadaa, line 13: the class definition appears to be correct

server/db/db.js
```diff
@@ -150,6 +152,32 @@ async function getClients(pgClient) {
...
+    )
+
+    if (results === undefined) {
+      return null
+    }
+
+    const allClients = await getClients(pgClient)
+    return results.rows.map(r => createLocationFromRow(r, allClients))
+  } catch (err) {
+    helpers.log(err.toString())
```
ahmaddaadaa, line 177: I would suggest logging the entire error using helper.log(err) instead of its string representation. This would provide us with more detailed error information

ahmaddaadaa (commented):
Nice work, the code changes looks clear and well organized. I added a couple of minor suggestions. 

It would be really interesting to observe its behaviors in a real situation.
Thank you

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
trisapeace: Oh cool. I didn't know about the `AND_THEN` syntax here. But looking at this documentation (https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#bdd-style-test-cases) , I don't think that `AND_THEN` is what I want to use. 

If I were to use `AND_THEN` then this would only run this single test:
```
GIVEN("A session that has not yet sent any alerts")
WHEN("the function is called with any string")
THEN("The function should return -1")
AND_THEN("The stillness timer value did not change")
```

but what I really want is for this to run two different tests with two different `REQUIRE` assertion:
```
GIVEN("A session that has not yet sent any alerts")
WHEN("the function is called with any string")
THEN("The function should return -1")
```
and
```
GIVEN("A session that has not yet sent any alerts")
WHEN("the function is called with any string")
THEN("The stillness timer value did not change")
```

It is generally good practice to only have one assertion per test. This is because if a test fails, then it doesn't execute the rest of the test code. So in this case, if the function **didn't** return -1, then it would fail there and wouldn't even run the code where we check to see if the stillness timer value changed or not. This gives us less information than if they were executed as two separate tests.

server/db/db.js
trisapeace: This is a good thing to experiment with in the future. Right now, this is basically boilerplate code that is used for all the database functions in this file and in the Button's `db.js` as well. So I don't want to change it. But I could see a project in the future to examine exactly what we want to log and then changing it for all the functions if that's what you decide to do.

Some things to consider:
1. I think that `helpers.log` takes a string value. So if you just send `helpers.log(err)` then it will just output `[Object object]` which is less helpful than `err.toString()`
2. Sometimes the full error contains information that would be a security risk if it were in the logs. For example, the DB password. If this error object contains the DB password in the logs, and that was sent to Sentry and stored on the server logs in Digital Ocean (or AWS) then that means it now exists in at least 3 places. Therefore, we are three times as likely to get hacked and our password compromised. 

So I agree that more information can be useful to us. This is outside the scope of this PR for the moment. If this is something that you'd like to look into, it would be cool if you make a ClickUp Task in the Backlog of the Devices Kanban, and it can be prioritized along with the other Tasks at our next Prioritization Meeting. Thanks!

trisapeace (commented):
Thanks for your suggestions @ahmaddaadaa ! I responded to them. Please let me know if you disagree with anything or if you have any other questions.

server/test/unit/BraveAlerterConfiguratorTest/alertSessionChangedCallbackTest.js
tmackinnon: Thanks for explaining Theresa! I think I understand now. 

So your test is making sure that `particleFunctions.resetStillnessTimer` is getting called when the chatbot_state is changed to `CHATBOT_STATE_WAITING_FOR_CATEGORY` for the _first_ time.

trisapeace: Exactly!

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -49,7 +50,10 @@ void setupStateMachine() {
     stateMachineDebugFlag = 0;
 
     // default to no duration alert sent
-    hasDurationAlertBeenSent = 0;
+    hasDurationAlertBeenSent = false;
+
+    // default to no stillnss alert sent
```
micahdbak, line 55: small typo: `stillnss` -> `stillness`

server/test/unit/ParticleFunctionsTest/resetStillnessTimerTest.js
```diff
@@ -0,0 +1,164 @@
...
+      expect(this.response).to.equal(-1)
+    })
+
+    it('should log the error', () => {
+      expect(helpers.log).to.be.calledWith(
+        `HTTP error 404 from https://api.particle.io/v1/products/12345/devices/radar_coreID/Reset_Stillness_Timer_For_Alerting_Session : for device ${this.deviceId}`,
+      )
+    })
+  })
+})
```
micahdbak, line 164: Well written test cases! I believe it covers all possibilities.

server/testingHelpers.js
```diff
@@ -94,14 +94,14 @@ function sessionFactory(overrides = {}) {
...
     overrides.chatbotState !== undefined ? overrides.chatbotState : CHATBOT_STATE.COMPLETED,
     overrides.alertType !== undefined ? overrides.alertType : ALERT_TYPE.SENSOR_STILLNESS,
     overrides.createdAt !== undefined ? overrides.createdAt : new Date('2021-10-05T20:20:20.000Z'),
     overrides.updatedAt !== undefined ? overrides.updatedAt : new Date('2021-10-05T20:20:55.000Z'),
     overrides.incidentCategory !== undefined ? overrides.incidentCategory : 'Overdose',
     overrides.respondedAt !== undefined ? overrides.respondedAt : new Date('2021-10-05T20:20:33.000Z'),
     overrides.respondedByPhoneNumber !== undefined ? overrides.respondedByPhoneNumber : '+17778886666',
+    overrides.location !== undefined ? overrides.location : locationFactory(),
   )
 }
```
micahdbak, line 107: I assume the change to the `Session` class to include more information on the location than just the id is necessary for the `resetStillnessTimer` to work? Could you explain which fields are needed?

micahdbak (commented):
Looks good! Just have a couple questions and one typo. Apart from that, the test plan looks thorough and this appears to be well implemented! 😊

firmware/boron-ins-fsm/src/stateMachine.cpp
trisapeace: Thanks! Fixed!

server/testingHelpers.js
trisapeace: Yes, you're right! I did need some more information on the location. Specifically, I needed the `radarCoreId`. In `BraveAlerterConfigurator`, the code is now:
```
           const oldStillnessTimer = await particleFunctions.resetStillnessTimer(
              session.location.radarCoreId,
              helpers.getEnvVar('PARTICLE_PRODUCT_GROUP'),
            )
            if (oldStillnessTimer > -1) {
              helpers.log(`Reset stillness timer for ${session.location.locationid} from ${oldStillnessTimer} to 0`)
            } else {
              helpers.log(`Did not reset stillness timer for ${session.location.locationid}`)
            }
```

Instead of changing the `Session` class, I could have just added a new DB call in `BraveAlerterConfigurator` to `db.getLocationData(session.locationId)` and that would have worked too and required fewer code changes.

However, I've been slowly moving our DB model classes over to this pattern of having the full referenced object retrieved and not just IDs. You'll see how the `Location` class in `Location.js` as a whole `client`, not just its ID. And same for `SensorsVital`. Similarly, the `Session` class in Buttons has the whole `button` object, not just its id.

It's a tradeoff. This style causes more DB calls than is probably necessary, but makes it easier to read and write the JS code. A lot of frameworks and languages offer something along these lines. We just do it ourselves, but things like Django for Python do it (and migration scripts) under the hood. It falls under the idea of object-relational-mapper (ORM) if you'd like to learn more (https://www.theserverside.com/definition/object-relational-mapping-ORM).

micahdbak: Cool! No problem with that. Thanks for explaining!

micahdbak (approved):
I approve this PR!

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -185,6 +186,18 @@ int stillness_timer_set(String input) {
     return returnFlag;
 }
 
+// returns the previous length of the stillness timer
+int reset_stillness_timer_for_alerting_session(String input) {
```
johnny-brave, line 190: I wasn't aware that we are adding a cloud function for resetting the stillness timer. Is this just the easiest way for the server to tell the firmware to zero the `state3_stillness_timer` variable?

firmware/README.md
```diff
@@ -433,6 +433,20 @@ Use this console function to force the boron to reset. Useful for development an
...
+
+Use this console function to reset the current stillness timer to the current time if and only if the current session has sent at least one alert. This acts to extend the period of time before subsequent Stillness Alerts.
+
+**Argument(s):**
+
+1. Enter any String
+
+**Return(s):**
+
+- The length of the stillness timer before it was reset, converted from a ` long` to an `int`.
```
johnny-brave, line 448: this does return a negative number if the stillness timer runs longer than ~24 days, but I don't think that will be an issue 😆

johnny-brave (approved):
Looks good! No real issues, just some comments/questions. I'm not very experienced with JS, so I will leave that mostly to the others.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -185,6 +186,18 @@ int stillness_timer_set(String input) {
...
+    int returnValue = -1;
+
+    if (hasDurationAlertBeenSent || hasStillnessAlertBeenSent) {
+        returnValue = (int)(millis() - state3_stillness_timer);
+        state3_stillness_timer = millis();
```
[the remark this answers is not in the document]
trisapeace: Yeah, I'm not a huge fan of the variable names either. But this seemed most consistent with the ones that were already there.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
trisapeace: As far as I know, using a cloud function via the API is the only way that we can communicate from the server to the firmware.

firmware/README.md
trisapeace: Good point! That is a pretty short time. I don't really have a good concept of how long our Sensors manage to stay online at a time. At least in this case, we don't do anything important with the value on the server side. It's just logged for our information. So if we see a negative number, maybe looking in the docs and seeing that there was a conversion would be enough to give the idea that this might be happening.

johnny-brave: the state3 timer is reset after a stillness alert is triggered, so it won't come up unless the timeout is >24 days

trisapeace: Oh yes, that would be weird. I guess if we had a client who insisted that they only wanted a single stillness alert per session, then we could set the Long Stillness Timer to a really big number and then it might happen. But I think we would discourage that, lol.

PR #232, opened 2023-09-18
Timeout "Window" in state 0 for allowable entrance into state 1 (CU-8678t4ztg)

johnny-brave:
Adding an additional entrance condition from state0 to state1.
- if the state0 timer exceeds a maximum timeout window, the firmware cannot enter state1
- this timer is reset when the paired door sensor switches from an open to a closed status
- the default value for the timeout will be 1 hour for uninitialized firmware
- a particle console function can be used to change the timeout value
- the console publish message will contain the current timer count for state 0 in `timer_status` along with the timeout in `window_size`

### Test Plan
___

- [x] Deploy firmware on test boron (with no memory of the window size) and connect to an INS radar
- [x] Use particle console to echo the initial value of the timeout and confirm that it is "3600"
- [x] Enter 10 in the function and confirm that it returns 10, echo again and confirm that it is still 10
- [x] View debug messages and confirm that the `window_size` reads `10000` and that the `timer_status` is increasing
- [x] Set the `intial_timer` to 30s
- [x] Open and close the paired door sensor and confirm that `timer_status` is reset
- [x] Physically shake the sensor to enter state1, after `timer_status` exceeds `window_size` stop shaking to return to state0, shake again to confirm that the sensor can no longer enter state1 
- [x] Set the `intial_timer` to 3s
- [x] Open and close the door sensor again, shake the sensor for 3s while within the 10s window and confirm that it enters state 2/3
- [x] Reset the sensor and echo the window size, confirming that it has remembered the 10s value

Oct 11th, 2024:

Ensure that the timer resets in these cases:
- [x] Open -> Missed Close -> Heartbeat + Closed
- [x] Closed -> Missed Open -> Close

Ensure that the timer does NOT reset in these cases:
- [x] Closed -> Heartbeat + Closed
- [x] Heartbeat + Closed -> Heartbeat + Closed
- [x] Unknown -> Heartbeat + Closed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ the code was deployed.
 ### Added
 
 - CORS configuration to Express Proxy Middleware.
+- Added a window of time after the door closes where the sensor is allowed to detect for occupation
+- Added console function for the window aswell as timer status and window size in Debug Publishes
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,14 @@ the code was deployed.
 ### Added
 
 - CORS configuration to Express Proxy Middleware.
-- Added a window of time after the door closes where the sensor is allowed to detect for occupation (CU-8678t4ztg)
-- Added console function for the window aswell as timer status and window size in Debug Publishes (CU-8678t4ztg)
-- "doorMissedFrequently" field to JSON data submitted by brave sensors that is sent to /api/heartbeat (CU-860rk8v2a)
-- Sentry log in the case that doorMissedFrequently is true in posted data from brave sensor (CU-860rk8v2a)
+- Added a timer in state0 that is reset when the door closes where the sensor is allowed to detect for occupation (CU-8678t4ztg).
+- Added console function for the occupation detection timer as well as the timer count and timeout value in Debug Publishes (CU-8678t4ztg).
+- "doorMissedFrequently" field to JSON data submitted by brave sensors that is sent to /api/heartbeat (CU-860rk8v2a).
+- Sentry log in the case that doorMissedFrequently is true in posted data from brave sensor (CU-860rk8v2a).
+
+### Fixed
+
+- Wording in consoleFunctionTests.cpp when functions are called with 'e'.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ the code was deployed.
 ### Added
 
 - CORS configuration to Express Proxy Middleware.
-- Added a timer in state0 that is reset when the door closes where the sensor is allowed to detect for occupation (CU-8678t4ztg).
+- Added a timer in state0 that is reset when the door closes during which the sensor is allowed to detect for occupation (CU-8678t4ztg).
 - Added console function for the occupation detection timer as well as the timer count and timeout value in Debug Publishes (CU-8678t4ztg).
 - "doorMissedFrequently" field to JSON data submitted by brave sensors that is sent to /api/heartbeat (CU-860rk8v2a).
 - Sentry log in the case that doorMissedFrequently is true in posted data from brave sensor (CU-860rk8v2a).
--- a/firmware/README.md
+++ b/firmware/README.md
@@ -12,7 +12,7 @@
      - [DEBUG_LEVEL](#debug_level)
      - [BRAVE_PRODUCT_ID for state machine](#brave_product_id-for-state-machine)
      - [INS_THRESHOLD](#ins_threshold)
-     - [STATE0_WINDOW_TIME](#state0_window_time)
+     - [STATE0_OCCUPANT_DETECTION_TIMER](#state0_occupant_detection_timer)
      - [STATE1_MAX_TIME](#state1_max_time)
      - [STATE2_MAX_DURATION](#state2_max_duration)
      - [STATE3_MAX_STILLNESS_TIME](#state3_max_stillness_time)
@@ -181,11 +181,11 @@ It is compared to the filtered inPhase values detected by the INS radar. Anythin
 
 The default level is set to 60. This is based on the radar testing documented [here](https://docs.google.com/document/d/12TLw6XE9CSaNpguytS2NCSCP0aZWUs-OncmaDdoTKfo/edit?usp=sharing).
 
-### STATE0_WINDOW_TIME
+### STATE0_OCCUPANT_DETECTION_TIMER
 
 This is defined via a macro in the stateMachine.h header file
 
-It is the length of time (or window) after the door_status changes from an open to a closed state where entrance to state1 is allowed. After the state 0 timer has surpassed this window, the state machine with still in state 0 until this timer reset (door open and close) 
+It is the length of time after the door_status changes from an open to a closed state where entrance to state1 is allowed. After the state 0 timer has surpassed this value, the state machine will stay in state 0 until this timer resets (door open and close) 
 
 The length of time is default to 1 hour defined as 3600000 milliseconds
 
@@ -332,21 +332,21 @@ Stillness timer is the length of time the Sensor sees stillness before publishin
 - The integer number of seconds of the current timer value, when e for echo is entered
 - -1: when bad data is entered
 
-### **window_size_set(String)**
+### **occupant_detection_timer_set(String)**
 
 **Description:**
 
-Window is the allowed time after the paired door sensor closes in which state 0 can transition to state 1. The default value is 60 minutes. Use this console function to set the window size (time) 
+This sets the allowed time after the paired door sensor closes in which state 0 can transition to state 1. The default value is 60 minutes.
 
 **Argument(s):**
 
-1. The integer number of seconds of the new window timeout.
-2. e - this is short for echo, and will echo the current window timeout
+1. The integer number of seconds of the new occupation detection timer.
+2. e - this is short for echo, and will echo the current occupation detection timer
 
 **Return(s):**
 
-- The integer number of seconds of the timeout value, when a new timeout value is entered
-- The integer number of seconds of the current timeout value, when e for echo is entered
+- The integer number of seconds of the timer value, when a new timer value is entered
+- The integer number of seconds of the current timer value, when e for echo is entered
 - -1: when bad data is entered
 
 ### **initial_timer_set(String)**
--- a/firmware/README.md
+++ b/firmware/README.md
@@ -12,6 +12,7 @@
      - [DEBUG_LEVEL](#debug_level)
      - [BRAVE_PRODUCT_ID for state machine](#brave_product_id-for-state-machine)
      - [INS_THRESHOLD](#ins_threshold)
+     - [STATE0_WINDOW_TIME](#state0_window_time)
      - [STATE1_MAX_TIME](#state1_max_time)
      - [STATE2_MAX_DURATION](#state2_max_duration)
      - [STATE3_MAX_STILLNESS_TIME](#state3_max_stillness_time)
@@ -179,6 +180,14 @@ This is defined via a macro in the stateMachine.h header file.
 It is compared to the filtered inPhase values detected by the INS radar. Anything above the threshold is considered movement, and anything below the threshold is considered stillness or an empty room.
 
 The default level is set to 60. This is based on the radar testing documented [here](https://docs.google.com/document/d/12TLw6XE9CSaNpguytS2NCSCP0aZWUs-OncmaDdoTKfo/edit?usp=sharing).
+
+### STATE0_WINDOW_TIME
+
+This is defined via a macro in the stateMachine.h header file
+
+It is the length of time (or window) after the door_status changes from an open to a closed state where entrance to state1 is allowed. After the state 0 timer has surpassed this window, the state machine with still in state 0 until this timer reset (door open and close) 
+
+The length of time is default to 1 hour defined as 3600000 milliseconds
 
 ### STATE1_MAX_TIME
 
--- a/firmware/README.md
+++ b/firmware/README.md
@@ -556,7 +556,7 @@ Debug Message
 1. INS_val: the current filtered inPhase value from the INS radar
 1. INS_threshold: the current threshold that we compare the INS_val against
 1. timer_status: if the current state uses a timer, this contains the time in milliseconds that the timer has counted up to thus far. If the current state does not contain a timer, this is set to 0.
-1. window_size: the window timeout value that that the idle state timer compares against
+1. occupation_detection_timer: the occupation detection timer value that that the idle state compares against
 1. initial_timer: the initial timer value that the initial state compares against
 1. duration_timer: the duration timer value that the duration and stillness states compare against
 1. stillness_timer: the stillness timer value that the stillness state compares against
--- a/firmware/boron-ins-fsm/src/BraveSensorProductionFirmware.ino
+++ b/firmware/boron-ins-fsm/src/BraveSensorProductionFirmware.ino
@@ -11,7 +11,7 @@
 #include "tpl5010watchdog.h"
 
 #define DEBUG_LEVEL            LOG_LEVEL_INFO
-#define BRAVE_FIRMWARE_VERSION 9060   // see versioning notes in the readme
+#define BRAVE_FIRMWARE_VERSION 3898   // see versioning notes in the readme
 #define BRAVE_PRODUCT_ID       14807  // 14807 = beta units, 15479 = production units
 
 PRODUCT_ID(BRAVE_PRODUCT_ID);             // you get this number off the particle console, see readme for instructions
--- a/firmware/boron-ins-fsm/src/consoleFunctions.cpp
+++ b/firmware/boron-ins-fsm/src/consoleFunctions.cpp
@@ -23,6 +23,7 @@ void setupConsoleFunctions() {
     // particle console function declarations, belongs in setup() as per docs
     Particle.function("Force_Reset", force_reset);
     Particle.function("Turn_Debugging_Publishes_On_Off", toggle_debugging_publishes);
+    Particle.function("Change_Occupant_Detection_Window", window_size_set);
     Particle.function("Change_Initial_Timer", initial_timer_set);
     Particle.function("Change_Duration_Timer", duration_timer_set);
     Particle.function("Change_Stillness_Timer", stillness_timer_set);
@@ -78,6 +79,40 @@ int toggle_debugging_publishes(String command) {
         returnFlag = -1;
     }
 
+    return returnFlag;
+}
+
+// returns initial timer length if valid input is given, otherwise returns -1
+int window_size_set(String input) {
+    int returnFlag = -1;
+
+    const char* holder = input.c_str();
+
+    // if e, echo the current threshold
+    if (*holder == 'e') {
+        EEPROM.get(ADDR_STATE0_WINDOW_TIME, state0_window_time);
+        returnFlag = state0_window_time / 1000;
+    }
+    // else parse new threshold
+    else {
+        int timeout = input.toInt();
+        // increase timeout value to from seconds to ms
+        timeout = timeout * 1000;
+
+        if (timeout == 0) {
+            // string.toInt() returns 0 if input not an int
+            // and a threshold value of 0 makes no sense, so return -1
+            returnFlag = -1;
+        }
+        else if (timeout < 0) {
+            returnFlag = -1;
+        }
+        else {
+            EEPROM.put(ADDR_STATE0_WINDOW_TIME, timeout);
+            state0_window_time = timeout;
+            returnFlag = state0_window_time / 1000;
+        }
+    }
     return returnFlag;
 }
 
--- a/firmware/boron-ins-fsm/src/consoleFunctions.cpp
+++ b/firmware/boron-ins-fsm/src/consoleFunctions.cpp
@@ -23,7 +23,7 @@ void setupConsoleFunctions() {
     // particle console function declarations, belongs in setup() as per docs
     Particle.function("Force_Reset", force_reset);
     Particle.function("Turn_Debugging_Publishes_On_Off", toggle_debugging_publishes);
-    Particle.function("Change_Occupant_Detection_Window", window_size_set);
+    Particle.function("Change_Occupant_Detection_Timer", occupant_detection_timer_set);
     Particle.function("Change_Initial_Timer", initial_timer_set);
     Particle.function("Change_Duration_Timer", duration_timer_set);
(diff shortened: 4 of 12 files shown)


trisapeace:
Task linked: [CU-8678t4ztg State Machine Adjustment: Entrance to State 1 only within a time window of closing the door](https://app.clickup.com/t/8678t4ztg)

firmware/boron-ins-fsm/src/BraveSensorProductionFirmware.ino
```diff
@@ -11,7 +11,7 @@
 #include "tpl5010watchdog.h"
 
 #define DEBUG_LEVEL            LOG_LEVEL_INFO
-#define BRAVE_FIRMWARE_VERSION 9060   // see versioning notes in the readme
+#define BRAVE_FIRMWARE_VERSION 3898   // see versioning notes in the readme
```
trisapeace, line 14: This change isn't actually related to the feature that you're implementing, so before this PR is merged, could you please change it back to 9060?

firmware/README.md
```diff
@@ -180,6 +181,14 @@ It is compared to the filtered inPhase values detected by the INS radar. Anythin
 
 The default level is set to 60. This is based on the radar testing documented [here](https://docs.google.com/document/d/12TLw6XE9CSaNpguytS2NCSCP0aZWUs-OncmaDdoTKfo/edit?usp=sharing).
 
+### STATE0_WINDOW_TIME
+
+This is defined via a macro in the stateMachine.h header file
+
+It is the length of time (or window) after the door_status changes from an open to a closed state where entrance to state1 is allowed. After the state 0 timer has surpassed this window, the state machine with still in state 0 until this timer reset (door open and close)
```
micahdbak, line 188: "With still in" should this be "will stay in"?

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -23,6 +23,7 @@ void setupConsoleFunctions() {
     // particle console function declarations, belongs in setup() as per docs
     Particle.function("Force_Reset", force_reset);
     Particle.function("Turn_Debugging_Publishes_On_Off", toggle_debugging_publishes);
+    Particle.function("Change_Occupant_Detection_Window", window_size_set);
```
trisapeace, line 26: I like this name! Maybe use `occupant_detection_timer` or something like that for the variables.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -81,6 +82,40 @@ int toggle_debugging_publishes(String command) {
     return returnFlag;
 }
 
+// returns initial timer length if valid input is given, otherwise returns -1
```
trisapeace, line 85: This comment still talks about `initial time`. I think it just needs to be updated.

firmware/boron-ins-fsm/src/BraveSensorProductionFirmware.ino
micahdbak, line 14: Is this version correct?

firmware/boron-ins-fsm/src/consoleFunctions.cpp
trisapeace: Although, maybe this is an indicated of copy-and-paste code. Is this a good place to factor out a function for cloud functions that update timer values? Then `window_size_set`, `initial_timer_set`, etc can all call that function. What do you think?

firmware/boron-ins-fsm/src/flashAddresses.h
```diff
@@ -34,6 +34,10 @@
 #define ADDR_STATE3_MAX_LONG_STILLNESS_TIME                 23  // uint32_t = 4 bytes
 #define ADDR_INITIALIZE_STATE3_MAX_LONG_STILLNESS_TIME_FLAG 27  // uint16_2 = 2 bytes
 
-// next available address is 27 + 2 = 29
+// new state machine constant for state0 window
+#define ADDR_INITIALIZE_STATE0_WINDOW_FLAG                  29  // uint16_2 = 2 bytes
+#define ADDR_STATE0_WINDOW_TIME                             31  // uint32_t = 4 bytes
+
+// next available address is 31 + 4 = 35
```
trisapeace, line 41: Thanks for remembering to update this! Your future self will thank you!

firmware/boron-ins-fsm/src/imDoorSensor.cpp
```diff
@@ -69,6 +70,11 @@ doorData checkIM() {
         // Checks if the 0th bit (counting from 0) of doorStatus is 1
         doorTamperedFlag = (currentDoorData.doorStatus & 0b0001) != 0;
 
+        // Set timer starting when door closes
+        if ((previousDoorData.doorStatus & 0b0010) != 0 && (currentDoorData.doorStatus & 0b0010) == 0) {
```
trisapeace, line 74: Looks good!

firmware/boron-ins-fsm/src/imDoorSensor.cpp
```diff
@@ -69,6 +70,11 @@ doorData checkIM() {
...
+            timeSinceDoorClosed = millis();
```
trisapeace, line 75: Yup, using `millis()` is exactly how I would do this too. It sets the value to **the time at which the door closed**.

However, when I read the variable name `timeSinceDoorClosed` I would have expected this to set it to 0. The door just closed; therefore, **the time since the door closed** is 0 milliseconds.

So I think this variable could have a clearer name to reduce the likelihood of confusion.

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -95,6 +98,21 @@ void initializeStateMachineConsts() {
...
+    if (initializeState0WindowFlag != INITIALIZE_STATE0_WINDOW_FLAG) {
+        EEPROM.put(ADDR_STATE0_WINDOW_TIME, state0_window_time);
+        initializeState0WindowFlag = INITIALIZE_STATE0_WINDOW_FLAG;
+        EEPROM.put(ADDR_INITIALIZE_STATE0_WINDOW_FLAG, initializeState0WindowFlag);
+        Log.info("State machine constant State3MaxLongStillnessTime was written to flash on bootup.");
+    }
+    else {
+        EEPROM.get(ADDR_STATE0_WINDOW_TIME, state0_window_time);
+        Log.info("State machine constant State3MaxLongStillnessTime was read from flash on bootup.");
+    }
```
trisapeace, line 115: Looks good!

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -126,10 +144,10 @@ void state0_idle() {
 
     Log.info("You are in state 0, idle: Door status, iAverage = 0x%02X, %f", checkDoor.doorStatus, checkINS.iAverage);
     // default timer to 0 when state doesn't have a timer
-    publishDebugMessage(0, checkDoor.doorStatus, checkINS.iAverage, 0);
+    publishDebugMessage(0, checkDoor.doorStatus, checkINS.iAverage, (millis() - timeSinceDoorClosed));
 
     // fix outputs and state exit conditions accordingly
-    if (((unsigned long)checkINS.iAverage > ins_threshold) && !isDoorOpen(checkDoor.doorStatus) && !isDoorStatusUnknown(checkDoor.doorStatus)) {
+    if (((unsigned long)checkINS.iAverage > ins_threshold) && !isDoorOpen(checkDoor.doorStatus) && !isDoorStatusUnknown(checkDoor.doorStatus) && millis() - timeSinceDoorClosed < state0_window_time) {
```
trisapeace, line 150: While you're in here, there is an optimization that can be made. What do you think?

[hint: C++ (like many other languages) does short-circuit evaluation (https://www.geeksforgeeks.org/short-circuiting-in-c-and-linux/). ]

firmware/boron-ins-fsm/src/stateMachine.h
```diff
@@ -12,13 +12,15 @@
...
+#define INITIALIZE_STATE0_WINDOW_FLAG                  0x8888
 
 // initial (default) values for state machine, can be changed via console function
 // or by writing something other than 0x8888 to the above flag in flash
-#define INS_THRESHOLD             60
-#define STATE1_MAX_TIME           15000    // ms = 15s
-#define STATE2_MAX_DURATION       1200000  // ms = 20 min
-#define STATE3_MAX_STILLNESS_TIME 120000   // ms = 2 minutes
+#define INS_THRESHOLD                   60
+#define STATE0_WINDOW_TIME              3600000  // ms = 60 min
```
trisapeace, line 20: From what we talked about earlier, I was expecting a higher default value. Why did you end up choosing 60 minutes?

micahdbak (commented):
Everything looks good! Looks like Travis failed the code formatting, though. If you run `python3 clang-format-all.py` from `firmware/boron-ins-fsm`, it should resolve all of those errors.

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -143,6 +143,60 @@ SCENARIO("Change_IM21_Door_ID", "[change door id]") {
...
+        state0_window_time = 60000;
+
+        WHEN("the function is called with 'e'") {
+            int returnFlag = window_size_set("e");
+
+            THEN("the initial timer value should remain the same") {
+                REQUIRE(state0_window_time == 60000);
+            }
+
+            THEN("the function should return the input") {
```
trisapeace, line 157: I think the description of this "THEN" should be updated. When I read "the function should return the input", it sounds like it was supposed to return "e", which is not the case.

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -143,6 +143,60 @@ SCENARIO("Change_IM21_Door_ID", "[change door id]") {
     }
 }
 
+SCENARIO("Set Window Time", "[window time]") {
```
trisapeace, line 146: Fantastic! Thanks for updating the tests!

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ the code was deployed.
 ### Added
 
 - CORS configuration to Express Proxy Middleware.
+- Added a window of time after the door closes where the sensor is allowed to detect for occupation (CU-8678t4ztg)
```
trisapeace, line 17: Awesome! Thanks for remembering to update the CHANGELOG!

firmware/boron-ins-fsm/src/flashAddresses.h
```diff
@@ -34,6 +34,10 @@
 #define ADDR_STATE3_MAX_LONG_STILLNESS_TIME                 23  // uint32_t = 4 bytes
 #define ADDR_INITIALIZE_STATE3_MAX_LONG_STILLNESS_TIME_FLAG 27  // uint16_2 = 2 bytes
 
-// next available address is 27 + 2 = 29
+// new state machine constant for state0 window
+#define ADDR_INITIALIZE_STATE0_WINDOW_FLAG                  29  // uint16_2 = 2 bytes
```
trisapeace, line 38: Looks like your changes are failing on Travis due to linting errors: https://app.travis-ci.com/github/bravetechnologycoop/BraveSensor/jobs/610053932

Take a look in the README on how to run the `clang` formatting scripts. There's one that will give you the list of problems and another that will actual fix the problems for you. You should be able to run this on your computer.

firmware/README.md
```diff
@@ -323,6 +332,23 @@ Stillness timer is the length of time the Sensor sees stillness before publishin
...
 
+### **window_size_set(String)**
+
+**Description:**
+
+Window is the allowed time after the paired door sensor closes in which state 0 can transition to state 1. The default value is 60 minutes. Use this console function to set the window size (time) 
+
+**Argument(s):**
+
+1. The integer number of seconds of the new window timeout.
```
trisapeace, line 343: This description in the README is correct. But I notice that you have this test in your test plan:

"Enter 10 in the function and confirm that it returns 60, echo again and confirm that it is still 10".

When you enter 10 in the function, I would have expected it to return 10, not 60. How did this test pass?

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -126,10 +144,10 @@ void state0_idle() {
 
     Log.info("You are in state 0, idle: Door status, iAverage = 0x%02X, %f", checkDoor.doorStatus, checkINS.iAverage);
     // default timer to 0 when state doesn't have a timer
-    publishDebugMessage(0, checkDoor.doorStatus, checkINS.iAverage, 0);
+    publishDebugMessage(0, checkDoor.doorStatus, checkINS.iAverage, (millis() - timeSinceDoorClosed));
```
trisapeace, line 147: Thanks for updating the last argument to this function. I think that will be helpful for debugging.

trisapeace (commented):
Looking great!!! You got this done really quickly. Thanks for putting in your test plan and trying everything out.

I've got a bunch of comments and questions, but please don't be worried about that. I pretty much always write a lot in code reviews. The purpose is to not only help you learn to write better code, but also about how to think about different problems that aren't always technical. Please let me know if anything doesn't make sense, or you disagree with me. This is a good place for discussion or on a call if that's easier.

If I haven't sent this to you already, I love this presentation on code review culture. I highly recommend it: https://www.youtube.com/watch?v=PJjmw9TRB7s

firmware/README.md
johnny-brave: yes, thanks for the catch

firmware/boron-ins-fsm/src/stateMachine.h
johnny-brave: I don't think there is much risk in immediately implementing a 1 hour timeout on all sensors to alleviate all middle of the night alerts (acting basically as the "hours of operation" fix), without really changing how production sensors are operating currently. Do you think this would be too risky?

firmware/README.md
johnny-brave: This was a typo in the test, it should return 10

firmware/boron-ins-fsm/src/stateMachine.cpp
johnny-brave: I see, so I should move `millis() - timeSinceDoorClosed < state0_window_time` to the first check since it should be the most likely to fail?

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
johnny-brave: I agree. I'll change all of the other instances of this in other SCENERIOS as well

firmware/boron-ins-fsm/src/consoleFunctions.cpp
johnny-brave: I see what you mean, it would be something like `int set_timer(uint32_t flashAddress, uint32_t* state_timer_variable, String input)`, but I think that'll be a seperate PR.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
johnny-brave: yep, that makes more sense, I'll change the other variable names

firmware/boron-ins-fsm/src/imDoorSensor.cpp
```diff
@@ -18,6 +18,7 @@ bool doorTamperedFlag = false;
 bool doorMessageReceivedFlag = false;
 unsigned long doorHeartbeatReceived = 0;
 unsigned long doorLastMessage = 0;
+unsigned long timeWhenDoorClosed = 0;
```
trisapeace, line 21: Can you test what happens when you restart the Sensor while the door sensor is closed?

I expect the first debug messages will have the 0x99 value.
Then after the first heartbeat, it will realize that the door is closed.
Does this realization reset the `timeWhenDoorClosed` or not? Do we want it to or not?

trisapeace (commented):
Thanks for making the changes. Here are a few more. You're almost there!

firmware/boron-ins-fsm/src/stateMachine.h
johnny-brave: No, I believe that was Duration alerts repeating. For Edmonton, I plan to decrease the state0 timer even further to maybe 10 minutes and reduce the initial timer to 5s.

People can definitely be occupying the bathroom for more than 1 hour, but if the sensor hasn't picked them up by an hour, I don't think it will in the next hour either. I would prefer to start with 1 hour.

CHANGELOG.md
```diff
@@ -14,9 +14,14 @@ the code was deployed.
 ### Added
 
 - CORS configuration to Express Proxy Middleware.
-- "doorMissedFrequently" field to JSON data submitted by brave sensors that is sent to /api/heartbeat (CU-860rk8v2a)
-- Sentry log in the case that doorMissedFrequently is true in posted data from brave sensor (CU-860rk8v2a)
+- Added a timer in state0 that is reset when the door closes where the sensor is allowed to detect for occupation (CU-8678t4ztg).
+- Added console function for the occupation detection timer as well as the timer count and timeout value in Debug Publishes (CU-8678t4ztg).
+- "doorMissedFrequently" field to JSON data submitted by brave sensors that is sent to /api/heartbeat (CU-860rk8v2a).
+- Sentry log in the case that doorMissedFrequently is true in posted data from brave sensor (CU-860rk8v2a).
```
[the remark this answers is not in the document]
johnny-brave: I added periods to the end of these lines

firmware/boron-ins-fsm/src/imDoorSensor.cpp
johnny-brave: 0x99 is 0b10011001 which is "closed + heartbeat". In the new logic I am implementing, a transition from "closed heartbeat" to "closed heartbeat" would not reset the timer. However, the timer does still reset when the sensor resets since `millis()` resets to 0 as well. so if the realization happens when the occupation timer has not been exceeded it can enter state 1. I believe we don't want this.

CHANGELOG.md
```diff
@@ -14,9 +14,14 @@ the code was deployed.
...
 - GitHub Actions to deploy to AWS infrastructure on Dev and Staging (CU-860ra8f7q).
+### Fixed
+
+- Wording in consoleFunctionTests.cpp when functions are called with 'e'.
```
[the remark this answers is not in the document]
johnny-brave: I usually pull from main, fix merge conflicts and then recommit. Would you recommend that I rebase moving forward? From what I've read, rebasing recommits every one of my changes as if I had done it after everyone else's changes. The "merge" approach is what I used for this feature, but I will try rebasing for future commits if you think it's better.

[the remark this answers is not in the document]
trisapeace: I find that sometimes merging into my branch creates a weird `diff` situation in GitHub where it shows code that I merged in as though I wrote it new in my branch. And that gets confusing. If you can avoid this, then merging is fine. @micahdbak also used merging in his most recent PR. I use rebase because it gives me a lot of control, but it does carry risk in that it's changing the past and needs a force push every time I use it.

firmware/boron-ins-fsm/src/imDoorSensor.cpp
trisapeace: Just to have it written down... I think we chatted about this on Friday. I think that when we talked it through we decided that resetting to `millis()` is fine here. You're right that it does mean that we'll maybe get into state 1 when we didn't want to, but that's a better default than not ending up in state 1 when we should in the case of a restart during the daytime hours.

firmware/boron-ins-fsm/src/imDoorSensor.cpp
```diff
@@ -69,6 +70,13 @@ doorData checkIM() {
         // Checks if the 0th bit (counting from 0) of doorStatus is 1
         doorTamperedFlag = (currentDoorData.doorStatus & 0b0001) != 0;
 
+        // Reset timer on receiving a door close message or transition from open to closed + heartbeat
+        if ((currentDoorData.doorStatus & 0b0010) == 0) {
+            if ((currentDoorData.doorStatus & 0b1000) == 0 || (previousDoorData.doorStatus & 0b0010) != 0) {
```
trisapeace, line 75: When I try to interpret this code, I get:
```
if (door is currently closed) {
   if (door is sent a heartbeat || door was previously open) {
      reset timer
   }
}
```

In other words:
```
If the door is currently closed and it's a heartbeat
OR
if the door is currently closed but was previously open
THEN
reset timer
```

Which isn't the same thing to me as the comment "Reset timer on receiving a door close message or transition from open to closed + heartbeat". This sounds to me like:
```
If the door is currently closed but was previously open
OR
If the door is currently a closed heartbeat but was previous open
THEN
reset timer
```

Am I misunderstanding something? Which is the version that you really wanted to happen?

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -319,8 +337,9 @@ void publishDebugMessage(int state, unsigned char doorStatus, float INSValue, un
             char debugMessage[622];
             snprintf(debugMessage, sizeof(debugMessage),
                      "{\"state\":\"%d\", \"door_status\":\"0x%02X\", \"INS_val\":\"%f\", \"INS_threshold\":\"%lu\", \"timer_status\":\"%lu\", "
-                     "\"initial_timer\":\"%lu\", \"duration_timer\":\"%lu\", \"stillness_timer\":\"%lu\"}",
-                     state, doorStatus, INSValue, ins_threshold, timer, state1_max_time, state2_max_duration, *max_stillness_time);
+                     "\"occupation_detection_timer\":\"%lu\", \"initial_timer\":\"%lu\", \"duration_timer\":\"%lu\", \"stillness_timer\":\"%lu\"}",
```
trisapeace, line 340: The next time someone updates this, could you please add the long stillness timer too? Thanks!

trisapeace (commented):
Fantastic! This was a big change for the first time making changes to the firmware. I think you're well prepared for any future changes that come your way!

firmware/boron-ins-fsm/src/imDoorSensor.cpp
trisapeace: Oops, sorry, I **was** reading it wrong. But I still don't think it matches up with the comment. Now I see that the code is:
```
If the door is currently closed and it's a not a heartbeat
OR
if the door is currently closed but was previously open
THEN
reset timer
```

trisapeace:
For future reference: Slides about this change are here: https://docs.google.com/presentation/d/1UmEY2Prv1ND-JZTCB8GoVLYiJSiwYVLpoFJG4IHXhYY/edit#slide=id.p

PR #234, opened 2023-09-20
Investigate the API authorize function (CU-8678uuvjm)

micahdbak:
I noticed inconsistency in the way that the API authorize function was defined, so this task was to implement unit tests to ensure that the API authorize function works as intended.

Test plan:
- [x] Implement unit tests for primary/secondary PA API key in the query/body of a request
- [x] Ensure that the fixed authorize function does not cause other tests to not pass


trisapeace:
Task linked: [CU-8678uuvjm Investigate functionality of the API authorize function and its primary and secondary keys](https://app.clickup.com/t/8678uuvjm)

PR #239, opened 2023-09-27
Production workflow

phenom5886:
- Added deploy workflow for new AWS production environment
- Added a check to see if the GH Ref is production branch

Note: Haven't been able to test the ref check, need this in main. If it doesn't work as expected then I can modify and test in a branch once this exists in main and open a new PR.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -1,0 +1,61 @@
+name: Deploy AWS Production
+
+run-name: Deploy ${{ github.ref }} to Production
+
+on:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-production.yml
```diff
@@ -0,0 +1,59 @@
+name: Deploy AWS Production
+
```
trisapeace, line 2: I've been adding this to include the branch name in the name of the run:

`run-name: Deploy ${{ github.ref }} to Production`

trisapeace (approved):

PR #242, opened 2023-10-04
Integrates internal alert system using Sentry, activating alerts when a sensor's INS value is equal to or less than zero (CU-8678v3y54).

ahmaddaadaa:
Implements internal alert functionality with Sentry alerts triggered when a sensor's INS value is less than or equal to zero [CU8678v3y54](https://app.clickup.com/t/8678v3y54)

Test Plane:
- [x] Ensure successful Travis test execution.
- [x] Develop and validate integration test to assess the functionality of the Sentry messaging system when the 'isINSZero' flag is triggered.  > `test/integration/vitalsTest/handleHeartbeatTest.js`
- [x] Deploy the new firmware on a test sensor.
- [x] Thoroughly test the firmware on the test sensor to validate functionality. >  `Observed the Heartbeat data, including the 'isINSZero' flag key triggering`
- [x] The Sentry message is displayed for a registered client (Micha's sensor)

### DIFF
--- a/firmware/boron-ins-fsm/clang-format-all.py
+++ b/firmware/boron-ins-fsm/clang-format-all.py
@@ -9,5 +9,5 @@
 
 import os
 
-os.system("find src -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format-12 -i")
-os.system("find test -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format-12 -i")
+os.system("find src -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format -i")
+os.system("find test -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format -i")
--- a/firmware/boron-ins-fsm/clang-format-all.py
+++ b/firmware/boron-ins-fsm/clang-format-all.py
@@ -9,5 +9,5 @@
 
 import os
 
-os.system("find src -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format -i")
-os.system("find test -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format -i")
+os.system("find src -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format-12 -i")
+os.system("find test -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format-12 -i")
--- a/firmware/boron-ins-fsm/src/stateMachine.cpp
+++ b/firmware/boron-ins-fsm/src/stateMachine.cpp
@@ -381,7 +381,8 @@ void getHeartbeat() {
     static unsigned long lastHeartbeatPublish = 0;
     
     // Check INS value
-    bool isINSZero = (checkINS3331() <= 0.0);
+    filteredINSData checkINS = checkINS3331();
+    bool isINSZero = (checkINS.iAverage < 0.0001);
 
     // 1st "if condition" is so that the boron publishes a heartbeat on startup
     // 2nd "if condition" is so that the boron publishes a heartbeat, when the doorMessageReceivedFlag is true.
--- a/firmware/boron-ins-fsm/src/stateMachine.cpp
+++ b/firmware/boron-ins-fsm/src/stateMachine.cpp
@@ -379,10 +379,6 @@ const char *resetReasonString(int resetReason) {
 
 void getHeartbeat() {
     static unsigned long lastHeartbeatPublish = 0;
-    
-    // Check INS value
-    filteredINSData checkINS = checkINS3331();
-    bool isINSZero = (checkINS.iAverage < 0.0001);
 
     // 1st "if condition" is so that the boron publishes a heartbeat on startup
     // 2nd "if condition" is so that the boron publishes a heartbeat, when the doorMessageReceivedFlag is true.
@@ -391,14 +387,18 @@ void getHeartbeat() {
     // 3rd "if condition" is true only if a heartbeat hasnt been published in the last SM_HEARTBEAT_INTERVAL
     if (lastHeartbeatPublish == 0 || (doorMessageReceivedFlag && ((millis() - doorHeartbeatReceived) >= HEARTBEAT_PUBLISH_DELAY)) ||
         (millis() - lastHeartbeatPublish) > SM_HEARTBEAT_INTERVAL) {
+        // Check INS value
+        filteredINSData checkINS = checkINS3331();
+        bool isINSZero = (checkINS.iAverage < 0.0001);
         static unsigned int didMissQueueSum = 0;
         static std::queue<bool> didMissQueue;
         // from particle docs, max length of publish is 622 chars, I am assuming this includes null char
         char heartbeatMessage[622] = {0};
         JSONBufferWriter writer(heartbeatMessage, sizeof(heartbeatMessage) - 1);
         writer.beginObject();
-        // Add "is_ins_zero" field to the JSON message
-        writer.name("is_ins_zero").value(isINSZero);
+      
+        // Add "isINSZero" field to the JSON message
+        writer.name("isINSZero").value(isINSZero);
 
         if (didMissQueue.size() > SM_HEARTBEAT_DID_MISS_QUEUE_SIZE) {
             // if oldest value did miss; subtract from the current amount
--- a/firmware/boron-ins-fsm/src/stateMachine.cpp
+++ b/firmware/boron-ins-fsm/src/stateMachine.cpp
@@ -396,7 +396,7 @@ void getHeartbeat() {
         char heartbeatMessage[622] = {0};
         JSONBufferWriter writer(heartbeatMessage, sizeof(heartbeatMessage) - 1);
         writer.beginObject();
-      
+
         // Add "isINSZero" field to the JSON message
         writer.name("isINSZero").value(isINSZero);
 
--- a/firmware/boron-ins-fsm/src/stateMachine.cpp
+++ b/firmware/boron-ins-fsm/src/stateMachine.cpp
@@ -379,6 +379,9 @@ const char *resetReasonString(int resetReason) {
 
 void getHeartbeat() {
     static unsigned long lastHeartbeatPublish = 0;
+    
+    // Check INS value
+    bool isINSZero = (checkINS3331() <= 0.0);
 
     // 1st "if condition" is so that the boron publishes a heartbeat on startup
     // 2nd "if condition" is so that the boron publishes a heartbeat, when the doorMessageReceivedFlag is true.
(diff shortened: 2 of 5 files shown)


server/vitals.js
```diff
@@ -280,6 +281,10 @@ async function handleHeartbeat(req, res) {
             await sendIsTamperedAlert(location, isTamperedFlag, mostRecentSensorVitals.isTampered)
           }
 
+          if (isINSZero) {
+            helpers.logSentry(`INS sensor is les or equal zero at ${location.locationid}`)
+          }
+
```
tmackinnon, line 287: Looks good! Just a small spelling mistake with the word "less" - I also wonder if it makes more sense to use "INS value" instead of "INS sensor".

firmware/boron-ins-fsm/src/stateMachine.cpp
```diff
@@ -387,13 +387,19 @@ void getHeartbeat() {
     // 3rd "if condition" is true only if a heartbeat hasnt been published in the last SM_HEARTBEAT_INTERVAL
     if (lastHeartbeatPublish == 0 || (doorMessageReceivedFlag && ((millis() - doorHeartbeatReceived) >= HEARTBEAT_PUBLISH_DELAY)) ||
         (millis() - lastHeartbeatPublish) > SM_HEARTBEAT_INTERVAL) {
+        // Check INS value
+        filteredINSData checkINS = checkINS3331();
+        bool isINSZero = (checkINS.iAverage < 0.0001);
```
tmackinnon, line 392: Awesome work!

tmackinnon (commented):
Great work Ahmad! I have a few notes. 
- Could you update your PR title to include the Click Up task and also make the title a bit more descriptive of the change as opposed to your branch name? 
- In your test plan, you say you "thoroughly tested the firmware" - could you provide some more context on what was done?
- Could you also add your "isINSZero" value to the tests in the `server/test/integration/vitalsTest/handleHeartbeatTest.js` file? And I think it would be valuable to create a test case in there as well for your functionality. (I would refer to [Micah's recent PR](https://github.com/bravetechnologycoop/BraveSensor/pull/230/files#diff-1c070ecbd98a883eef187b46bfd4a74df51761872e7854421f355b061ae7d0b0) for inspiration)

Otherwise its looking good!

firmware/boron-ins-fsm/clang-format-all.py
```diff
@@ -9,5 +9,5 @@
 
 import os
 
-os.system("find src -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format -i")
-os.system("find test -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format -i")
\ No newline at end of file
+os.system("find src -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format-12 -i")
+os.system("find test -iname *.h -o -iname *.cpp -o -iname *.ino | xargs clang-format-12 -i")
```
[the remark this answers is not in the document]
ahmaddaadaa: I specified `-12` in `clang-format-12` to indicate the specific version I used. If I were to remove it and use just `clang-format`, the system would default to the installed version, which might vary across systems. My intention was to ensure consistent code formatting across all future pull requests. Different versions can have distinct formatting rules and features, so specifying a version helps maintain a uniform style. I tested version 12 with Travis, and it proved to be a reliable choice for our project.

server/vitals.js
ahmaddaadaa: Thanks for spotting the typo!

firmware/boron-ins-fsm/clang-format-all.py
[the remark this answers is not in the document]
tmackinnon: If you remember from our training with Theresa, she said that there were inconsistencies with the two clang files, so some systems needed to use it with the `-12` and some needed to use it without. Because of this, I don't think we should be changing the file. Also since Theresa was aware of this and continued to leave the files as is in her subsequent PR's, I think we should do the same.

ahmaddaadaa:
> Great work Ahmad! I have a few notes.
> 
> * Could you update your PR title to include the Click Up task and also make the title a bit more descriptive of the change as opposed to your branch name?
> * In your test plan, you say you "thoroughly tested the firmware" - could you provide some more context on what was done?
> * Could you also add your "isINSZero" value to the tests in the `server/test/integration/vitalsTest/handleHeartbeatTest.js` file? And I think it would be valuable to create a test case in there as well for your functionality. (I would refer to [Micah's recent PR](https://github.com/bravetechnologycoop/BraveSensor/pull/230/files#diff-1c070ecbd98a883eef187b46bfd4a74df51761872e7854421f355b061ae7d0b0) for inspiration)
> 
> Otherwise its looking good!

Thank you for your feedback. I have updated the title as per your request. Regarding your inquiry about the sensor testing, in this scenario, I meticulously monitored the sensor's data output, updated every second on the PA dashboard. I closely observed the Heartbeat data, including the 'isINSZero' flag key triggering. The data flow remained consistent throughout the test.

Additionally, I have added vital test within the 'handleHeartbeat' file to verify the expected behavior of logging a Sentry alert when the INS sensor falls below the threshold. The test successfully passed, confirming the robust functionality of the code.

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -215,6 +215,17 @@ describe('vitals.js integration tests: handleHeartbeat', () => {
...
+      const response = await chai.request(server).post('/api/heartbeat').send({
+        coreid: radar_coreID,
+        data: `{"isINSZero":true,"doorMissedMsg":0,"doorMissedFrequently":false,"doorLastMessage":-1,"doorLowBatt":-1,"doorTampered":-1,"resetReason":"NONE","states":[]}`,
+        api_key: webhookAPIKey,
+      })
+
+      expect(response).to.have.status(200)
+      expect(helpers.logSentry).to.have.been.calledWith(`INS sensor is less or equal zero at ${testLocation1Id}`)
+    })
+
```
tmackinnon, line 228: This is looking good but I'm wondering if instead of adding your `it` statement to an existing test, you should instead create a new test (starting with the describe statement) at the bottom of the file, so that you are just testing your own functionality. See what Micah had done is his PR that I referenced.

ahmaddaadaa: I am going to keep it the same since my test statement relates to the same description field as in the previous code, specifically when a heartbeat message is sent. Creating a separate description field is unnecessary and would complicate the overall code, making it harder to track and read.

trisapeace:
Task linked: [CU-8678v3y54 Internal alert when a Sensors INS value <= 0](https://app.clickup.com/t/8678v3y54)

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -215,6 +215,17 @@ describe('vitals.js integration tests: handleHeartbeat', () => {
       sandbox.restore()
     })
 
+    it('should log a Sentry alert when the INS sensor value is less or equal to zero', async () => {
```
micahdbak, line 218: Quick note: because we need to deploy this PR very soon, maybe you can add a follow-up PR for the test cases in this file: you should add `isINSZero: false` to every other `data` request in this file, to make sure that that field does not change the logic of the other test cases.

micahdbak (approved):
Everything looks good to go for this morning!

Johnny did request that sequential alerts should not be sent for when a sensor's INS sensor is giving only 0.0, but I think that change will require more time for you to implement.

I think it would be a good idea to go through with this PR to get the change implemented in production ASAP, but then you can work on a follow up PR with a few changes (see my comment about test-cases, and Johnny's request).

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -215,6 +215,17 @@ describe('vitals.js integration tests: handleHeartbeat', () => {
...
+      const response = await chai.request(server).post('/api/heartbeat').send({
+        coreid: radar_coreID,
+        data: `{"isINSZero":true,"doorMissedMsg":0,"doorMissedFrequently":false,"doorLastMessage":-1,"doorLowBatt":-1,"doorTampered":-1,"resetReason":"NONE","states":[]}`,
+        api_key: webhookAPIKey,
+      })
+
+      expect(response).to.have.status(200)
+      expect(helpers.logSentry).to.have.been.calledWith(`INS sensor is equal to or less than zero at ${testLocation1Id}`)
```
trisapeace, line 226: In general, automated tests should have only a single assertion/expects-statement in them. This is so that earlier failures do not hide the results of later tests. 

In this case, if the response had a status code of 201 instead of 200, the test would fail immediately and you would have no way of knowing whether or Sentry was called with the correct message or not.

If you split this up into two tests, one that checks for the return value and the other that checks for the Sentry message, then you will have a better overall picture of the program when one, the other, or both fail.

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -215,6 +215,17 @@ describe('vitals.js integration tests: handleHeartbeat', () => {
       sandbox.restore()
     })
 
+    it('should log a Sentry alert when the INS sensor value is less or equal to zero', async () => {
+      const response = await chai.request(server).post('/api/heartbeat').send({
+        coreid: radar_coreID,
+        data: `{"isINSZero":true,"doorMissedMsg":0,"doorMissedFrequently":false,"doorLastMessage":-1,"doorLowBatt":-1,"doorTampered":-1,"resetReason":"NONE","states":[]}`,
```
trisapeace, line 221: Thanks @tmackinnon for adding `"isINSZero":false` to the other tests in https://github.com/bravetechnologycoop/BraveSensor/pull/247/files !

server/vitals.js
```diff
@@ -280,6 +281,10 @@ async function handleHeartbeat(req, res) {
             await sendIsTamperedAlert(location, isTamperedFlag, mostRecentSensorVitals.isTampered)
           }
 
+          if (isINSZero) {
```
trisapeace, line 284: @ahmaddaadaa , it's unclear to me how or whether you tested the case when the INS value was 0. I don't see that test case in your test plan. Were you able to do it?

firmware/boron-ins-fsm/clang-format-all.py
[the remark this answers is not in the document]
trisapeace: I'd actually be OK with it changing in the files to make them consistently what Travis expects. Then some people will always need to change it and others never, rather than having it like it is now where everyone needs to change it in one file but not the other, lol.

trisapeace (commented):
Cool. This looks great. Glad you were able to get it up and working for the clients. I also saw somewhere in Slack that it's already found some real life examples of INS <= 0 in production, so it's already proving its worth. Very cool.

As for Johnny's request to not send repeated Sentry messages, is there a ClickUp Task for that follow-up work? If not and if it's still needed, can you please create the Task?

PR #243, opened 2023-10-10
Fixed response object of /api/message-clients to match conventions

micahdbak:
Upon porting this feature to the BraveButtons repository, I discovered that the response object returned by /api/message-clients does not match the conventions described at the top of the api.js file in this repository.

This PR does not change the functionality of this API request; it simply moves some of the response object under a `data` object. Please read the conventions information at the top of api.js for more information.

Test Plan:
- [x] Passes all previously created tests for /api/message-clients.
- [x] Returned response object matches the conventions outlined in api.js.

### DIFF
--- a/server/api.js
+++ b/server/api.js
@@ -9,7 +9,7 @@
  *  - Must return a JSON object containing the following keys:
  *    - status:   which will be either "success" or "error"
  *    - data:     the desired JSON object, if there is one
- *    - message:  a human-readable explaination of the error, if there was one and this is appropriate. Be careful
+ *    - message:  a human-readable explanation of the error, if there was one and this is appropriate. Be careful
  *                to not include anything that will give an attacker extra information
  */
 
@@ -88,9 +88,12 @@ async function messageClients(req, res) {
       const message = req.body.message
       const response = {
         status: 'success',
-        message,
-        contacted: [],
-        failed: [],
+        data: {
+          message,
+          contacted: [],
+          failed: [],
+        },
+        message: '',
       }
 
       for (const client of clients) {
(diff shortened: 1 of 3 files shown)


server/api.js
```diff
@@ -88,9 +88,12 @@ async function messageClients(req, res) {
...
-        message,
-        contacted: [],
-        failed: [],
+        data: {
+          message,
+          contacted: [],
+          failed: [],
+        },
+        message: '',
       }
```
ahmaddaadaa, line 97: make sure the format of the data provided here matches the structure expected by the lambda function handling your POST request. you might need to adjust the lambda function to handle the data in the same format as provide here. you have added new fields and there should be accounted for in the Lambda function as well.

tmackinnon (approved):
Good catch Micah! Looks good to me.

ahmaddaadaa (commented):
double check the Lambda function

server/api.js
micahdbak: Hi Ahmad, which Lambda function are you referring to? What I have changed is the information that is sent in response to a successful POST request. The information POSTed to this API call has not changed.

At present, the response information is not used anywhere; it will be handled in PA, which I will work on in a future PR.

server/api.js
```diff
@@ -9,7 +9,7 @@
  *  - Must return a JSON object containing the following keys:
  *    - status:   which will be either "success" or "error"
  *    - data:     the desired JSON object, if there is one
- *    - message:  a human-readable explaination of the error, if there was one and this is appropriate. Be careful
+ *    - message:  a human-readable explanation of the error, if there was one and this is appropriate. Be careful
```
trisapeace, line 12: Thanks for fixing the typo!

server/api.js
```diff
@@ -88,9 +88,12 @@ async function messageClients(req, res) {
...
         status: 'success',
-        message,
-        contacted: [],
-        failed: [],
+        data: {
+          message,
+          contacted: [],
+          failed: [],
+        },
+        message: '',
```
trisapeace, line 96: Cool. Thanks for making it conform to the description!

PR #245, opened 2023-10-18
Write 1 byte when initializing Door ID (CU-8678xpmpu)

johnny-brave:
During initialization of the Door Sensor, the value written into EEPROM was not specified as a 1 byte data field. This caused the address fields for the long stillness timer to be overwritten with 0's. This bugfix writes uin8_t data to the door sensor upon initialization to prevent  unintended overwriting of other flash addresses.

Test plan:
---

- [x] Flash firmware onto a new boron and confirm that all initial values are correct including the long stillness timer
- [x] Write new value in the long-stillness timer using the particle console function and echo to confirm that it was saved
- [x] Reset boron and confirm that the new value persists through reset
- [x] Flash firmware onto an existing boron and confirm that previous memory values are written correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ the code was deployed.
 ### Fixed
 
 - Specify 1 byte write into Door ID flash addresses, preventing erroneous overwrite of long stillness timer flash addresses (CU-8678xpmpu). 
+- Test data object in handleHeartbeatTest to include "isINSZero" key-value (CU-8678y7w9y).
+- Clang formatting error by removing extra space.
 
 ## [10.0.0] - 2023-10-12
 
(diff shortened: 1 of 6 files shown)


micahdbak (commented):
Hi Johnny, please merge in the latest changes from main. This will resolve the Travis checks that are currently failing

micahdbak (approved):
Looks good to merge!

trisapeace (commented):
Good catch @johnny-brave ! Nice fix. Glad you figured it out.

Did you also check to make sure that writing a new door sensor doesn't overwrite the long stillness timer?

trisapeace:
P.S. Also, for a PR like this where there are a lot of commits with changes and messages like "formatting", it would lead to a cleaner commit history to use "Squash and Merge" instead of "Merge". This will create a single commit in the `main` branch and you can write a single commit message explaining one time what it was that this PR does.

PR #248, opened 2023-10-25
Deadlock bug fixes (CU-8678wudz8)

tmackinnon:
Made some small changes to see if they will have any effect on the increased deadlock errors that have been coming in. I will monitor the Sentry error logs after these changes are deployed, then revisit to amend/update. Changes are outlined below and in the clickup task comments. I don't _think_ I need to add new tests for these changes, but let me know if you think otherwise.

**vitals.js**
- Added error handling to `sendLowBatteryAlert()` to stop subsequent db queries  

**BraveAlerterConfigurator.js**
- Added retry to `alertSessionChangedCallback()` per convo with Theresa 
- Retrying `beginTransaction` as we only have on shot at this function working (vs heartbeat that comes in often and can capture data the next time its called) 
- Retry has no time delay as of right now, but could be something to look into later

**db.js**
- Removed lock on tables that are not used in the functions that call beginTransaction
- Another solution could be to have `beginTransaction` take in a parameter that represents the exact tables needed to be locked - currently I don't think is necessary unless `beginTransaction` starts to gets used in more functions
- Added explicit return values to `beginTransaction` and `getMostRecentSensorsVitalWithLocationid`in their catch blocks
- Update parameter in `getMostRecentSensorsVitalWithLocationid()` so that it takes the whole `location` object INSTEAD of just `locationid` (since the location info is already available when this function gets called) . This would eliminate the need to call getLocations reducing the use of this db resource, and ideally reducing chance of deadlock.
   - Update the function name to `getMostRecentSensorsVitalWithLocation` to reflect this change

**checkHeartbeatTest.js**
- Update function name from `getMostRecentSensorsVitalWithLocationid` to `getMostRecentSensorsVitalWithLocation`

Test Plan:
- [x] Test case in `checkHeartbeatTest.js` passes: Error is logged from the `SendLowBatteryAlert` function when `db.beginTransaction` returns null 
- [x] Ensure Travis is passing
- [x] Deploy to Dev with no issues
- [x] Smoke test passes

### DIFF
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -90,12 +90,8 @@ class BraveAlerterConfigurator {
     try {
       pgClient = await db.beginTransaction()
       if (pgClient === null) {
-        // If the beginTransaction function failed to start the transaction, retry it once more before logging an error
-        pgClient = await db.beginTransaction()
-        if (pgClient === null) {
-          helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
-          return
-        }
+        helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
+        return
       }
       session = await db.getSessionWithSessionId(alertSession.sessionId, pgClient)
 
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -90,8 +90,12 @@ class BraveAlerterConfigurator {
     try {
       pgClient = await db.beginTransaction()
       if (pgClient === null) {
-        helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
-        return
+        // If the beginTransaction function failed to start the transaction, retry it once more before logging an error
+        pgClient = await db.beginTransaction()
+        if (pgClient === null) {
+          helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
+          return
+        }
       }
       session = await db.getSessionWithSessionId(alertSession.sessionId, pgClient)
 
--- a/server/db/db.js
+++ b/server/db/db.js
@@ -50,13 +50,12 @@ function createSensorsVitalFromRow(r, allLocations) {
   return new SensorsVital(r.id, r.missed_door_messages, r.is_door_battery_low, r.door_last_seen_at, r.reset_reason, r.state_transitions, r.created_at, r.is_tampered, location)
 }
 
-async function beginTransaction() {
+async function beginTransaction(retryCount = 0) {
   if (helpers.isDbLogging()) {
     helpers.log('STARTED: beginTransaction')
   }
 
   let pgClient = null
-  let isRetry = false
 
   try {
     pgClient = await pool.connect()
@@ -68,10 +67,9 @@ async function beginTransaction() {
 
     await pgClient.query('LOCK TABLE clients, sessions, locations')
   } catch (e) {
-    if (!isRetry) {
-      isRetry = true
+    if (retryCount < 1) {
       helpers.logError(`Error running the beginTransaction query: ${e}. Will retry beginTransaction again.`)
-      beginTransaction()
+      beginTransaction(retryCount + 1)
     }
     helpers.logError(`Error running the beginTransaction query: ${e}`)
     if (pgClient) {
--- a/server/db/db.js
+++ b/server/db/db.js
@@ -71,6 +71,7 @@ async function beginTransaction() {
     if (pgClient) {
       try {
         await this.rollbackTransaction(pgClient)
+        return null
       } catch (err) {
         helpers.logError(`beginTransaction: Error rolling back the errored transaction: ${err}`)
       }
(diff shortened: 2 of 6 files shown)


server/vitals.js
```diff
@@ -162,6 +162,10 @@ async function sendLowBatteryAlert(locationid) {
     const timeoutInMillis = parseInt(helpers.getEnvVar('LOW_BATTERY_ALERT_TIMEOUT'), 10) * 1000
 
     pgClient = await db.beginTransaction()
+    if (pgClient === null) {
+      helpers.logError(`sendLowBatteryAlert: Error starting transaction`)
```
trisapeace, line 166: You could add a test to make sure that this error log happens when `pgClient` is `null`.

trisapeace (approved):
Fantastic.

I have hope that this will make a real difference in the deadlocks!

I made some comments and suggestions, but nothing serious. You're choice if you make any of the changes.

server/BraveAlerterConfigurator.js
```diff
@@ -90,8 +90,12 @@ class BraveAlerterConfigurator {
     try {
       pgClient = await db.beginTransaction()
       if (pgClient === null) {
-        helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
-        return
+        // If the beginTransaction function failed to start the transaction, retry it once more before logging an error
+        pgClient = await db.beginTransaction()
+        if (pgClient === null) {
+          helpers.logError(`alertSessionChangedCallback: Error starting transaction`)
+          return
```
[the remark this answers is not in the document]
tmackinnon: Oh that's a good point! Are you thinking in the catch block of `beginTransaction`?

server/db/db.js
```diff
@@ -65,12 +65,13 @@ async function beginTransaction() {
 
     await pgClient.query('BEGIN')
 
-    await pgClient.query('LOCK TABLE clients, notifications, sessions, locations, sensors_vitals, sensors_vitals_cache')
```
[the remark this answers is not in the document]
tmackinnon: Ah yes that makes a lot of sense. And yes we definitely can revisit the idea of passing table names in as parameters in the future, if we find it necessary.

server/BraveAlerterConfigurator.js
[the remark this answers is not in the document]
trisapeace: Yes, that was what I was thinking.

server/db/db.js
```diff
@@ -1176,10 +1177,11 @@ async function getMostRecentSensorsVitalWithLocationid(locationid, pgClient) {
       return null
     }
 
-    const allLocations = await getLocations(pgClient)
+    const allLocations = [location]
     return createSensorsVitalFromRow(results.rows[0], allLocations)
```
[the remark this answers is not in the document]
tmackinnon: You're so right, will be much clearer that way!

server/db/db.js
```diff
@@ -67,6 +68,11 @@ async function beginTransaction() {
 
     await pgClient.query('LOCK TABLE clients, sessions, locations')
   } catch (e) {
+    if (!isRetry) {
+      isRetry = true
```
trisapeace, line 72: Hmmm, does this work? I would have expected `isRetry` to be `false` again each time you call `beginTransaction`.

tmackinnon: Hmm good point, I think instead I'll add a default parameter for the retryCount then use number for my conditional. Stay tuned!

server/vitals.js
tmackinnon: Good idea, added!

server/db/db.js
```diff
@@ -1176,10 +1181,10 @@ async function getMostRecentSensorsVitalWithLocationid(locationid, pgClient) {
       return null
     }
 
-    const allLocations = await getLocations(pgClient)
-    return createSensorsVitalFromRow(results.rows[0], allLocations)
+    return createSensorsVitalFromRow(results.rows[0], [location])
```
micahdbak, line 1184: Very nice change here. Removes an unnecessary (and probably slow) query to get all locations; great catch.

server/test/integration/vitalsTest/handleHeartbeatTest.js
```diff
@@ -9,6 +9,7 @@ const sinon = require('sinon')
 const { factories, helpers } = require('brave-alert-lib')
 const { braveAlerter, db, server } = require('../../../index')
 const { locationDBFactory, sensorsVitalDBFactory } = require('../../../testingHelpers')
+// const { sendLowBatteryAlert } = require('../../../vitals')
```
micahdbak, line 12: Is there a reason that this line is only commented out, as opposed to removed?

micahdbak (approved):
Looks good! Looks like what you've changed is pretty small, but I hope that makes an impact on the deadlocks. I look forward to seeing what results from this.

server/test/integration/vitalsTest/handleHeartbeatTest.js
tmackinnon: Oh whoops thank you for catching that! That line shouldn't be in there. I thought I would maybe need it when I was writing the test but I didn't and I guess forgot to delete it. Thanks!!

server/db/db.js
tmackinnon: Thanks!

PR #250, opened 2023-11-06
beginTransaction bug fix (CU-8679376g6)

tmackinnon:
Fixed `beginTransaction` function where it retries calling the function once more, after erroring out the first time it's called. The `return` keyword was missing from recursion call, causing unexpected problems in server. `return await` has been added. Also the rollback functionality has been moved, so that if `pgClient` is truthy when there's an error, it gets rolledback (whether it is attempt 0 or 1). 

Per Theresa's suggestion (see below) I've split the original function into two, `beginTransaction` and `runBeginTransactionWithRetries`. `beginTransaction` calls `runBeginTransactionWithRetries(0)` and all the previous logic has been moved into `runBeginTransactionWithRetries`. This is so that `beginTransaction` cannot get called with different parameters values, and `runBeginTransactionWithRetries` will not be accessed outside of `db.js`.

Also added integration testing for this functionality to ensure this doesn't affect production server again. 

## Test Plan
Test in `beginTransactionTest.js`
- [x] Happy Path - beginTransaction is successful on first attempt
- [x] beginTransaction fails first attempt because of a deadlock but is successful upon retry
- [x] beginTransaction fails first attempt and second attempt

Other Testing
- [x] Deploy to dev - make sure nothing breaks
- [x] Smoketest
- [x] Travis Passes


trisapeace:
Task linked: [CU-8679376g6 Fix beginTransaction retry error](https://app.clickup.com/t/8679376g6)

server/test/unit/dbTest/beginTransactionTest.js
```diff
@@ -0,0 +1,105 @@
...
+const { expect, use } = require('chai')
+const { describe, it } = require('mocha')
+const sinon = require('sinon')
+const sinonChai = require('sinon-chai')
+const rewire = require('rewire')
+
+// In-house dependencies
+const { helpers } = require('brave-alert-lib')
+
+const db = rewire('../../../db/db')
```
trisapeace, line 11: Neat. Looks like you're becoming an expert in `rewire`!

server/test/unit/dbTest/beginTransactionTest.js
```diff
@@ -0,0 +1,105 @@
...
+  afterEach(() => {
+    sandbox.restore()
+  })
+
+  describe('when beginTransaction fails the first time because of a deadlock, then successfully connects after retry', () => {
+    let clientStub
+    beforeEach(async () => {
+      clientStub = { query: sandbox.stub() }
+      sandbox.spy(db, 'beginTransaction')
+      poolConnectStub.onCall(0).rejects(new Error('deadlock detected')).onCall(1).resolves(clientStub)
```
trisapeace, line 39: Nice!

server/test/unit/dbTest/beginTransactionTest.js
```diff
@@ -0,0 +1,105 @@
...
+      expect(helpers.logError).to.be.calledWith(`Error running the beginTransaction query: Error: deadlock detected`)
+    })
+
+    it('should log the retry', () => {
+      expect(helpers.log).to.be.calledWith(`Retrying beginTransaction.`)
+    })
+
+    it('should not return null', () => {
+      expect(pgClient).to.not.be.null
+    })
```
trisapeace, line 53: This is not only not `null`, but it should be `clientStub`, right?

server/test/unit/dbTest/beginTransactionTest.js
```diff
@@ -0,0 +1,105 @@
...
+
+    it('should log the deadlock error', () => {
+      expect(helpers.logError).to.be.calledWith(`Error running the beginTransaction query: Error: deadlock detected`)
+    })
+
+    it('should log the retry', () => {
+      expect(helpers.log).to.be.calledWith(`Retrying beginTransaction.`)
+    })
+
```
trisapeace, line 50: I suggest adding a test here like the one that you have below:
```    
it('should lock the clients, sessions, and locations tables', () => {
      expect(clientStub.query).to.be.calledWith(`LOCK TABLE clients, sessions, locations`)
})
```

server/test/unit/dbTest/beginTransactionTest.js
```diff
@@ -0,0 +1,105 @@
...
+      clientStub = { query: sinon.stub() }
+      sandbox.spy(db, 'beginTransaction')
+      poolConnectStub.onCall(0).resolves(clientStub)
+      pgClient = await db.beginTransaction()
+    })
+
+    it('should not log any errors', () => {
+      expect(helpers.logError).to.not.be.called
+    })
+
```
trisapeace, line 68: I suggest adding a check that the retry is not logged.

server/db/db.js
```diff
@@ -67,19 +67,21 @@ async function beginTransaction(retryCount = 0) {
...
         await this.rollbackTransaction(pgClient)
-        return null
       } catch (err) {
         helpers.logError(`beginTransaction: Error rolling back the errored transaction: ${err}`)
       }
     }
+
+    if (retryCount < 1) {
+      helpers.log(`Retrying beginTransaction.`)
+      return await beginTransaction(retryCount + 1)
```
trisapeace, line 82: By having the `retry` a parameter into this function, you're opening this function up to questions such as "What happens if you call `beginTransaction(3)`?"

From the outside of `db.js`, we only ever want to call `beginTransaction` with no parameters, right?

What do you think about splitting this up into two functions? Like this:
1. `beginTransaction` (which takes no parameters and is exported and accessible from outside db.js)
2. `runBeginTransactionWithRetries` or whatever you think would be a good name for it (which takes an integer `retry` parameter and is not accessible from outside db.js)

Then `beginTransaction` will be just `return await runBeginTransactionWithRetries(0)`

and you can rename what you've written here to be `runBeginTransactionWithRetries` instead.

tmackinnon: Oh this is a really good flag, that I hadn't thought of! I like the idea of splitting up the function into two. I'll give that a go.

server/test/unit/dbTest/beginTransactionTest.js
tmackinnon: Yes - I originally had an assertion for both bot `null` and `expect(pgClient).to.be.equal.to(clientStub)` but it felt redundant. Would you suggest to use both?

server/test/unit/dbTest/beginTransactionTest.js
tmackinnon: Good idea - addded!

server/test/unit/dbTest/beginTransactionTest.js
tmackinnon: Added!

server/db/db.js
trisapeace: You see this sort of pattern quite often in languages like Java that allow you to reuse a function name with different parameters. Unfortunately you can't do that in JS, so the two functions need different names, which can be annoying. Oh well! Hopefully it makes things clearer.

server/db/db.js
```diff
@@ -50,7 +50,7 @@ function createSensorsVitalFromRow(r, allLocations) {
   return new SensorsVital(r.id, r.missed_door_messages, r.is_door_battery_low, r.door_last_seen_at, r.reset_reason, r.state_transitions, r.created_at, r.is_tampered, location)
 }
 
-async function beginTransaction(retryCount = 0) {
+async function runBeginTransactionWithRetries(retryCount) {
```
ahmaddaadaa, line 53: Can you explain the rationale behind renaming the function from `beginTransaction` to `runBeginTransactionWithRetries`? I couldn't find any instances where `runBeginTransactionWithRetries` is updated with the new name in files like BraveAlerterConfiguration.js and vitals.js, which depend on it.

ahmaddaadaa, line 53: Does `retryCount` need to be initialized anywhere?

tmackinnon: Nope! Since retryCount is a parameter, it will get the value of whatever value is passed when the function is called.

The code before had `beginTransaction(retryCount = 0)` because I wanted to give `retryCount` a default value of `0` in the case of `beginTransaction` being called with no argument. But because we've changed code, giving `retryCount` a default value is not necessary.

server/test/integration/dbTest/beginTransactionTest.js
```diff
@@ -0,0 +1,112 @@
...
+
+    it('should log the second error', () => {
+      expect(helpers.logError).to.be.calledWith(`Error running the beginTransaction query: Error: some error`)
+    })
+
+    it('should return null', () => {
+      expect(pgClient).to.be.null
+    })
+  })
+})
```
micahdbak, line 112: Great tests!

micahdbak: Minor question; should these tests be under `integration` or `unit`? My understanding is that unit tests stub surrounding logic so as to observe only the functionality of some unit (class, function), whereas integration tests check the functionality of some unit as it integrates with another unit. Correct my understanding if I'm wrong, but I feel like these tests are unit tests and not integration tests?

micahdbak (commented):
Everything looks good! Just have a small question.

server/test/integration/dbTest/beginTransactionTest.js
tmackinnon: Very fair question! I actually asked Theresa the same thing once I split the functionality into two functions, and she said it would be considered integration tests, because we are hitting the exported `beginTransaction`(since that's what we want to work) as opposed to just testing `runBeginTransactionWithRetries`.

micahdbak: OK, cool!

server/db/db.js
```diff
@@ -89,6 +91,10 @@ async function beginTransaction(retryCount = 0) {
   return pgClient
 }
 
+async function beginTransaction() {
+  return runBeginTransactionWithRetries(0)
```
micahdbak, line 95: I believe this should be `return await runBeginTransactionWithRetries(0)`. As `beginTransaction` is an `async` function, it returns a Promise that is resolved upon a return statement; but as `runBeginTransactionWithRetries` is also an async function, and it is being returned without awaiting, then `beginTransaction` returns a Promise that when awaited gives another Promise... which I don't think is what you want.

server/test/integration/dbTest/beginTransactionTest.js
micahdbak, line 112: I think it would be valuable to write some test cases where the return value of `beginTransaction` is used as a condition for running some subsequent code that is stubbed. That way, you can check that subsequent code is not run in the event of a deadlock error. This could be added as an additional `it` statement within the describes you've already implemented.

micahdbak (commented):
One important comment on async funkiness and a less important suggestion for more test cases!

server/db/db.js
micahdbak: I think it's possible that because there are no `await`s used in `beginTransaction`, it being `async` is being ignored, and the first level Promise is not occurring at all, and the Promise from `runBeginTransactionWithRetries` is being returned instead. However, I think it'd be good to be more explicit in the code on how the Promises are being handled, than to hope JavaScript interprets this in such a way! (I.e., add the `await` before `runBeginTransactionWithRetries`.)

tmackinnon: Thanks Micah, I actually originally had `return await runBeginTransactionWithRetries(0)` in my code, but adding `await` caused an ESLint error, saying that adding `await` added redundancy, which is why it was taken out. 

I looked into the ESLint error here: https://eslint.org/docs/latest/rules/no-return-await and it looks like this rule was eventually deprecated. What do you think about me changing the code to the following so that I can include `await` and bypass the ESLint error: 
```
async function beginTransaction() {
  const pgClient = await runBeginTransactionWithRetries(0)
  return pgClient
}
```

micahdbak: Oo, I think that's a great workaround!! I was thinking an alternative is to just remove `async` from `beginTransaction`, but then `await`ing a non-async function might make ESLint complain, even though `beginTransaction` always returns a Promise. The example you've just written is super clear and I prefer it to removing async from the function :) .

tmackinnon: Thanks Micah, I actually originally had `return await runBeginTransactionWithRetries(0)` in my code, but adding `await` caused an ESLint error, saying that adding `await` added redundancy, which is why it was taken out. 

I looked into the ESLint error here: https://eslint.org/docs/latest/rules/no-return-await and it looks like this rule was eventually deprecated. What do you think about changing the code to the following so that I can include `await` and bypass the ESLint error: 
```
async function beginTransaction() {
  try {
    const pgClient = await runBeginTransactionWithRetries(0)
    return pgClient
  } catch (e) {
    helpers.logError(`Would add a more descriptive error: ${e}`)
    return null
  }
}
```

server/test/integration/dbTest/beginTransactionTest.js
tmackinnon: Hmm I'm not sure I understand what you mean, do you mind clarifying? Is it that, if there's a deadlock you want to have a test that `return null` doesn't happen on line 85?

server/db/db.js
tmackinnon: Updated!

server/test/integration/dbTest/beginTransactionTest.js
micahdbak: I mean in the sense that the reason we rolled back to v10.0.0 was because a call to `beginTransaction` that needed to retry had exited early, and continued onto later code, while the retry call was still running. If you could replicate that scenario in a test case in that other code waits for `beginTransaction` to complete, and you test for whether or not that code was run, that would be good to observe.

server/db/db.js
micahdbak: Yay!

micahdbak: I also like the addition of catching any errors.

server/test/integration/dbTest/beginTransactionTest.js
tmackinnon: Thanks for clarifying Micah! I see what you mean but I think what you're describing is already being tested (correct me if I'm wrong). 

Currently in `db.js` the only line after the retry is `return null` on line 85. So if there was a deadlock, it would go to the catch block from line line 70-82, on line 82 the retry would get called. If the code wasn't working as expected and continued on (in the way that caused the bug) then it would hit line 85 and return `null`. This is case is already tested in the first describe block on line 55-57 in `beginTransactionTest.js`:
```
    it('should return a valid pgClient', () => {
      expect(pgClient).to.be.equal(clientStub)
    })
```
If the code continued on and hit line 85 (returning null) this test would not pass.

micahdbak: Cool! I didn't notice that, but I think that's sufficient. Thanks!

tmackinnon: Thanks Micah!

micahdbak (approved):
No comments. I think this PR is good to go!

server/test/integration/dbTest/beginTransactionTest.js
```diff
@@ -0,0 +1,112 @@
...
+      expect(helpers.log).to.be.calledWith(`Retrying runBeginTransactionWithRetries.`)
+    })
+
+    it('should log the second error', () => {
+      expect(helpers.logError).to.be.calledWith(`Error running the runBeginTransactionWithRetries query: Error: some error`)
+    })
+
+    it('should return null', () => {
+      expect(pgClient).to.be.null
+    })
```
trisapeace, line 110: In this case, we're also going to try to rollback the transaction right? You could add a test for that case too.

trisapeace (approved):
Cool. Looks good to me. I've commented on a couple of things, but they are both minor and up to you.

server/db/db.js
```diff
@@ -89,6 +91,15 @@ async function beginTransaction(retryCount = 0) {
   return pgClient
 }
 
+async function beginTransaction() {
+  try {
+    const pgClient = await runBeginTransactionWithRetries(0)
+    return pgClient
```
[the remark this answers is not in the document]
tmackinnon: Thanks Theresa, I originally tried to have it as one line but our ESLint wouldn't allow it, calling it "redundant". But I looked into the ESLint error documentation: https://eslint.org/docs/latest/rules/no-return-await and it turns out this rule was deprecated, so that's why I had to implement it this way.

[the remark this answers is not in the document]
trisapeace: Haha, OK. Thanks for trying! All good :smile:

server/test/integration/dbTest/beginTransactionTest.js
tmackinnon: Oh this is a good point! My thinking was that because I hadn't made any changes to the rollback code (besides moving it up a few lines) that I wouldn't need to test it. But maybe I should just to be safe.

tmackinnon: Added! This was a good flag because (as per our discussion in Slack) we realized `this.rollbackTransaction(pgClient)` should actually be `rollbackTransaction(pgClient)` since `this` isn't referring to a class, just the module `db.js`. 

Thanks @trisapeace!

server/test/integration/dbTest/beginTransactionTest.js
```diff
@@ -85,7 +85,10 @@ describe('db.js integration tests: beginTransaction', () => {
 
   describe('when beginTransaction fails twice, deadlocking on the first attempt and throwing an error with a valid pgClient on the second attempt', () => {
     let clientStub
+    let rollbackTransactionStub
     beforeEach(async () => {
+      rollbackTransactionStub = sandbox.stub()
+      db.__set__('rollbackTransaction', rollbackTransactionStub)
```
trisapeace, line 91: Because `rollbackTransaction` is exported by `db.js`, I think that you can use `sandbox.stub(db, 'rollbackTransaction') instead of `rewire`.

server/test/integration/dbTest/beginTransactionTest.js
trisapeace: Thanks you! I'm glad we found that. It might also have led to some of the weirdness around deadlocks.

What might be fun now is to test on your local machine. You can turn on your DB Logging and make it so that an error is always thrown for the first query and then look in the logs to make sure that it does the rollback.

server/test/integration/dbTest/beginTransactionTest.js
```diff
@@ -108,5 +111,9 @@ describe('db.js integration tests: beginTransaction', () => {
     it('should return null', () => {
       expect(pgClient).to.be.null
     })
+
+    it('should try to rollback the transaction', () => {
+      expect(rollbackTransactionStub).to.be.called
```
trisapeace, line 116: Cool. What do you think about adding `expect(rollbackTransactionStub).to.not.be.called` to the other test cases in this file?

trisapeace (commented):
Thanks for questioning the `this`!

server/test/integration/dbTest/beginTransactionTest.js
tmackinnon: After looking into this, stubbing rollbackTransaction doesn't seem to be working, and we think perhaps the rewire on line 11 is causing this `const db = rewire('../../../db/db')`. Because we still need to use the rewire for other reasons, I will keep this code as is.

server/test/integration/dbTest/beginTransactionTest.js
tmackinnon: Ahh very this is a very good point - will do!

server/test/integration/dbTest/beginTransactionTest.js
tmackinnon: As per our discussion, the using the debugger (when the rollbackStub wasn't working at first) confirmed that `rollbackTransaction()` was in fact being called hitting all the code within the function. So I don't think I'll need to do this as well. Thanks for all your support Theresa!

trisapeace (approved):
Sweet. Thanks @tmackinnon !

PR #253, opened 2023-11-20
Fixed response for PA Google API in the event of failure

micahdbak:
The previous PR for the PA Google API had an error in that requests that fail (Unauthorized, or Bad request) never resolved.
This PR fixes that.

Test Plan:
- [x] Existing test cases pass
- [x] Works with PA as expected


tmackinnon (approved):
Looks good - approved!

PR #258, opened 2023-12-08
Add Route to Check the Database Connection (CU-86dqkw7ef)

tmackinnon:
## PA Route - Checking Database Connection
This PR is part of the broader Clickup task to create a system status page in PA. These changes include creating a route to check if the BraveSensors database connection is working as expected (by running one simple db query). PA will hit this route with a `googleIdToken` and `braveKey`. Upon getting the response, PA will  display a green icon if the connection is working or a red icon if it is not. 

### Test Plan:
- Unit Test for `pa.handleCheckDatabaseConnection`
   - [x] Valid request with successful database connection
   - [x] Request with invalid braveKey
   - [x] Valid request with database error
- Other Testing
   - [x] Test PA route with local environment where server is working but database is not connnected (status 503)
   - [x] Travis Passes
   - [x] Deploy to Dev - Test with Postman
      - [x] Valid API Key & GoogleIdToken
      - [x] Invalid API Key & valid GoogleIdToken
      - [x] Valid API Key & invalid GoogleIdToken
   - [x] Deploy to Dev - Smoketest
   - [x] Deploy to Dev - test route is working with PA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ the code was deployed.
 
 ## [Unreleased]
 
+## Added
+
+- PA route /pa/check-database-connection to check if database connection is working as expected (CU-86dqkw7ef).
+
 ## [10.3.0] - 2023-12-07
 
 ### Added
--- a/server/pa.js
+++ b/server/pa.js
@@ -209,6 +209,36 @@ async function handleMessageClients(req, res) {
   }
 }
 
+const validateCheckDatabaseConnection = Validator.body(['braveKey', 'googleIdToken']).trim().notEmpty()
+
+async function handleCheckDatabaseConnection(req, res) {
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  if (validationErrors.isEmpty()) {
+    const braveAPIKey = req.body.braveKey
+
+    if (paApiKeys.includes(braveAPIKey)) {
+      try {
+        const client = await db.getFirstClient()
+        if (client !== null) {
+          res.status(200).send({ message: 'success' })
+        } else {
+          res.status(500).send({ message: 'Error in Database Access' })
+        }
+      } catch (err) {
+        helpers.logError(err)
+        res.status(500).send({ message: 'Error in Database Access' })
+      }
+    } else {
+      res.status(401).send({ message: 'Invalid API Key' })
+    }
+  } else {
+    const errorMessage = `Bad request to ${req.path}: ${validationErrors.array()}`
+    helpers.log(errorMessage)
+    res.status(400).send(errorMessage)
+  }
+}
+
 module.exports = {
   validateGetGoogleTokens,
   getGoogleTokens,
@@ -222,4 +252,6 @@ module.exports = {
   validateGetSensorClients,
   validateSensorPhoneNumber,
   validateMessageClients,
+  validateCheckDatabaseConnection,
+  handleCheckDatabaseConnection,
 }
--- a/server/pa.js
+++ b/server/pa.js
@@ -219,12 +219,8 @@ async function handleCheckDatabaseConnection(req, res) {
 
     if (paApiKeys.includes(braveAPIKey)) {
       try {
-        const client = await db.getFirstClient()
-        if (client !== null) {
-          res.status(200).send({ message: 'success' })
-        } else {
-          res.status(500).send({ message: 'Error in Database Access' })
-        }
+        await db.getCurrentTime()
+        res.status(200).send({ message: 'success' })
       } catch (err) {
         helpers.logError(err)
         res.status(500).send({ message: 'Error in Database Access' })
--- a/server/pa.js
+++ b/server/pa.js
@@ -223,7 +223,6 @@ async function handleCheckDatabaseConnection(req, res) {
         res.status(200).send({ message: 'success' })
       } catch (err) {
         helpers.logError(err)
-        helpers.logSentry(err)
         res.status(503).send({ message: 'Error in Database Access' })
       }
     } else {
--- a/server/pa.js
+++ b/server/pa.js
@@ -219,11 +219,12 @@ async function handleCheckDatabaseConnection(req, res) {
 
     if (paApiKeys.includes(braveAPIKey)) {
       try {
-        await db.getCurrentTime()
+        await db.getCurrentTimeForHealthCheck()
         res.status(200).send({ message: 'success' })
       } catch (err) {
         helpers.logError(err)
-        res.status(500).send({ message: 'Error in Database Access' })
+        helpers.logSentry(err)
+        res.status(503).send({ message: 'Error in Database Access' })
       }
     } else {
       res.status(401).send({ message: 'Invalid API Key' })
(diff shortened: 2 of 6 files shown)


server/pa.js
```diff
@@ -209,6 +209,32 @@ async function handleMessageClients(req, res) {
...
+  if (validationErrors.isEmpty()) {
+    const braveAPIKey = req.body.braveKey
+
+    if (paApiKeys.includes(braveAPIKey)) {
+      try {
+        await db.getCurrentTime()
+        res.status(200).send({ message: 'success' })
+      } catch (err) {
+        helpers.logError(err)
+        res.status(500).send({ message: 'Error in Database Access' })
```
trisapeace, line 226: What do you think about using status code `503` to be a little more specific?
Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ the code was deployed.
```
trisapeace, line 11: This looks like a nice article talking about doing this sort of thing: https://testfully.io/blog/api-health-check-monitoring/

The part about specifying no-cache when hitting your endpoint is a good idea to remember when you work on the PA side of thing.

server/routes.js
```diff
@@ -35,6 +35,7 @@ function configureRoutes(app) {
   app.post('/pa/get-sensor-clients', pa.validateGetSensorClients, googleHelpers.paAuthorize, pa.handleGetSensorClients)
   app.post('/pa/sensor-twilio-number', pa.validateSensorPhoneNumber, googleHelpers.paAuthorize, pa.handleSensorPhoneNumber)
   app.post('/pa/message-clients', pa.validateMessageClients, googleHelpers.paAuthorize, pa.handleMessageClients)
+  app.post('/pa/check-database-connection', pa.validateCheckDatabaseConnection, googleHelpers.paAuthorize, pa.handleCheckDatabaseConnection)
```
trisapeace, line 38: Will `googleHelpers.paAuthorize` work with a `GET` request?

I feel like that would be more appropriate than a `POST` in this case since it's getting the information of whether it's online or not.

trisapeace, line 38: What do you think about making the route more generic as suggested in the article, something like `/pa/health` or `pa/check-health`? 

I'm OK either way. What you've written is accurate for what it's doing now and if it were extended in the future then the route could change at that time too.

server/test/unit/paTest/handleCheckDatabaseConnectionTest.js
```diff
@@ -0,0 +1,74 @@
...
+      )
+    })
+    it('should respond with status 401', () => {
+      expect(this.res.status).to.be.calledWith(401)
+    })
+  })
+
+  describe('for a valid request where an error occurs during database access', () => {
+    beforeEach(async () => {
+      sandbox.stub(db, 'getCurrentTime').rejects(new Error('fake error'))
```
trisapeace, line 59: This looks good.

Can you please also stub the `db.getCurrentTime` call for the other tests too? Because these are unit tests, we don't want them actually accessing the DB otherwise they could unexpectedly fail. For the other tests, you'll set it up so that the DB call returns successfully.

server/pa.js
trisapeace, line 226: I wonder if we want a Sentry message here too?

trisapeace (commented):
Looking good!

server/routes.js
tmackinnon: Good question! I originally tried a GET request but I realized `googleHelpers.paAuthorize` looks for the `googleIdToken` in `req.body` and if I made a GET request in PA, data can only get passed via parameters (as opposed to a POST request, where you can pass data via the request body). So because of that I chose to use a POST request.

server/test/unit/paTest/handleCheckDatabaseConnectionTest.js
tmackinnon: Yes will do!

server/pa.js
tmackinnon: Yes I think that's a good idea, will update!

CHANGELOG.md
tmackinnon: Oh that's a really good idea! I'll make sure to imnclude that in the request headers in PA.

server/routes.js
tmackinnon: I was actually struggling to decide on an appropriate name for this route, so I appreciate your thoughts on this! I like `pa/health` so I'll update it to that.

trisapeace: Gotcha. I do think that we'll eventually want `GET` calls to work from PA, when we start moving the dashboard over there. So it's not something that needs to be changed now, but maybe @micahdbak can think about what changes that might entail.

trisapeace: :+1: 

As they say, the two hardest problems in software engineering:
1. Naming things
2. Cache invalidation
3. Off-by-one errors

micahdbak: The main problem with authenticating a GET request is that the googleIdToken is far too long to be contained in the query string of a request... I think we could look into setting Authorization headers for the request? Though I have no experience with that...

server/pa.js
tmackinnon: I believe `helpers.logError(err)` does log to Sentry. Unless you're talking about a different line?

tmackinnon: Ahh that's a good call - I will add it!

server/routes.js
trisapeace: Oh! OK. I wonder how other people do it.

server/pa.js
trisapeace: Yes, you're right, it does log to Sentry too!

tmackinnon: Whoops I second guessed myself and deleted my comment about `helpers.logError()` including a sentry log. I will leave as is.

server/pa.js
```diff
@@ -209,6 +209,32 @@ async function handleMessageClients(req, res) {
...
+    } else {
+      res.status(401).send({ message: 'Invalid API Key' })
+    }
+  } else {
+    const errorMessage = `Bad request to ${req.path}: ${validationErrors.array()}`
+    helpers.log(errorMessage)
+    res.status(400).send(errorMessage)
+  }
+}
+
```
micahdbak, line 237: Yay! Simple but hopefully effective.

server/test/unit/paTest/handleCheckDatabaseConnectionTest.js
```diff
@@ -0,0 +1,76 @@
...
+      )
+    })
+    it('should log the error', () => {
+      expect(helpers.logError).to.be.called
+    })
+    it('should respond with status 503', () => {
+      expect(this.res.status).to.be.calledWith(503)
+    })
+  })
+})
```
micahdbak, line 76: Good test cases!

micahdbak (approved):
Everything looks good, thanks for starting this task!

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ the code was deployed.
 
 ## [Unreleased]
 
+## Added
+
+- PA route /pa/check-database-connection to check if database connection is working as expected (CU-86dqkw7ef).
```
trisapeace, line 16: @micahdbak , when you make the next change in Sensors, would you please update the route in the CHANGELOG to `/pa/health`? Thanks!

PR #262, opened 2024-01-10
Add stop/start environment jobs

phenom5886:


.github/workflows/stop-environment.yml
```diff
@@ -0,0 +1,66 @@
+name: Stop AWS Environment
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "The environment you wish to start"
```
trisapeace, line 7: @phenom5886 , I think there's a copy-paste mistake here and that this should be "stop" instead of "start".

.github/workflows/start-environment.yml
```diff
@@ -0,0 +1,67 @@
+name: Start AWS Environment
+
+on:
+  workflow_dispatch:
+    inputs:
```
trisapeace, line 5: @phenom5886 , I notice that there isn't an input field for the number of hours before the environment stop itself like Tara described in [Slack](https://bravely.slack.com/archives/C060STZJY4C/p1702424244278689). Is this something that you're planning to add once this is in `main`?

trisapeace:
Task linked: [CU-8678ydyqw Turn off/on Dev and Staging Servers during off hours](https://app.clickup.com/t/8678ydyqw)

trisapeace (commented):
Thanks @phenom5886 .
This is great! So looking forward to this feature. I know that you're planning to make some changes now that this is in `main`, but I added some comments just to highlight a couple things for you.

.github/workflows/start-environment.yml
micahdbak: I believe that portion of the request was a little more complex and might bleed into a new block of Alex's hours, so we decided to keep this initial version simple.

@phenom5886 , just to clarify, do these Actions turn the respective databases on and off as well?

trisapeace: Thanks @micahdbak . 

The way that I read the [Slack message](https://bravely.slack.com/archives/C060STZJY4C/p1703003883616199) from @phenom5886 was that part #1 would be a quick win (which includes an input field for number of hours), but that part #2 where the environments were turned on automatically during deployment was going to take too long. Was this scope updated later somewhere else?

PR #267, opened 2024-01-26
Upgrade brave-alert-lib to v12.3.0 (CU-13kqjz8).

micahdbak:
The updates to brave-alert-lib does not concern BraveSensor just yet; it allows PA to be authorized with the Authorization header, in addition to submission of Google ID Token in the body of a post request.


trisapeace:
Task linked: [CU-13kqjz8 Combine Buttons and Sensor servers](https://app.clickup.com/t/13kqjz8)

PR #268, opened 2024-02-05
Console function to toggle external BLE antenna usage

johnny-brave:
This is a draft PR for a limited release firmware version will allow the sensor to toggle between using the internal and external antenna for BLE.

---
Test Plan

- [ ] start with a sensor with the external antenna attached
- [ ] select 0 in the console function to use the internal antenna
- [ ] check if the sensor receives door sensor open/close messages at incremental distance away from the sensor (note the distance at which the door sensor stops receiving messages)
- [ ] select 1 in the console function to use the external antenna
- [ ] verify that the sensor is able to receive door sensor messages beyond the internal antenna maximum and note the maximum distance for the external antenna
- [ ] remove the external antenna from the sensor
- [ ] select 0 in particle function and verify that the sensor receives door sensor message at the same distance threshold as previously recorded


CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ the code was deployed.
 
 ## [Unreleased]
 
+<<<<<<< HEAD
```
trisapeace, line 15: Some residue from merges.

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -359,6 +360,42 @@ int im21_door_id_set(String command) {
     return (int)strtol(buffer, NULL, 16);
 }
 
+int toggle_ble_antenna(String command) {
```
trisapeace, line 363: I suggest changing the name to give some indication of what 1 and 0 mean. Since 1 (which normally means true) means the External Antenna and 0 (which normally means false) is the Internal Antenna, what do you think about calling this `toggle_external_ble_antenna`?

trisapeace, line 363: Could you please document this function in the firmware/README.md file?

firmware/boron-ins-fsm/src/imDoorSensor.h
```diff
@@ -39,13 +40,22 @@ typedef struct IMDoorID {
...
+
 //************************global variable declarations***************************
 // extern os_queue_t bleQueue;
 extern os_queue_t bleHeartbeatQueue;
 
 // needs to be global because it is used in setup(), loop(), and console function
 extern IMDoorID globalDoorID;
 
+// antenna that is being used for bluetooth LE (internal or external)
+extern uint8_t bleAntennaType;
```
trisapeace, line 57: I would have expected the type here to be `antennaType` instead of `uint8_t` since that is what you're going to be storing in it. 

If you want to make sure that the enum is of type `uint8_t` (which I think is a good idea), it looks like you can do something like this example:
```
// enum that takes 16 bits
enum smallenum: [std::int16_t](http://en.cppreference.com/w/cpp/types/integer)
{
    a,
    b,
    c
};
```
from here (https://en.cppreference.com/w/cpp/language/enum#Example)

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -6,6 +6,7 @@
 #include "../src/stateMachine.h"
 
 IMDoorID globalDoorID = {0xAA, 0xAA, 0xAA};
+uint8_t bleAntennaType;
```
trisapeace, line 9: So this would now use `antennaType` as the type instead of `uint8_t`.

firmware/boron-ins-fsm/src/imDoorSensor.h
```diff
@@ -39,13 +40,22 @@ typedef struct IMDoorID {
     unsigned char byte3;
 } IMDoorID;
 
+typedef enum
+{
+    INTERNAL,
+    EXTERNAL
+} antennaType;
```
trisapeace, line 47: Reading this value later in the file, I found it difficult to identify that `antennaType` was a datatype because its values are the types of antennas. 

What do you think about using the `_t` convention? So this would be something like `antennaType_t`?

firmware/boron-ins-fsm/src/BraveSensorProductionFirmware.ino
```diff
@@ -23,8 +23,6 @@ void setup() {
     // enable reset reason
     System.enableFeature(FEATURE_RESET_INFO);
 
-    // use internal antenna on Boron for BLE
-    BLE.selectAntenna(BleAntennaType::INTERNAL);
```
trisapeace, line 27: I have vague recollections about there being a timing issue with selecting the antenna. I wonder whether it needs to be selected before we run the `setupIM()` function? Or whether it's something that can actually be changed after the initial setup function is done. I'm going to try to see if I can find anything written down about this, but I wonder whether this should be left in so that there is at least the internal antenna when the devices powers up and then hopefully it will be changed when it hits the first run of  `loop`.

trisapeace: No, I can't find anything. And it's possible that my recollections were related to the wifi antenna that we used to have on the Argons. 

So nothing to do about this comment other than be extra cautious to make sure that this works before deploying it to everyone.

firmware/boron-ins-fsm/src/flashAddresses.h
```diff
@@ -38,6 +38,10 @@
 #define ADDR_INITIALIZE_STATE0_OCCUPANT_DETECTION_TIMER_FLAG 29  // uint16_2 = 2 bytes
 #define ADDR_STATE0_OCCUPANT_DETECTION_TIMER                 31  // uint32_t = 4 bytes
 
-// next available address is 31 + 4 = 35
+// bluetooth antenna type
+#define ADDR_INITIALIZE_BLE_ANTENNA_TYPE_FLAG 35  // uint16_t = 2 byte
+#define ADDR_BLE_ANTENNA_TYPE                 37  // uint8_t = 1 byte
+
+// next available address is 37 + 1 = 38
```
trisapeace, line 45: Thanks for updating the next available address!

firmware/boron-ins-fsm/src/imDoorSensor.cpp
```diff
@@ -12,6 +12,7 @@
 // initialize door ID to agreed upon intial value
 IMDoorID globalDoorID = {0xAA, 0xAA, 0xAA};
 os_queue_t bleQueue;
+uint8_t bleAntennaType = antennaType::INTERNAL;
```
trisapeace, line 15: This would also be `antennaType` instead of `uint8_t`.

firmware/boron-ins-fsm/src/imDoorSensor.cpp
```diff
@@ -58,6 +59,32 @@ void initializeDoorID() {
...
 
+void initializeAntennaType() {
+    uint16_t initializeBLEAntennaTypeFlag;
+
+    // Boron flash memory is initialized to all F's (1's)
+    EEPROM.get(ADDR_INITIALIZE_BLE_ANTENNA_TYPE_FLAG, initializeBLEAntennaTypeFlag);
+    Log.info("Antenna type flag is 0x%04X", initializeBLEAntennaTypeFlag);
+
+    if (initializeBLEAntennaTypeFlag != INITIALIZE_ANTENNA_TYPE_FLAG) {
+        uint8_t antennaType_default = antennaType::INTERNAL;
```
trisapeace, line 70: This would also be `antennaType` instead of `uint8_t`.

firmware/boron-ins-fsm/test/DoorSensorTests
trisapeace, line 1: Could you please add this file to the `firmware/.gitignore` and remove it from this PR so that it isn't added to the repo? We shouldn't need to commit any binary files because these will be compiled when they are run.

firmware/boron-ins-fsm/test/ins3331Tests
trisapeace, line 1: Could you please add this file to the `firmware/.gitignore` and remove it from this PR so that it isn't added to the repo? We shouldn't need to commit any binary files because these will be compiled when they are run.

firmware/boron-ins-fsm/test/consoleFunctionTests.cpp
```diff
@@ -563,3 +564,109 @@ SCENARIO("Reset_Stillness_Timer_For_Alerting_Session", "[reset stillness timer f
         }
     }
 }
+
+SCENARIO("Set BLE Antenna Type", "[ble antenna type]") {
```
trisapeace, line 568: Thanks for adding new tests! These look great!

firmware/boron-ins-fsm/test/mocks/mock_ble.h
```diff
@@ -70,6 +70,14 @@ class BleScanFilter {
 };
 
 // Fake class for BLE
+
```
trisapeace, line 73: Cool. Glad you figured out what needed to be added to this file!

firmware/boron-ins-fsm/src/consoleFunctions.cpp
johnny-brave: maybe "use_external_ble_antenna"?

firmware/boron-ins-fsm/src/imDoorSensor.h
johnny-brave: perhaps it's better to do away with this enum altogether. I thought it would make the code more readable, but I think that can be accomplished by changing the variable to `bool useExternalBLEAntenna`

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -359,6 +360,42 @@ int im21_door_id_set(String command) {
...
+        // any string longer than 1 char is invalid input, so
+        returnFlag = -1;
+    }
+    // if e, echo whether debug publishes are on
+    else if (*holder == 'e') {
+        returnFlag = usingExternalBLEAntenna;
+    }
+    else if (*holder == '0') {
+        usingExternalBLEAntenna = false;
+        EEPROM.put(ADDR_BLE_ANTENNA_TYPE, usingExternalBLEAntenna);
```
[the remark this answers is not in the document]
johnny-brave: I don't believe `usingExternalBLEAntenna` is defined in `consoleFunction.h`, it is a global bool defined in `imDoorSensor.h`. I also verified that bool is indeed 1 byte long by reading the memory on bootup.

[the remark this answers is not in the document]
micahdbak: Oh, shoot, I misread the definition of the function as the variable; disregard this comment then :)

firmware/boron-ins-fsm/src/consoleFunctions.cpp
```diff
@@ -30,6 +30,7 @@ void setupConsoleFunctions() {
     Particle.function("Change_Long_Stillness_Timer", long_stillness_timer_set);
     Particle.function("Change_INS_Threshold", ins_threshold_set);
     Particle.function("Change_IM21_Door_ID", im21_door_id_set);
+    Particle.function("Use External BLE Antenna", use_external_ble_antenna);
```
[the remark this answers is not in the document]
johnny-brave: Hm, I wonder why these function names use underscores. I doesn't seem to be required. Nevertheless, I will change the function name for consistency.

firmware/README.md
```diff
@@ -444,6 +446,23 @@ The IM door sensors each have a sticker on them with their door IDs. On the bott
...
+**Argument(s):**
+
+1. Enter 1 to use the external antenna
+2. Enter 0 to use the internal antenna
+
+**Return(s):**
+
+- 1 or 0, the value that was inputted
+- 1 or 0, the current selection of the antenna if 'e' is inputted
+- -1, invalid input detected
```
trisapeace, line 464: Since the behaviour is pretty much the same as the `toggle_debugging_publishes` console function, could you please express the Arguments and Returns in the same way? I don't mind if you use your way in both or that you change yours to match the `toggle_debugging_publishes`. Either seem clear to me, I'd just like them to be consistent. Thanks!

firmware/boron-ins-fsm/src/imDoorSensor.h
trisapeace: Sure! A boolean makes sense to me and would probably require less new code.

firmware/boron-ins-fsm/test/DoorSensorTests
trisapeace, line 1: Sweet! Good job removing this :+1:  It can be tricky to get the order right when you're also then adding it to the gitignore.

firmware/boron-ins-fsm/src/flashAddresses.h
```diff
@@ -38,6 +38,10 @@
 #define ADDR_INITIALIZE_STATE0_OCCUPANT_DETECTION_TIMER_FLAG 29  // uint16_2 = 2 bytes
 #define ADDR_STATE0_OCCUPANT_DETECTION_TIMER                 31  // uint32_t = 4 bytes
 
-// next available address is 31 + 4 = 35
+// bluetooth antenna type
+#define ADDR_INITIALIZE_BLE_ANTENNA_TYPE_FLAG 35  // uint16_t = 2 byte
+#define ADDR_BLE_ANTENNA_TYPE                 37  // uint8_t = 1 byte
```
trisapeace, line 43: I'd also change these names to make it clearer that true means external.

And the second value is now a boolean, right? So the comment can be updated?

firmware/boron-ins-fsm/src/imDoorSensor.cpp
```diff
@@ -58,6 +59,32 @@ void initializeDoorID() {
...
+    if (initializeBLEAntennaTypeFlag != INITIALIZE_ANTENNA_TYPE_FLAG) {
+        bool externalBLEAntennaDefault = false;
+        EEPROM.put(ADDR_BLE_ANTENNA_TYPE, externalBLEAntennaDefault);
+        initializeBLEAntennaTypeFlag = INITIALIZE_ANTENNA_TYPE_FLAG;
+        EEPROM.put(ADDR_INITIALIZE_BLE_ANTENNA_TYPE_FLAG, initializeBLEAntennaTypeFlag);
+        Log.info("BLE antenna type was written to flash on bootup.");
+    }
+    else {
+        EEPROM.get(ADDR_BLE_ANTENNA_TYPE, usingExternalBLEAntenna);
+        if (usingExternalBLEAntenna == true) {
```
trisapeace, line 78: Because `usingExternalBLEAntenna` is a boolean, so it is more idomatic to leave off the `== true` and just have `if (usingExternalBLEAntenna) {`

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ the code was deployed.
 
 ## [Unreleased]
```
trisapeace, line 13: I'm a little concerned about this code being deployed to machines in production without at least a release tag.

If something funny happens on those machines, I'd want to have a way to look back and see the history of what was deployed. Do you know what commit you used to create the binary that is deployed at the clients who have this already? (Remember not to use their names on our public GitHub :wink: ). 

What do you think of a `10.8.0-beta` to indicate that this is a pre-release version? ([ref](https://semver.org/spec/v2.0.0.html))

trisapeace (commented):
Looking great!

firmware/boron-ins-fsm/src/flashAddresses.h
micahdbak: @johnny-brave , do you envision there being further BLE antenna types beyond 'internal' and 'external'? If not, this should probably be updated to something like `ADDR_USING_EXTERNAL_BLE_ANTENNA` or something

PR #273, opened 2024-02-27
Use GitHub Actions instead of Travis (CU-86drj1yqw).

micahdbak:
This PR is in progress.


trisapeace:
Task linked: [CU-86drj1yqw Look at viability of GitHub Actions vs Travis](https://app.clickup.com/t/86drj1yqw)

micahdbak:
Immediately noticing:
- GitHub Actions is *must* faster than Travis.
  - This is likely because Actions are performed in __parallel__, whereas Travis is __sequential__.
  - This is great for several devs working at the same time!
- GitHub Actions is configured very similarly to Travis.
  - Easy to work with!

Still need to:
- [ ] Configure PostgreSQL 12 for the Run Server Tests Workflow
- [ ] Create Workflow for Firmware Tests (+ Linting and Formatting Checks)

.travis.yml
```diff
@@ -1,82 +0,0 @@
...
-    # Run security audit, linting, unit tests, and integration tests
-    - stage: 'server test'
-      language: node_js
-      # node_js: node version specified in .nvmrc
-      install:
-        - cd server
-        - npm ci
-      addons:
-        # install PostgreSQL
-        postgresql: '12'
```
trisapeace, line 53: We might run into a fun time with this. 12 is a old version of postgres (they are up to 16 now ([ref](https://www.postgresql.org/docs/release/))). I'm not sure what the blocker was for upgrading, if there was one.

This looks like it might be helpful: https://til.simonwillison.net/github-actions/different-postgresql-versions although it might be slower than the default service container system that they describe here: https://til.simonwillison.net/github-actions/different-postgresql-versions

trisapeace (commented):
Great start!

.github/workflows/run-server-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+  PA_CLIENT_SECRET_TEST: fakeclientsecret
+  PARTICLE_WEBHOOK_API_KEY_TEST: webhookKey
+
+jobs:
+  ServerTests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js
+        uses: actions/setup-node@v3
```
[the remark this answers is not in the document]
micahdbak: Oo, cool! Yeah I'll change it to v4.

.travis.yml
micahdbak: Looks like it would definitely be slower, and in the interest of minimizing used GitHub Actions minutes, I'll probably keep it at version 12 for now. But I'll add a note that testing different versions of postgres could be beneficial in the future.

trisapeace: Oh, mini miscommunication. I didn't mean that we should try different versions. But just that in that first link he was able to specify a particular version to use, which I don't see in the default service container.

.github/workflows/run-server-tests.yml
```diff
@@ -0,0 +1,59 @@
...
+run-name: Running Server Tests for ${{ github.ref }}
+
+on: [push]
+
+env:
+  PG_USER_TEST: brave
+  PG_DATABASE_TEST: brave
+  PG_PASSWORD_TEST: actions
+  PG_PORT_TEST: 5432
+  PG_HOST_TEST: localhost
```
trisapeace, line 12: Maybe try `postgres` instead of `localhost`. (Ref: https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers#running-jobs-in-containers ).

PR #274, opened 2024-02-27
Fixes for failed deployment

micahdbak:
This is a debug PR to fix some issues that came up during deployment of Sensor v10.8.0.

Issues:
- The smoketest code didn't post `data` to /api/sensorEvent.
- A newly created session that is resettable isn't resettable by the db.
- There is still some mention of Device OS 3.3.1 in the server/README.md that I missed.

Resolved:
- [x] Smoketest issue
- [x] createSession
  - createSession branches into two insert queries; I only updated the first with isResettable logic.

Test Plan:
- [x] Existing test cases pass
- [x] Reset chatbot flow works as expected
  - [x] For an existing session, a prompt is given and a request is accepted once the sensor states that the number of published alerts exceeds the server defined threshold
  - [x] For an existing session, a prompt is given and a request is accepted once the sensor states that the number of published alerts exceeds the server defined threshold
  - [x] The regular chatbot flow works regardless of the reset prompt being given

### DIFF
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -74,9 +74,6 @@ class BraveAlerterConfigurator {
         return
       }
       session = await db.getSessionWithSessionId(alertSession.sessionId, pgClient)
-      console.log('--- FOUND SESSION FOR ALERT SESSION CHANGED CALLBACK ---')
-      console.log(session)
-      console.log(`IS RESETTABLE: ${session.isResettable}`)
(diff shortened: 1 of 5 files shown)


trisapeace (commented):
Awesome! Super glad you caught the `createSession` thing before it was deployed!

micahdbak:
> Super glad you caught the createSession thing before it was deployed!
Benefits of our thorough deployment process! :D

PR #286, opened 2024-07-05
Ew Travis to GitHub Actions (CU-86dthadwg)

elanawood:
Updated repository to use Github Actions instead of Travis.


trisapeace:
Task linked: [CU-86dthadwg Use GitHub Actions instead of Travis CI](https://app.clickup.com/t/86dthadwg)

.github/workflows/CI.yml
```diff
@@ -0,0 +1,135 @@
...
+        run: |
+          sudo service postgresql restart
+          until pg_isready -h localhost -p 5433 -U brave -d bravesensor; do
+            echo "Waiting for PostgreSQL to be ready..."
+            sleep 1
+          done
+
+      - name: Setup PostgreSQL
+        working-directory: ./server
+        run: ./setup_postgresql_local.sh
```
micahdbak, line 102: You'll want to copy the contents of this file and run them directly from this workflow file, because this file is intended for local environments specifically, not necessarily CI. It might break in the future

micahdbak: You'll want to use the setup_postgresql.sh script instead, which is called from inside of setup_postgresql_local.sh

.github/workflows/CI.yml
```diff
@@ -0,0 +1,135 @@
...
+
+      - name: Run Tests
+        working-directory: ./server
+        run: npm run test
+
+  firmware_lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Print message for future implementation
+        run: echo 'Please implement this in the future using GitHub Actions.'
```
micahdbak, line 121: We can make this an upcoming task! The old Travis implementation was removed because of errors

micahdbak (commented):
Overall looking good! One minor change for the Setup PostgreSQL step

Also, I'd recommend discussing with Darrick the names for each step in the Server job. I think the steps should be the same between Buttons/Sensor because the tests are very similar

.github/workflows/CI.yml
```diff
@@ -0,0 +1,120 @@
...
+
+      - name: Check environment variables
+        run: |
+          echo "PG_USER=$PG_USER"
+          echo "PG_DATABASE=$PG_DATABASE"
+          echo "PG_PASSWORD=$PG_PASSWORD"
+          echo "PG_PORT=$PG_PORT"
+          echo "PG_HOST=$PG_HOST"
+
+      - name: Set up PSQL
```
micahdbak, line 69: I think either this or the "Setup PostgreSQL" steps need to be renamed - because they mean the same thing. Maybe this one should be "Install PostgreSQL"?

micahdbak (commented):
One job name change

Darrickg:
dont forget to fix the badges in the readme!

micahdbak (approved):
Great work!! Feel free to merge

PR #294, opened 2024-08-19
Only display clients on the dashboard

CCheng00:
Previously, the dashboard would display all sensors, we suspect that this may be what is causing the load times to be so long.

Removed the sensors on the dashboard, they will still be viewable through the client page.

### DIFF
--- a/server/mustache-templates/landingPage.mst
+++ b/server/mustache-templates/landingPage.mst
@@ -42,13 +42,6 @@
         </div>
     </div>
     
-    <script>
-        function reloadPage() {
-            location.reload(true)
-        }
-        setInterval(reloadPage, 30000)
-    </script>
-    
     <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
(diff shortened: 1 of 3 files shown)


server/mustache-templates/landingPage.mst
```diff
@@ -34,34 +32,16 @@
                     {{#clients}}
                         <tr class="table-data table-secondary">
                             <th scope="row"><a href="/clients/{{id}}">{{displayName}}</a></th>
-                            <td></td>
-                            <td></td>
                             <td>{{#isSendingAlerts}}y{{/isSendingAlerts}}{{^isSendingAlerts}}NOPE{{/isSendingAlerts}}</td>
                             <td>{{#isSendingVitals}}y{{/isSendingVitals}}{{^isSendingVitals}}NOPE{{/isSendingVitals}}</td>
                         </tr>
-                        {{#locations}}
```
micahdbak, line 42: Great!

server/mustache-templates/landingPage.mst
```diff
@@ -34,34 +32,16 @@
...
                     {{/clients}}
                     </tbody>
                 </table>
             </div>
         </div>
     </div>
     
-    <script>
-        function reloadPage() {
-            location.reload(true)
```
micahdbak, line 60: Hmm, I think this script should be kept - location isn't referring to a sensor I'm pretty sure.

This reloads the page every 30 seconds

micahdbak (approved):
Looks great! Thanks!

micahdbak:
actually, @CCheng00 , please update the CHANGELOG.md before squash & merging!

PR #297, opened 2024-08-21
Create setup_reporting_db.sh for monthly data dumps (CU-86duh0eeu)

micahdbak:
From the output of the script:

This script is unsafe - it does the following:

1. It will download data from and slow the source database to a halt.
2. It will drop all necessary data and schemas from the destination database.
3. It will upload the downloaded data to the destination database.

Testing:
- [ ] Run the script when performing the monthly data dump to verify it works

### DIFF
--- a/server/setup_reporting_db.sh
+++ b/server/setup_reporting_db.sh
@@ -33,6 +33,7 @@ if [ $answer != "Y" ]; then
 	printf "OK - Won't do anything then.\n"
 	exit
 fi
+printf "\n"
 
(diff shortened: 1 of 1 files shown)


trisapeace:
Task linked: [CU-86duh0eeu Finish setup_reporting_db.sh for Buttons and Sensors](https://app.clickup.com/t/86duh0eeu)

CCheng00:
Script now works, reporting db available on digital ocean

PR #312, opened 2024-09-26
Make fallback number optional for Sensors dashboard (CU-86du49q25)

Nakul727:
- Changed the sensors dashboard to make the fallback numbers optional for new and edit client
- Updated db query calls to handle empty array for fallback numbers
- Database defaults to {} empty array if we pass empty or undefined parameters for fallback numbers
- Added test cases for a request that contains valid non-empty fields but with no fallbackPhoneNumbers for both new client and edit client functions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ the code was deployed.
 ## [Unreleased]
 
 - Update pw migration instructions (CU-86duhae9t)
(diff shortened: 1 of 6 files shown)


trisapeace:
Task linked: [CU-86du49q25 Make fallback number optional](https://app.clickup.com/t/86du49q25)

CCheng00 (approved):
Good work and nice test cases!

PR #321, opened 2024-12-11
Add status and commissioned_at column for Sensors (CU-86dv9uxta)

Nakul727:
- Added migration script 54 which adds the `status` and `commissioned_at` column to clients
- Added new status field in the edit client dashboard page 
- Updated edit client test cases with new status column

The commissioned_at column in the database is set to NULL initially when the client is created and status is set to LIVE as default (for existing clients in the database). When a new client is created from the dashboard, the status is set to TESTING by default.


trisapeace:
Task linked: [CU-86dv9uxta Add 'commissioned_at' column for Sensors](https://app.clickup.com/t/86dv9uxta)

PR #327, opened 2025-01-28
Reset State to Idle During Occupation if a fresh door close is received (CU-86dvbnzx4).

johnny-brave:
One way that false occupation occurs is if the door contact "open" message is missed by the main sensor. This PR aims to reset the FSM to state 0 if a "fresh" door close message is received while the FSM is in state 2 or 3.

Occupation Detection Timer is already set by this scenario, this reset flag can be set in the same place.


trisapeace:
Task linked: [CU-86dvbnzx4 Set FSM to 0 upon reception of a "fresh close" from the BLE Door Contact](https://app.clickup.com/t/86dvbnzx4)

PR #389, opened 2026-05-14
Cc health endpoint

CCheng00:
Adding a health endpoint for system monitoring


CCheng00:
Also includes MS teams events in the dashboard, add prod env var before deployment

PR #393, opened 2026-06-16
Retry Heartbeat if failed

johnny-brave:
This PR attempts to retry the heartbeat publish to particle if it failed the first time each heartbeat interval.


CCheng00:
Added async + did a bench test

